PR #5, opened 2020-12-21
Merge PluginError into Result

ddebroy:
Merge the PluginError struct into Result. This allows the NRI skeleton and invocation to encode and un-marshal to a single type and cleans up the code a bit.

Error messages from injected errors:
```
nri invoke: plugin: testplugin: injected structured error
```
```
nri invoke: plugin: testplugin: failed to unmarshal plugin output: unexpected end of JSON input: output: <empty> exit code: 1
```
```
nri invoke: plugin: testplugin: failed to unmarshal plugin output: invalid character 'i' looking for beginning of value: output: "injected invalid output"
```
```
nri invoke: plugin: testplugin: exec: "missing": executable file not found in $PATH
```
Signed-off-by: Deep Debroy <ddebroy@apple.com>

### DIFF
--- a/client.go
+++ b/client.go
@@ -108,20 +108,27 @@ func (c *Client) invokePlugin(ctx context.Context, name string, r *types.Request
 	cmd.Stderr = os.Stderr
 
 	out, err := cmd.Output()
+	msg := "output:"
+	if len(out) > 0 {
+		msg = fmt.Sprintf("%s %q", msg, out)
+	} else {
+		msg = fmt.Sprintf("%s <empty>", msg)
+	}
 	if err != nil {
-		// ExitError is returned when there is a non-zero exit status
-		if _, ok := err.(*exec.ExitError); ok {
-			var pe types.PluginError
-			if err := json.Unmarshal(out, &pe); err != nil {
-				return nil, errors.Wrapf(err, "%s: %s", name, out)
-			}
-			return nil, &pe
+		if exitErr, ok := err.(*exec.ExitError); ok {
+			// ExitError is returned when there is a non-zero exit status
+			msg = fmt.Sprintf("%s exit code: %d", msg, exitErr.ExitCode())
+		} else {
+			// plugin did not get a chance to run, return exec err
+			return nil, err
 		}
-		return nil, errors.Wrapf(err, "%s: %s", name, out)
 	}
 	var result types.Result
 	if err := json.Unmarshal(out, &result); err != nil {
-		return nil, err
+		return nil, errors.Errorf("failed to unmarshal plugin output: %s: %s", err.Error(), msg)
+	}
+	if result.Err() != nil {
+		return nil, result.Err()
 	}
 	return &result, nil
 }
--- a/types/v1/types.go
+++ b/types/v1/types.go
@@ -2,7 +2,7 @@ package v1
 
 import (
 	"encoding/json"
-	"fmt"
+	"github.com/pkg/errors"
 )
 
 // Plugin type and configuration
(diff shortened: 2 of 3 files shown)


types/v1/types.go
```diff
@@ -122,6 +101,10 @@ type Result struct {
 	Plugin string `json:"plugin"`
 	// Version of the plugin
 	Version string `json:"version"`
+	// If the result is an error
+	IsError bool `json:iserror`
```
crosbymichael, line 105: we don't need field like this.  I would just have the `ErrorMessage` and then you can put a `.Err() error` method on this struct that returns `nil` if error message == "" or an errors.New(r.ErrorMessage)` if there is an error.  It makes the flow more clear

ddebroy: done

types/v1/types.go
```diff
@@ -122,6 +101,16 @@ type Result struct {
 	Plugin string `json:"plugin"`
 	// Version of the plugin
 	Version string `json:"version"`
+	// error details if IsError is set
+	ErrorMessage string `json:"errormessage"`
```
crosbymichael, line 105: I'd probably just change this to `Error string `json:"error"``

crosbymichael (changes requested):

types/v1/types.go
ddebroy: done

crosbymichael (approved):
LGTM

client.go
```diff
@@ -108,20 +108,27 @@ func (c *Client) invokePlugin(ctx context.Context, name string, r *types.Request
...
+	}
 	if err != nil {
-		// ExitError is returned when there is a non-zero exit status
-		if _, ok := err.(*exec.ExitError); ok {
-			var pe types.PluginError
-			if err := json.Unmarshal(out, &pe); err != nil {
-				return nil, errors.Wrapf(err, "%s: %s", name, out)
-			}
-			return nil, &pe
+		if exitErr, ok := err.(*exec.ExitError); ok {
```
mxpv, line 118: `cmd.ProcessState.ExitCode()` can eliminate `if` branch ?

client.go
```diff
@@ -108,20 +108,27 @@ func (c *Client) invokePlugin(ctx context.Context, name string, r *types.Request
...
+		} else {
+			// plugin did not get a chance to run, return exec err
+			return nil, err
 		}
-		return nil, errors.Wrapf(err, "%s: %s", name, out)
 	}
 	var result types.Result
 	if err := json.Unmarshal(out, &result); err != nil {
-		return nil, err
+		return nil, errors.Errorf("failed to unmarshal plugin output: %s: %s", err.Error(), msg)
```
mxpv, line 128: how do you fill about: 

`errors.Errorf("...  output: %q , exit code: %d", err.Error(), out, cmd.ProcessState.ExitCode())`

in worst case it'll be `output: "", exit code: -1`, but will remove if branches above at all for string formatting.

client.go
ddebroy: Are you suggesting eliminating the `if` statement checking `exec.ExitError`? That allows us to return early if the plugin did not get a chance to run - say a general exec error due to the plugin binary missing or not being in ELF format. In those situations, I think, we want: `plugin: testplugin: exec: \"testplugin\": executable file not found in $PATH` (rather than trying to unmarshal the output and then reporting that there was a failure to unmarshal output).

mxpv: hm, i got the idea. I would rather use `LookPath` to make sure the binary exists before running before even executing the program.(I'd prefer to avoid nested if branches when possible for the sake of readability). But this one would work too.

client.go
ddebroy: `output: ""` was something I wanted to avoid in the error messages as that makes the debug logs in containerd a little unreadable with escapes - `output: \"\"` and may make the person reading the log wonder if something was not captured. The `output: <empty>`  is relatively more clear and authoritative.

mxpv (approved):
LGTM

PR #6, opened 2021-01-06
fix missing import for fmt in skel

crosbymichael:
Signed-off-by: Michael Crosby <michael@thepasture.io>


crosbymichael:
Merging as this is a quick formatting and missing import fix.

PR #15, opened 2021-09-14
Update the examples in README.md

fuweid:
Signed-off-by: Wei Fu <fuweid89@gmail.com>


estesp (approved):
LGTM

go.mod
```diff
@@ -3,9 +3,12 @@ module github.com/containerd/nri
 go 1.14
 
 require (
+	github.com/containerd/cgroups v1.0.1
```
mikebrow, line 6: thinking about whether the example content should affect the go.mod for the project

mikebrow (commented):
Just a question about whether examples should result in having to modify the project go mod...

go.mod
mikebrow: but I do like that it's built tested..

fuweid: Updated. PTAL. Thanks!

mikebrow (approved):
LGTM

Issue #3, opened 2020-11-11
Building a resource assignment policy using NRI, including policies for native/compute resources.


klihub:
We'd like to try and retrofit the current functionality for pod/container resource assignment in [CRI Resource Manager/CRI-RM](https://github.com/intel/cri-resource-manager) as an NRI plugin. Our goals are
  - primary: implement a reasonable default hardware-topology aware assignment policy, and
  - secondary: provide a way for plugging in special application-/vertical-specific policies

The resources of interest are
  - the tangible HW resources the kernel let's us arbitrate
    - native/compute resources (CPU, memory, huge pages)
    - devices
    - LLC cache and memory bandwidth
  - a few other things the kernel let's us control, for instance
    - block I/O throttling
    - RT-scheduling/arbitration of time slices set aside for RT processes

To get our full policy scope working we'd need a way to
  1. track information about running pods and containers
  2. tap into container creation/deletion/update requests and let the runtime know about resource allocation decisions
  3. have an assortment of pre- and post-variants/hooks related to container life-cycle events (create/delete/start/stop)
  4. make changes to existing/running containers, not just the ones being created

These bits of functionality are necessary for the following reasons.
1) We need to track what resources are allocated to containers vs. what is free. Likewise, we need to be able to figure out the relationship between pods and containers, so things like intra-pod affinity can be implemented, where containers within a pod are put close to each other in the hardware topology sense.
2) Resources are assigned to containers in connection with certain lifecycle events. A resource allocation policy plugin needs to tap into these events, make decisions, and let the rest of the container runtime know about/apply the decisions.
3) Enforcing some of the policy decisions require more accurate alignment with the lifecycle of the container than the CRI requests can provide. For instance, assigning a container to an LLC cache clos/class happens by writing container process pids to a special pseudo-filesystem entry. Therefore enforcing a container LLC class is best done in connection with the CRI start request, once a process to run the container command has been forked but before it has actually exec'd the eventual container command. Tapping into the realted basic CRI requests do not provide the necessary resolution for achieving this.
4) Sometimes even simple resource policy decisions related to processing a CRI request has further resource-related consequences on containers other than the one directly involved in the CRI request. For instance, when a container running on a set of exclusively allocated CPUs is deleted, all containers without exlcusive CPU allocations and running in the same 'HW topology CPU pool' should be updated to allow them to run on the newly freed cpuset. 

[...]

PR #16, opened 2021-09-16
Extend scope to enable common pluggable runtime extensions.

klihub:
## Background

This PR aims to extend NRI beyond its current v1 capabilities. The primary goal is to allow NRI to act as a common infrastructure for plugging in extensions to runtimes. Among other things, these runtime extensions could then implement vendor- and domain-specific custom logic for improved container configuration.

The PR is both related to this [KubeCon talk](https://kccnceu2021.sched.com/event/iE1Y/maximizing-workloads-performance-with-smarter-runtimes-krisztian-litkey-alexander-kanevskiy-intel), and to [this NRI issue](https://github.com/containerd/nri/issues/3).

## Description

**Note**: For a more detailed description please refer to [the README](https://github.com/klihub/nri/blob/pr/proto/draft/v2alpha1/README.md).

This PR

  - defines a plugin API that plugins use to interact with NRI
  - defines a runtime API that runtimes use to integrate to NRI
  - implements a stub that can be used as a library or template for writing plugins
  - implements sample plugins demonstrating some of NRI's extended capabilities

Using the plugin- and runtime-integration APIs allows one to hook plugins into the lifecycle of containers managed by a runtime. These plugins can then adjust a limited set of container parameters in connection with some of the container's lifecycle state transitions.

The proposal changes the v1 hook-like incarnation of plugins to more daemon-like entities, where a single instance of a plugin is responsible for handling the full stream of events intended for that plugin. It also defines a ttRPC-based plugin protocol which allows plugins to subscribe to lifecycle events of pods and containers and to make changes to containers during a container's creation, update and stop events. Adjustment is possible to the following container parameters during a container's creation:

  - labels, annotations
  - environment variables
  - set of OCI hooks
  - container mounts
  - devices available to the container
  - assigned resources (as defined by the OCI Spec)
  - container cgroup path

Other than during creation, a container's assigned resources can be updated by plugin.

## Related PRs

There are two more PRs closely related to this proposal:

- [containerd integration to this extended scope NRI](https://github.com/containerd/containerd/pull/6019)
- [cri-o integration to this extended scope NRI](https://github.com/cri-o/cri-o/pull/5318)

Additionally there is a work-in-progress implementation of a more complex, custom resource assignment algorithm which exercises most of the capabilities offered by this proposal.


cpuguy83:
I'm not sure I like the idea of making this much more complex.
Today one can do stateful plugins by having the plugin call out to a daemon using whatever protocol they like.

That said, I keep wanting use NRI to do things and it doesn't have the information available to do what I need.

v2alpha1/pkg/api/api.proto
```diff
@@ -0,0 +1,472 @@
...
+  string name = 3;
+  ContainerState state = 4;
+  map<string, string> labels = 5;
+  map<string, string> annotations = 6;
+  repeated string args = 7;
+  repeated string env = 8;
+  repeated Mount mounts = 9;
+  Hooks hooks = 10;
+  LinuxContainer linux = 11;
+}
```
AkihiroSuda, line 290: Can we just pass OCI `config.json` as string here?
Requests to update the config can be represented as JSON patches

https://en.wikipedia.org/wiki/JSON_Patch

v2alpha1/pkg/api/optional.go
```diff
@@ -0,0 +1,311 @@
...
+   limitations under the License.
+*/
+
+package api
+
+import (
+	"os"
+)
+
+//
```
haircommander, line 23: nit this seens like an extra comment

v2alpha1/pkg/api/api.proto
klihub: Well, it would definitely be possible technically. The primary reason why we decided against it so far is the scope of changes we want to allow vs. the full OCI Spec scope. 

Since we want to restrict changes via NRI to a relatively small subset of the OCI Spec, it felt more natural (and easier) to bolt most of the restrictions into the protocol itself. IOW, to have the protocol itself already prevent 'syntactically' one from expressing most of the disallowed changes as opposed to have a protocol that 'syntactically' allows arbitrary changes which then need to be 'semantically' restricted with a bunch of extra code.

A minor/secondary reason is related to the 'API-ish' aspects of the protocol. Using JSON patches on the OCI Spec seems awkward as an API to me, but this is just my personal subjective opinion. This point also might be a bit moot, since we prefer to see the plugin stub as the primary API towards plugin writers instead of the low-level ttRPC protocol itself. If the low-level protocol was designed with a full OCI-Spec input/JSON patch output approach, this fact could be hidden from the user by/in the stub as well.

v2alpha1/pkg/api/optional.go
klihub: Removed extra comment.

mikebrow:
Note: with respect to heading to an rpc (ttrpc/grpc) see cni v2 discussion migrating to rpc https://github.com/containernetworking/cni/issues/785

klihub:
@AkihiroSuda @mikebrow @haircommander If you have some time, PTAL. I tried to address a bunch of the first review round comments: simplified the ttRPC protocol a bit, added a bunch of test cases, and fixed bugs found by those, cleaned up some of the most offending bits.

klihub:
> Some follow ups from the community meeting:
> 
> * We need to make a decision about the plugin model to support in containerd, whether we can generalize it for this use case using the shim model (@fuweid, @mikebrow )

NRI itself does not need much for plugins so I think anything should be fine for us, as long as plugins can
- be launched both by the runtime and by external means (in which case they connect to the runtime on their own accord and identify/register themselves, after which they should be mostly indistinguishable from runtime-launched plugins)
- be used in CRI-O and any other potential runtime implementations

> * Whether this can be integrated in containerd directly with the the sandbox model along with the CRI updates to break it into components (@mxpv)

fengwang666:
@klihub I'm interested in using nri to allocate node resources with Kata Containers. Is there any expectation when this will be merged to containerd?

klihub:
> @klihub I'm interested in using nri to allocate node resources with Kata Containers. Is there any expectation when this will be merged to containerd?

We have been making very little practical progress (as in terms of reviews/feedback) and slowly on this front. In the previous containerd community meeting this was briefly discussed (wrt. some other pending PRs and this one). One of the reasons is of course, that this is really a huge piece to chew/review. But another reason seems to be that the key upstream maintainers and contributors are still very busy working on their own stuff/turf so they can't spare cycles for reviewing big PRs. I *hope* that this situation will gradually ease during the fall/autumn as/if those key folks get their own stuff merged, freeing up more capacity for reviews.

That said, I really don't have any date estimates. My *hope* is that upstream involvement activity in reviews would gradually pick up, then we could start making real practical progress, and hopefully get a initial version in by the end of the year. Meanwhile one of the maintainers offered his help to shepherd me in this and I'm working on the next iterations of this and the containerd integration PRs, trying to clean up/improve anything I can.

If you are interested in this, then any and all feedback is more then welcome. Also if you are interested in writing an experimental plugin of your own to test and see if anything is broken or missing in the current version from your point of view, then I can help you get started with that.

v2alpha1/pkg/stub/stub.go
```diff
@@ -0,0 +1,724 @@
...
+// Name returns the full indexed name of the plugin.
+func (stub *stub) Name() string {
+	return stub.idx + "-" + stub.name
+}
+
+// Connect returns a connection to the NRI/runtime.
+func Connect() (stdnet.Conn, error) {
+	conn, err := ConnFromEnv()
+	if err == ErrNoConn {
+		conn, err = Dial(api.DefaultSocketPath)
```
askervin, line 403: Do you think that `Connect()` could use an environment variable to override `api.DefaultSocketPath`? Or maybe reuse `NRI_PLUGIN_SOCKET` with two different schemes. For instance:
```
NRI_PLUGIN_SOCKET=fd://42
NRI_PLUGIN_SOCKET=unix:///var/run/nri-test.sock
```

haosdent:
@klihub Awesome work!
A naive question, how to make k8s-scheduler aware of the changes from NRI? Via `labels, annotations?

klihub:
> @klihub Awesome work! A naive question, how to make k8s-scheduler aware of the changes from NRI? Via `labels, annotations?

@haosdent, @kad, @askervin
I am not sure what exactly you mean by that question so I can only guess. But in general, regardless of whether you experiment with new resource assignment algorithms as external to kubelet or as new kubelet *Manager policies, the algorithm you implement should not violate the general assumptions and therefore break the accounting calculation/book-keeping made by/in the scheduler.

I know that adhering to that principle then presents certain limitations for what your algorithms can do. Those limitations are already a source of problem for some of the existing policies/algorithms in the current *Managers' in kubelet. But this is IMO a related but entirely independent problem of its own in the current K8s infrastructure which anyway needs to be solved independently and regardless of what we are doing here. We do have some initial ideas about how to lift some of those restrictions with pretty simple changes.

Anyway, if you keep that restriction in mind then you don't need to make the scheduler aware of any changes, because the scheduler pre-calculates and administers the changes in resource availability when it picks and assigns a node for the workload.

Or did you have something else/something particular in mind ?

haosdent:
@klihub Thanks for your reply

> don't need to make the scheduler aware of any changes

I guess in some special cases, we may need to make k8s-scheduler aware of the changes in NRI. For example, NRI allocates NUMA for containers and makes the current node doesn't have insufficient CPU/mem in the same NUMA node although the total CPU/mem is enough.

So I was thinking if Kubernetes side could be aware of NRI changes via such mechanisms, e.g. via the annotation or some other fields in ContainerStatus.

klihub:
> @klihub Thanks for your reply
> 
> > don't need to make the scheduler aware of any changes
> 
> I guess in some special cases, we may need to make k8s-scheduler aware of the changes in NRI. For example, NRI allocates NUMA for containers and makes the current node doesn't have insufficient CPU/mem in the same NUMA node although the total CPU/mem is enough.
> 
> So I was thinking if Kubernetes side could be aware of NRI changes via such mechanisms, e.g. via the annotation or some other fields in ContainerStatus.

Yes, this is an existing and independent problem. AFAIK it also hits the existing in-kubelet policies when for instance configured to strict NUMA alignment mode. The core of the problem mostly comes from the scheduler's simplified view of a 'flat resource space'. Similar problems arise if you try to implement strict full physical core allocation (vs. a total of X hyperthreads).  And in many other cases when either the hierarchical nature of resources comes into the picture or when your resource model is such that the resources are not fully orthogonal (IOW allocation from one set of resources has an affect on the amount/availability of other resources and vice versa).

It will need to be solved regardless of what we do here. Sometimes you can do workarounds using extended resources, but often when you have a non-orthogonal resource model that is not enough.

haosdent:
> when you have a non-orthogonal resource model that is not enough.

@klihub True, thanks for the explanations. 👍

mikebrow (approved):
LGTM

README-v0.1.0.md
```diff
@@ -0,0 +1,115 @@
+# nri - Node Resource Interface
+
+[![PkgGoDev](https://pkg.go.dev/badge/github.com/containerd/nri)](https://pkg.go.dev/github.com/containerd/nri)
+[![Build Status](https://github.com/containerd/nri/workflows/CI/badge.svg)](https://github.com/containerd/nri/actions?query=workflow%3ACI)
+[![codecov](https://codecov.io/gh/containerd/nri/branch/main/graph/badge.svg)](https://codecov.io/gh/containerd/nri)
+[![Go Report Card](https://goreportcard.com/badge/github.com/containerd/nri)](https://goreportcard.com/report/github.com/containerd/nri)
+
+*This project is currently in DRAFT status*
```
mikebrow, line 8:
```suggestion
*This version of NRI is supported through the included v010-adapter plugin.*
```

README-v0.1.0.md
```diff
@@ -0,0 +1,115 @@
+# nri - Node Resource Interface
+
+[![PkgGoDev](https://pkg.go.dev/badge/github.com/containerd/nri)](https://pkg.go.dev/github.com/containerd/nri)
```
mikebrow, line 3: these project status widgets should be moved back to the readme

README.md
```diff
@@ -1,115 +1,306 @@
-# nri - Node Resource Interface
+## Node Resource Interface, Revisited
 
-[![PkgGoDev](https://pkg.go.dev/badge/github.com/containerd/nri)](https://pkg.go.dev/github.com/containerd/nri)
```
mikebrow, line 3: let's leave this section at the top level and leave the widgets here..

README.md
```diff
@@ -1,115 +1,306 @@
...
-[![Build Status](https://github.com/containerd/nri/workflows/CI/badge.svg)](https://github.com/containerd/nri/actions?query=workflow%3ACI)
-[![codecov](https://codecov.io/gh/containerd/nri/branch/main/graph/badge.svg)](https://codecov.io/gh/containerd/nri)
-[![Go Report Card](https://goreportcard.com/badge/github.com/containerd/nri)](https://goreportcard.com/report/github.com/containerd/nri)
+### Goal
 
-*This project is currently in DRAFT status*
```
mikebrow, line 8: let's leave the DRAFT notice here till we bump a release version

mikebrow (commented):
some nit comments on the readme files..

mikebrow:
looking very good just a couple nits on the readme.. another manual run against the now merged containerd/containerd main build with your new NRI service plugin and good to go I think.

README.md
klihub: Moved to README.md.

README.md
klihub: Moved back to README.md.

README-v0.1.0.md
klihub: Committed to/updated README-v0.1.0.md.

dmcgowan (approved):

mikebrow (approved):
LGTM
Thanks a ton this looks great.

pkg/api/mount.go
```diff
@@ -0,0 +1,89 @@
...
+import (
+	"sort"
+
+	rspec "github.com/opencontainers/runtime-spec/specs-go"
+)
+
+const (
+	// SELinuxRelabel is a Mount pseudo-option to request relabeling.
+	SELinuxRelabel = "relabel"
+)
```
AkihiroSuda, line 28: Is this constant  not consumed?

klihub: Hmm... no. Support for relabeling an injected mount (which was requested using that pseudo mount option) was dropped at one point, as it was unclear if it is needed. But I forgot to remove the related const. I'll file a PR for removing it.

PR #26, opened 2023-02-14
Make plugin timeouts configurable, adjust default socket path.

klihub:
Update defaults for plugin registration and request processing timeouts and make them configurable. This should help testing plugins with a complex startup/registration sequence in very slow CI environments where registration otherwise sometimes triggers a timeout.

Also adjust the default socket path to `/var/run/nri/nri.sock` and create the parent directory for the socket with `0700` access rights instead of `0755`. Having an intermediate directory makes it easier for plugins that run as a K8s daemonset to survive a runtime restart an reconnect to the NRI socket.

### DIFF
--- a/pkg/adaptation/plugin.go
+++ b/pkg/adaptation/plugin.go
@@ -36,8 +36,16 @@ import (
 )
 
 const (
-	pluginRegistrationTimeout = 2 * time.Second
-	pluginRequestTimeout      = 2 * time.Second
+	// DefaultPluginRegistrationTimeout is the default timeout for plugin registration.
+	DefaultPluginRegistrationTimeout = 5 * time.Second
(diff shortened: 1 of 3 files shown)


pkg/adaptation/plugin.go
```diff
@@ -59,6 +66,16 @@ type plugin struct {
 	r      *Adaptation
 }
 
+// SetPluginRegistrationTimeout sets the timeout for plugin registration.
+func SetPluginRegistrationTimeout(t time.Duration) {
```
fuweid, line 70: I think we should use lock for safy.

klihub: Updated with a read-write lock and getters.

fuweid (approved):

PR #28, opened 2023-02-22
fixes: fix resource adjustment to properly ignore unset/unadjusted fields.

klihub:
When adjusting CPU or memory resources, check for each field whether it is set (IOW adjusted by a plugin) before updating. The field getters return the corresponding golang zero value for unset fields. Therefore unconditionally adjusting unset Spec fields resets the fields instead. This manifests itself by all container CPU resources getting cleared when NRI is enabled but no plugins are running or none of the running plugins are adjust resources.

### DIFF
--- a/pkg/runtime-tools/generate/generate_suite_test.go
+++ b/pkg/runtime-tools/generate/generate_suite_test.go
@@ -1,0 +1,354 @@
+/*
+   Copyright The containerd Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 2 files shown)


mikebrow (approved):
LGTM

pkg/runtime-tools/generate/generate_suite_test.go
```diff
@@ -0,0 +1,354 @@
...
+			},
+		},
+	}
+	for _, o := range options {
+		o(spec)
+	}
+	return spec
+}
+
+func Int64(v int64) *int64 {
```
fuweid, line 348: I think we can use generic here, like 

```
func ToPtr[T any](v T) *T {
    return &v
}
```

Just share a idea.

fuweid (approved):

klihub:
@fuweid Could we tag the current HEAD as v0.3.0 ? Then I could update cobtsinerd #8140 to point at it.

I might have one more fix that we could try to sqeeze in, but it should not break backward compatibility, so I think that wouldn't require a tag (and we could always tag it as v0.3.1 if we prefer releases to be tag-aligned).

pkg/runtime-tools/generate/generate_suite_test.go
klihub: Thanks ! Good point and I'll include that change in a subsequent PR when adding more test cases to the suite.

PR #29, opened 2023-02-23
fixes: sort final adjusted mount slice.

klihub:
If mounts get adjusted, sort the final mount slice to make sure that any potential subdirectories are mounted only after their parents. Otherwise a parent mount could shadow/mask the content of a previously mounted subdirectory.

### DIFF
--- a/pkg/runtime-tools/generate/generate.go
+++ b/pkg/runtime-tools/generate/generate.go
@@ -18,6 +18,9 @@ package generate
 
 import (
 	"fmt"
+	"os"
+	"path/filepath"
+	"sort"
 	"strings"
 
 	rspec "github.com/opencontainers/runtime-spec/specs-go"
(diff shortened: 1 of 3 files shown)


pkg/runtime-tools/generate/generate.go
```diff
@@ -354,10 +358,60 @@ func (g *Generator) AdjustMounts(mounts []*nri.Mount) error {
...
+
+	g.Generator.Config.Mounts = mounts
+}
+
+// orderedMounts defines how to sort an OCI Spec Mount slice.
+// This is the almost the same implementation sa used by CRI-O and Docker,
+// with a minor tweak for stable sorting order (easier to test):
+//
+//	https://github.com/moby/moby/blob/17.05.x/daemon/volumes.go#L26
+type orderedMounts []rspec.Mount
```
fuweid, line 384: It't not blocked issue. Just thinking that we can export it into utils pkg, instead of copy-parse.
Handle it in the follow-up.

fuweid (approved):

fuweid:
@klihub is it ok to create v0.3.0 tag after this merged?

klihub:
@fuweid Yes, let's tag it. I have no other PRs ATM in the pipeline.

fuweid:
> @fuweid Yes, let's tag it. I have no other PRs ATM in the pipeline.
> 
> 

nice! I will create tag in two days.

Issue #33, opened 2023-03-10
Need more details on config paths and formats for hook-injector


swgriffith:
I've been able to get the hook injector working on a cluster with containerd 1.7-rc3, both by running the hook-injector directly and by dropping it in the /opt/nri/plugins directory, however I'm having issues working out where the hook-injector config files should be placed. I'm following the details shared here: 

https://github.com/containerd/nri/tree/main/plugins/hook-injector

I'll keep testing, but any guidance would be apprciated.

swgriffith:
I worked it out, and I'm happy to PR some added details, unless there's a larger documentation refresh already in progress.

klihub:
I assume that by 'hook injector config files' you refer to the actual hook schema files et. al, which the hook manager uses to decide whether and how to inject any hooks into a container ? If you have fixes for the docs then any PR is welcome.

Could you also include a fix for the outdated/broken hook manager link while at it ? I just noticed when trying to follow the link above through the injector to the hook manager that it has apparently been split out ages ago to `https://github.com/containers/common/tree/main/pkg/hooks`. Our link neither points to the exact version we use (commit 8bcc086b1b9d) nor to an up-to-date split out version. We should point our link to the exact sha1 version we use.

mikebrow:
Please reopen if you believe more should be documented here.  Thx!

PR #34, opened 2023-03-14
Updated the OCI Hook Injector README to resovle broken links to the p…

swgriffith:
Resolves issue #33 

Updates the links to podman documentation to reference the proper sha1 and adds additional details to help guide people in testing the hook-injector plugin, including relevant path details.

### DIFF
--- a/plugins/hook-injector/README.md
+++ b/plugins/hook-injector/README.md
@@ -4,7 +4,7 @@ The [OCI runtime configuration](https://github.com/opencontainers/runtime-spec/b
 
 ## Testing
 
-You can test this plugin using a Kubernetes cluster/node with a container runtime that has NRI support enabled ([Enabling NRI in Containerd](https://github.com/containerd/containerd/blob/main/docs/NRI.md#enabling-nri-support-in-containerd)). Once you've enabled NRI on your runtime, you can use the [sample hook configuration](etc/containers/oci/hooks.d), placing it at `/etc/containers/oci/hooks.d`, and the [sample hook](usr/local/sbin/demo-hook.sh), placing it at `/usr/local/sbin/`.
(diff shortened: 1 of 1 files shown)


klihub (approved):
Looks good to me. Thanks for the fixes !

klihub:
> Maybe it would be more sense to point to top of the tree like https://github.com/containers/common/tree/main/pkg/hooks and https://github.com/containers/common/blob/main/pkg/hooks/docs/oci-hooks.5.md instead of some older commit id in podman repo?

We deliberately chose to point to the exact version we use in the plugin. That is then a different story that we should really file another PR to update the plugin to use a recent version, and then (remember to) update the link in the same PR as well.

swgriffith:
@klihub and @kad - It sounds like we're leaving it as is. Let me know if you need me to change the URLs.

kad:
I'm ok with @klihub opinion, so lgtm.

plugins/hook-injector/README.md
```diff
@@ -1,21 +1,15 @@
...
 
 ## Testing
 
-You can test this plugin using a kubernetes cluster/node with a container
-runtime that has NRI support enabled. Start the plugin on the target node
-(`hook-injector -idx 10`), put a suitable [OCI hook configuration](https://github.com/containers/podman/blob/main/pkg/hooks/docs/oci-hooks.5.md)
-in place, then create a pod with matching containers and check the results.
+You can test this plugin using a Kubernetes cluster/node with a container runtime that has NRI support enabled ([Enabling NRI in Containerd](https://github.com/containerd/containerd/blob/main/docs/NRI.md#enabling-nri-support-in-containerd)). Once you've enabled NRI on your runtime, you can use the [sample hook configuration](etc/containers/oci/hooks.d), placing it at `/etc/containers/oci/hooks.d`, and the [sample hook](usr/local/sbin/demo-hook.sh), placing it at `/usr/local/sbin/`.
```
fuweid, line 7: could you please remove the `[sample hook configuration](etc/containers/oci/hooks.d)` and just use `sample hook configuration` or link to real path? Thanks

swgriffith: Sorry for the delay on this fix. I've updated.

mikebrow (approved):
/LGTM thanks!

PR #36, opened 2023-04-06
docs: add a chapter about security considerations.

klihub:
Add a few paragraphs with security considerations about protecting access to the NRI socket and deploying NRI plugins as K8s DaemonSets.

### DIFF
--- a/README.md
+++ b/README.md
@@ -312,6 +312,48 @@ The following sample plugins exist for NRI:
 Please see the documentation of these plugins for further details
 about what and how each of these plugins can be used for.
 
+## Security Considerations
+
+From a security perspective NRI plugins should be considered part of the
+container runtime. NRI does not implement granular access control to the
+functionality it offers. Access to NRI is controlled by restricting access
+to the systemwide NRI socket. If a process can connect to the NRI socket
+and send data, it has access to the full scope of functionality available
+via NRI.
+
+In particular this includes
+
+  - injection of OCI hooks, which allow for arbitrary execution of processes with the same privilege level as the container runtime
+  - arbitrary changes to mounts, including new bind-mounts, changes to the proc, sys, mqueue, shm, and tmpfs mounts
+  - the addition or removal of arbitrary devices
+  - arbitrary changes to the limits for memory, CPU, block I/O, and RDT resources available, including the ability to deny service by setting limits very low
+
+The same precautions and principles apply to protecting the NRI socket as
+to protecting the socket of the runtime itself. Unless it already exists,
+NRI itself creates the directory to hold its socket with permissions that
+allow access only for the user ID of the runtime process. By default this
+limits NRI access to processes running as root (UID 0). Changing the default
+socket permissions is strongly advised against. Enabling more permissive
+access control to NRI should never be done without fully understanding the
+full implications and potential consequences to container security.
+
+### Plugins as Kubernetes DaemonSets
+
+When the runtime manages pods and containers in a Kubernetes cluster, it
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -312,6 +312,38 @@ The following sample plugins exist for NRI:
...
 about what and how each of these plugins can be used for.
 
+## Security Considerations
+
+From a security perspective NRI plugins should be considered part of the
+container runtime. NRI does not implement granular access control to the
+functionality it offers. Access to NRI is controlled by restricting access
+to the systemwide NRI socket. If a process can connect to the NRI socket
+and send data, it has access to the full scope of functionality available
+via NRI.
```
samuelkarp, line 322: I think since this is discussing security, it's better to err on the side of being more verbose and explicit.  Assuming that someone hasn't read enough of the readme to comprehend the "full scope of functionality", we can spell it out a bit.  Here's a possible set of things to describe:

* injection of OCI hooks, which allow for arbitrary execution of processes with the same privilege level as the container runtime
* arbitrary changes to mounts, including new bind-mounts, changes to the proc, sys, mqueue, shm, and tmpfs mounts
* the addition or removal of arbitrary devices
* arbitrary changes to the limits for memory, CPU, block I/O, and (on Intel systems) RDT resources available, including the ability to deny service by setting limits very low

README.md
```diff
@@ -312,6 +312,38 @@ The following sample plugins exist for NRI:
...
+
+The same precautions and principles apply to protecting the NRI socket as
+to protecting the socket of the runtime itself. Unless it already exists,
```
cji, line 325: Is there an existing list of these precautions we provide for the runtime socket that we could link to or summarize here? Or can we be more explicit about how we recommend protecting and restricting access to the socket? 
e.g. do not change the default file permissions.

README.md
```diff
@@ -312,6 +312,38 @@ The following sample plugins exist for NRI:
...
+NRI itself creates the directory to hold its socket with permissions that
+allow access only for the user ID of the runtime process. In practice this
+limits NRI access to processes running as root (UID 0).
```
cji, line 328: How do you feel about changing "In practice" to "By default"? I think we'd want to recommend here that users don't change the default ownership or permissions of the directory or socket file.

klihub: Thanks ! I'll update the proposed additions about security consideration with this (and the above comments).

README.md
klihub: Updated, spelling out explicitly all of the above mentioned possibilities.

README.md
klihub: > Is there an existing list of these precautions we provide for the runtime socket that we could link to or summarize here?

No, at least based on quick git grep I could not find anything relevant in the in-tree containerd documentation.

> Or can we be more explicit about how we recommend protecting and restricting access to the socket? e.g. do not change the default file permissions.

So, should we explicitly discourage changing the socket permissions ? State something like 

```
Changing the default socket permissions is strongly advised against. Enabling more permissive access control to NRI should never be done without fully understanding the full implications and potential consequences to container security.
```

mikebrow (approved):
LGTM thanks for putting this together!

README.md
samuelkarp: That wording looks good to me.

kad (approved):
/lgtm

cji (approved):
LGTM thanks for writing this! 🎉 🎉 🎉

samuelkarp (approved):
LGTM, thanks for writing this!

PR #37, opened 2023-05-04
api: initialize OCI LinuxMemory resources to empty.

klihub:
When converting to OCI LinuxResources, always use an empty `&LinuxMemory{}` instead of a `nil` one. Some runtimes try to dereference it unconditionally. Leaving it unset panics on these.


mikebrow (approved):
LGTM

samuelkarp (approved):

PR #40, opened 2023-05-17
stub: pass context to plugins, pass updated resources to UpdateContainers.

klihub:
Don't hide the ttRPC request context from plugins. Pass the context on to the plugin's request handler. Once we have opentelemetry ttRPC instrumentation support, this should allow us to properly propagate trace spans and have detailed instrumentation on the plugins side properly nested into the related original request on the runtime side.

While making API-breaking changes, also fix UpdateContainer adding an argument with the updated resources.

Update sample plugins and tests with the updated interfaces and add basic tests for UpdateContainer which we lacked so far.

### DIFF
--- a/pkg/adaptation/suite_test.go
+++ b/pkg/adaptation/suite_test.go
@@ -470,13 +470,13 @@ func (m *mockPlugin) onClose() {
 	}
 }
 
-func (m *mockPlugin) Configure(cfg, runtime, version string) (stub.EventMask, error) {
+func (m *mockPlugin) Configure(_ context.Context, cfg, runtime, version string) (stub.EventMask, error) {
 	m.q.Add(PluginConfigured)
 
 	return m.mask, nil
 }
 
-func (m *mockPlugin) Synchronize(pods []*api.PodSandbox, ctrs []*api.Container) ([]*api.ContainerUpdate, error) {
+func (m *mockPlugin) Synchronize(_ context.Context, pods []*api.PodSandbox, ctrs []*api.Container) ([]*api.ContainerUpdate, error) {
 	for _, pod := range pods {
 		m.pods[pod.Id] = pod
 	}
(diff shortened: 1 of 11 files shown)


plugins/differ/nri-differ.go
```diff
@@ -177,74 +177,74 @@ func (p *plugin) differ(apifunc string, pod *api.PodSandbox, container *api.Cont
...
 }
 
-func (p *plugin) PostStartContainer(pod *api.PodSandbox, container *api.Container) error {
+func (p *plugin) PostStartContainer(_ context.Context, pod *api.PodSandbox, container *api.Container) error {
 	p.differ("PostStartContainer", pod, container)
 	return nil
 }
 
-func (p *plugin) UpdateContainer(pod *api.PodSandbox, container *api.Container) ([]*api.ContainerUpdate, error) {
+func (p *plugin) UpdateContainer(_ context.Context, pod *api.PodSandbox, container *api.Container, r *api.LinuxResources) ([]*api.ContainerUpdate, error) {
```
mikebrow, line 230: might want to split the LinuxResources change from the ctx change..

pkg/adaptation/suite_test.go
```diff
@@ -470,13 +470,13 @@ func (m *mockPlugin) onClose() {
 	}
 }
 
-func (m *mockPlugin) Configure(cfg, runtime, version string) (stub.EventMask, error) {
+func (m *mockPlugin) Configure(_ context.Context, cfg, runtime, version string) (stub.EventMask, error) {
```
mikebrow, line 473: why the undeclared param.. for ctx?

mikebrow (commented):
couple thoughts

pkg/stub/stub.go
```diff
@@ -764,3 +767,129 @@ func (stub *stub) setupHandlers() error {
 
 	return nil
 }
+
+type origConfigureInterface interface {
```
mikebrow, line 771: deprecations?

pkg/adaptation/suite_test.go
klihub: mockPlugin never uses or passes the context on to any test-case specific handlers.

plugins/differ/nri-differ.go
klihub: I split out the update/resource-related changes to two additional separate commits: one to add the missing resource parameter to the stub plugin UpdateContainer handler, and another to update the mocked test plugins and to implement basic tests for UpdateContainers.

@mikebrow I'll file those in a separate PR. I interpreted it as your preference from the above comment, as opposed to having them in separate commits but part of the same PR.

pkg/stub/stub.go
klihub: Do you mean that I should make these old interfaces all exported (OrigConfigureInterface, etc) and mark them deprecated in their docstring ?

plugins/differ/nri-differ.go
mikebrow: splitting the commit is fine.. just need to have a different description in the PR for that additional change..

pkg/stub/stub.go
mikebrow: I was reading through the code trying to gronk what/if any plan there was for deprecating the pre-context injected versions of the api

- new release, no deprecations
- new api version, with deprecations
- ...

mikebrow: I mean I can see the code rejects.. :-)

klihub: > I mean I can see the code rejects old handlers.. :-)

Ah, ok that. I think we haven't declared the APIs stable yet. So I think in principle it should be still okay to make API breaking changes. We kept those old interfaces as unexported ones prefixed with `orig`, just to be able to test for them and bail out with an error message if a plugin implements old interfaces.

I don't know what would be the best option here. 1) Don't test for old interfaces and let plugin creation fail with something like `plugin XYZ does not implement any NRI request handlers`. 2) Do what we do now. 3) Mark the interfaces deprecated and still accept/handle plugins that implement old interfaces. Is that extra trouble worth it at this point, yet ?

BTW, once we merge this I'll need to go and file another update to address the missing resources parameter from UpdateContainers, which is another API-breaking change.

klihub: Okay. Based on our discussion, I ditched all the tests for old interfaces. Plugins anyway need to be updated and if they are not, the will fail to start up with a `plugin XYZ does not implement any NRI request handlers` error message.

I updated the documentation with a note about the current lack of guaranteed API stability before v1.0.0. I gave a guesstimate for a v1.0.0 release. Finally, I added a brief release notes file which lists the changes so far between each tagged release, and for the pending untagged v0.4.0.

mikebrow (commented):
nit.. commit header for fix plugin UpdateContainerInterface also states 
"Refuse instantiating a Stub for any plugin
that implements the old interface."

mikebrow (approved):
/LGTM

mikebrow:
***: requires version tag, and container runtime release notes (containerd/crio) for breaking api change when they vendor the next tag

fuweid (approved):
LGTM

PR #43, opened 2023-05-18
sync sandboxes and containers after starting the pre-installed plugins

Iceber:
For plugins registered with nri.sock, NRI will send the full set of pods and containers after starting.
https://github.com/containerd/nri/blob/83295992ed466e5589a1a83ba3ad2bd6b6feb199/pkg/adaptation/adaptation.go#L405-L430

For pre-installed plugins, it is still necessary to synchronize the sandboxes and containers data to the plugin after launch，otherwise the plugin will never be able to get information about existing resources

### DIFF
--- a/pkg/adaptation/adaptation.go
+++ b/pkg/adaptation/adaptation.go
@@ -346,10 +346,25 @@ func (r *Adaptation) startPlugins() (retErr error) {
 		}
 
 		if err := p.start(r.name, r.version); err != nil {
-			return err
+			return fmt.Errorf("failed to start NRI Plugin %q: %w", name, err)
 		}
 
 		plugins = append(plugins, p)
+	}
+
+	syncPlugins := func(ctx context.Context, pods []*PodSandbox, containers []*Container) ([]*ContainerUpdate, error) {
+		var updates []*ContainerUpdate
+		for _, plugin := range plugins {
+			us, err := plugin.synchronize(ctx, pods, containers)
+			if err != nil {
+				return nil, fmt.Errorf("failed to sync NRI Plugin %q: %w", plugin.name(), err)
+			}
+			updates = append(updates, us...)
+		}
+		return updates, nil
+	}
+	if err := r.syncFn(noCtx, syncPlugins); err != nil {
+		return fmt.Errorf("failed to synchronize NRI Plugins: %w", err)
 	}
 
 	r.plugins = plugins
--- a/pkg/adaptation/adaptation.go
+++ b/pkg/adaptation/adaptation.go
@@ -336,40 +336,48 @@ func (r *Adaptation) startPlugins() (retErr error) {
 	}()
 
 	for i, name := range names {
-		log.Infof(noCtx, "starting plugin %q...", name)
+		log.Infof(noCtx, "starting pre-installed NRI plugin %q...", name)
 
 		id := ids[i]
-
 		p, err := r.newLaunchedPlugin(r.pluginPath, id, name, configs[i])
 		if err != nil {
-			return fmt.Errorf("failed to start NRI plugin %q: %w", name, err)
+			log.Warnf(noCtx, "failed to initialize pre-installed NRI plugin %q: %v", name, err)
+			continue
 		}
 
 		if err := p.start(r.name, r.version); err != nil {
-			return fmt.Errorf("failed to start NRI Plugin %q: %w", name, err)
+			log.Warnf(noCtx, "failed to start pre-installed NRI plugin %q: %v", name, err)
+			continue
 		}
 
 		plugins = append(plugins, p)
 	}
 
-	syncPlugins := func(ctx context.Context, pods []*PodSandbox, containers []*Container) ([]*ContainerUpdate, error) {
-		var updates []*ContainerUpdate
-		for _, plugin := range plugins {
+	// Although the error returned by syncPlugins may not be nil, r.syncFn could still ignores this error and returns a nil error.
+	// We need to make sure that the plugins are successfully synchronized in the `plugins`
+	syncPlugins := func(ctx context.Context, pods []*PodSandbox, containers []*Container) (updates []*ContainerUpdate, err error) {
+		startedPlugins := plugins
+		plugins = make([]*plugin, 0, len(plugins))
+		for _, plugin := range startedPlugins {
 			us, err := plugin.synchronize(ctx, pods, containers)
 			if err != nil {
-				return nil, fmt.Errorf("failed to sync NRI Plugin %q: %w", plugin.name(), err)
+				plugin.stop()
+				log.Warnf(noCtx, "failed to synchronize pre-installed NRI plugin %q: %v", plugin.name(), err)
+				continue
 			}
+
+			plugins = append(plugins, plugin)
 			updates = append(updates, us...)
+			log.Infof(noCtx, "pre-installed NRI plugin %q synchronization success", plugin.name())
 		}
 		return updates, nil
 	}
 	if err := r.syncFn(noCtx, syncPlugins); err != nil {
-		return fmt.Errorf("failed to synchronize NRI Plugins: %w", err)
+		return fmt.Errorf("failed to synchronize pre-installed NRI Plugins: %w", err)
 	}
 
 	r.plugins = plugins
 	r.sortPlugins()
-
 	return nil
 }
 


klihub (approved):
Thanks for the fix ! LGTM.

Iceber:
@klihub I changed the call to syncFn after each pre-install plugin launch to only call syncFn once after all plugins are launched, which avoids unnecessary multiple calls to syncFn in the case of a large number of plugins
https://github.com/containerd/nri/compare/104c8e844981be230d53ffac4f9eee5d1db0634e..6f7af44d11899f94a6f619bab0bba6b012bd8e4e

PTAL, Thanks

klihub:
> @klihub I changed the call to syncFn after each pre-install plugin launch to only call syncFn once after all plugins are launched, which avoids unnecessary multiple calls to syncFn in the case of a large number of plugins
> https://github.com/containerd/nri/compare/104c8e844981be230d53ffac4f9eee5d1db0634e..6f7af44d11899f94a6f619bab0bba6b012bd8e4e
> 
> PTAL, Thanks
> 

True, better to do it that way.
LGTM.

klihub (approved):

pkg/adaptation/adaptation.go
```diff
@@ -334,12 +334,27 @@ func (r *Adaptation) startPlugins() (retErr error) {
...
 
 		plugins = append(plugins, p)
 	}
 
+	syncPlugins := func(ctx context.Context, pods []*PodSandbox, containers []*Container) ([]*ContainerUpdate, error) {
+		var updates []*ContainerUpdate
+		for _, plugin := range plugins {
+			us, err := plugin.synchronize(ctx, pods, containers)
+			if err != nil {
+				return nil, fmt.Errorf("failed to sync NRI Plugin %q: %w", plugin.name(), err)
```
mikebrow, line 348: should this be a warning instead and continue?

pkg/adaptation/adaptation.go
```diff
@@ -334,12 +334,27 @@ func (r *Adaptation) startPlugins() (retErr error) {
...
+			}
+			updates = append(updates, us...)
+		}
+		return updates, nil
+	}
+	if err := r.syncFn(noCtx, syncPlugins); err != nil {
```
mikebrow, line 354: do we need to be in r.lock here..

mikebrow (commented):
see questions

pkg/adaptation/adaptation.go
Iceber: There are two possible approaches to consider: skipping the synchronization if it fails or returning an error.

I chose to return an error for the following reason: `pre-install plugins` and `external plugins` are different in some ways. 
The `pre-install plugin binary` is launched by the NRI adaptation, so it is difficult for users to know whether the binary plugin has been launched successfully, apart from the logs. 
Therefore, users may overlook the fact that the `plugin binary` is not actually working. That's why I require that all `pre-install plugins` requested by the user must be launched(started, synced) successfully. 
This is similar to starting and returning an error if it fails.
https://github.com/containerd/nri/blob/fa3ab8fbf3c5e58f652a7a54f750f9b6a02416f7/pkg/adaptation/adaptation.go#L336-L338

On the other hand, external plugins are usually registered by external binaries, and we could perceive more clearly whether the NRI registration and synchronization are successful or not.

pkg/adaptation/adaptation.go
Iceber: Already locked in `Start`
https://github.com/containerd/nri/blob/fa3ab8fbf3c5e58f652a7a54f750f9b6a02416f7/pkg/adaptation/adaptation.go#L140-L150

pkg/adaptation/adaptation.go
mikebrow: I suppose two other patterns would be:
3)  to introduce config/management for requiring/ordering (add requires logic), auto restarting etc.. for plugin start && synch
4) attempt to start and synch all and report all errors

that said should we stop any that we started if there's an error..

Iceber: > 3) to introduce config/management for requiring/ordering (add requires logic), auto restarting etc.. for plugin start && synch

Could you introduce more details? thanks

mikebrow: current containerd plugins have a requires section for specifying which must plugins be init'd.. before initing a plugin

we probably need an additional state beyond init for the container plugin that would cover ready for work.. for example in the cri case ready may mean all containers/pods restarted ready for grpc calls

nri might need a requires list for the plugins (unless we go with external management via systemd? or whatnot) and then within that you could have the concept of ready which might at least cover init and sync ^  

additionally we might want to have a flag in the plugin config regarding if sync is required for "ready"

mikebrow: @dmcgowan thoughts?

champtar:
Can we merge this as is ? It does fix a blocking bug

mikebrow:
@klihub let's chat with @Iceber if still interested .. maybe at the next or subsequent containerd call

Iceber:
@mikebrow @klihub Do you have any new ideas for this pr?

I think synchronizing all and reporting back all errors is a good way which gives a more complete picture of the errors that may occur

Of course if still think that skipping and printing warnings is a good way to go, I'll adopt it

Also for more complex configurations my thoughts are:
1. pre-installed binaries are required by default, I don't think it's necessary for the user to place a bunch of plugin binaries in the directory that don't care if they are successfully launched or not
2. you can enable/disable some pre-installed plugins by configuring the plugin name, so that only some plugins can be started, or some plugins won't be started at all
I think this feature can be placed in another pr


ci fails from `errors.Join`, It comes from 1.20, we might consider upgrading go.mod to 1.20

klihub:
> ci fails from `errors.Join`, It comes from 1.20, we might consider upgrading go.mod to 1.20

I [filed PR #88](https://github.com/containerd/nri/pull/88) to bump the minimum golang requirement to 1.20. It should be fine for all of our downstream packages.

klihub:
> Hi @mikebrow @klihub Do you have any new ideas for this pr?
> 
> I think synchronizing all and reporting back all errors is a good way which gives a more complete picture of the errors that may occur

I think synchronizing pre-launched plugins (pre-installed plugins in your terminology) would be essential to make sure that we have an identical handshake/initial message flow, regardless of whether a plugin is pre-launched or not. There are packages out there which do expect the synchronization message to come before they consider themselves to be ready to start processing other NRI messages. I made a mistake and that is why I completely overlooked that pre-launched plugins do not get properly synchronized during registration. It was not intentional.

That said, I think synchronizing pre-launched plugins is only/really relevant for the case when an already active runtime (with existing containers) gets restarted (after all, on initial startup the set of pods and containers to sync with the plugins is empty). So a related question is whether the runtime itself is able to properly determine/rediscover the state of pods and containers by the time pre-launched plugins are being started and what to do if it is not. And this needs to be checked/tested both with containerd and cri-o because they might behave differently. I have a vague recollection that containerd itself would behave slightly differently for 2.0 than 1.7, but I haven't had time to check this now.

Iceber:
> That said, I think synchronizing pre-launched plugins is only/really relevant for the case when an already active runtime (with existing containers) gets restarted (after all, on initial startup the set of pods and containers to sync with the plugins is empty). So a related question is whether the runtime itself is able to properly determine/rediscover the state of pods and containers by the time pre-launched plugins are being started and what to do if it is not. And this needs to be checked/tested both with containerd and cri-o because they might behave differently. I have a vague recollection that containerd itself would behave slightly differently for 2.0 than 1.7, but I haven't had time right no to check/test this in practice.

yeah, this is very important, and can be very bad if the synchronization is incomplete with pods and containers.

In containerd 2.0, the `recover` function initializes the data before the nri is initialized.
https://github.com/containerd/containerd/blob/45bc430dd12539ee93f908750ce24e52e1122e9a/internal/cri/server/service.go#L257-L300

pkg/adaptation/adaptation.go
```diff
@@ -346,15 +346,40 @@ func (r *Adaptation) startPlugins() (retErr error) {
 		}
 
 		if err := p.start(r.name, r.version); err != nil {
-			return err
+			return fmt.Errorf("failed to start NRI Plugin %q: %w", name, err)
```
mikebrow, line 349: let's be more explicit in the logs for this start loop... by adding pre-installed... e.g. "starting pre-installed plugins..." at the beginning of this func..

mikebrow: to be consistent with the sync changes below these new and start errors should also "continue"

"failed to initialize pre-installed NRI plugin %q: %w"
"failed to start pre-installed NRI plugin %q: %w"

pkg/adaptation/adaptation.go
```diff
@@ -346,15 +346,40 @@ func (r *Adaptation) startPlugins() (retErr error) {
...
+				errs = append(errs, fmt.Errorf("[%s] %w", plugin.name(), err))
+			} else {
+				plugins = append(plugins, plugin)
+				updates = append(updates, us...)
+			}
+		}
+		return updates, errors.Join(errs...)
+	}
+	if err := r.syncFn(noCtx, syncPlugins); err != nil {
+		return fmt.Errorf("failed to synchronize NRI Plugins: %w", err)
```
mikebrow, line 378:
```suggestion
		return fmt.Errorf("failed to synchronize the following pre-installed NRI plugins: %w", err)
```

pkg/adaptation/adaptation.go
```diff
@@ -346,15 +346,40 @@ func (r *Adaptation) startPlugins() (retErr error) {
...
+		startedPlugins := plugins
+		plugins = make([]*plugin, 0, len(plugins))
+		for _, plugin := range startedPlugins {
+			us, err := plugin.synchronize(ctx, pods, containers)
+			if err != nil {
+				plugin.stop()
+				errs = append(errs, fmt.Errorf("[%s] %w", plugin.name(), err))
+			} else {
+				plugins = append(plugins, plugin)
+				updates = append(updates, us...)
```
mikebrow, line 372: I suppose for now, since we are going to continue on a single failure etc.. let's output the successes here

log.Infof(noCtx, "pre-installed plugin synchronization success %q", plugin.name())

mikebrow (changes requested):
I like this change just some comments to increase logging to help the plugin admin

** In 1.1 let's consider a more formal process with a manager plugin or 3) introduce config/management for requiring/ordering (add requires logic), auto restarting etc.. for plugin start && synch

mikebrow:
@Iceber wdyt about the logging changes..

Iceber:
> @Iceber wdyt about the logging changes..

Sorry I missed the email on this pr, I'll take care of it soon

pkg/adaptation/adaptation.go
Iceber: The meaning of “the following pre-installed NRI plugins” is bound to the errors returned by syncPlugins.

Although the errors returned by syncPlugins **should be** returned in r.syncFn, there is no way to guarantee that syncFn will correctly return error strings that match the meaning of “the following pre-installed NRI plugins” when used by the user. 

Should we be able to continue with a more generic description, such as `fmt.Errorf(“failed to synchronize pre-installed NRI plugins: %w”, err)`?

Iceber:
@mikebrow I've added a new commit for better viewing of the changes

I've returned plugins that failed to start to syncFn via syncPlugins. after all, those plugins that didn't start are also sync failures/unsyncable plugins

mikebrow (approved):
LGTM

Thanks!

mikebrow:
@klihub did you want to take one more look with the additional changes?

klihub:
Yes, I'll do that.

klihub:
> Yes, I'll do that.

It'll take me a bit more time though. I'm not at the keyboard ATM.

pkg/adaptation/adaptation.go
```diff
@@ -335,26 +335,53 @@ func (r *Adaptation) startPlugins() (retErr error) {
...
-		log.Infof(noCtx, "starting plugin %q...", name)
+		log.Infof(noCtx, "starting pre-installed NRI plugin %q...", name)
 
 		id := ids[i]
-
 		p, err := r.newLaunchedPlugin(r.pluginPath, id, name, configs[i])
 		if err != nil {
-			return fmt.Errorf("failed to start NRI plugin %q: %w", name, err)
+			errs = append(errs, fmt.Errorf("[%s] %w", name, err))
+			log.Warnf(noCtx, "failed to initialize pre-installed NRI plugin %q: %w", name, err)
```
klihub, line 346: With `log.Warnf()` we should use `%v` (value in default format) instead of `%w` (wrap error, but only understood by `fmt.Errorf()`).

pkg/adaptation/adaptation.go
```diff
@@ -335,26 +335,53 @@ func (r *Adaptation) startPlugins() (retErr error) {
...
+	// We need to make sure that the plugins are successfully synchronized in the `plugins`
+	syncPlugins := func(ctx context.Context, pods []*PodSandbox, containers []*Container) (updates []*ContainerUpdate, err error) {
+		startedPlugins := plugins
+		plugins = make([]*plugin, 0, len(plugins))
+		for _, plugin := range startedPlugins {
+			us, err := plugin.synchronize(ctx, pods, containers)
+			if err != nil {
+				plugin.stop()
+				errs = append(errs, fmt.Errorf("[%s] %w", plugin.name(), err))
+				log.Warnf(noCtx, "failed to synchronize pre-installed NRI plugin %q: %w", plugin.name(), err)
```
klihub, line 369: Same here, `log.*f()` should use `%v`, not `%w`.

klihub (changes requested):
It looks pretty good. I spotted a few wrongly formatted log messages (error wrapping vs. default value format %verb), and one oddly returned (value, error) combo.

Then I have a question to @mikebrow about what we should do if a pre-installed plugin failed to start up or get synchronized. The original code errors out and this one keeps that behavior, but now I started to have second thoughts about that. Do we really want to prevent the runtime from starting up in this case, or just log an error, ignore the failed plugins and keep going ?

If we get the few log formatting problems fixed, maybe also change that odd return value combo to the standard convention of 'return a nil-value with a non-nil error', then I think this is good to go in.

And if @mikebrow is of the opinion that we should err on the side of caution and ignore failing plugins instead of preventing the runtime from starting up, then we should also change that behavior.

klihub:
ping @Iceber Apart from a few small nits it looks very good.

pkg/adaptation/adaptation.go
```diff
@@ -335,26 +335,53 @@ func (r *Adaptation) startPlugins() (retErr error) {
...
+				continue
+			}
+
+			plugins = append(plugins, plugin)
+			updates = append(updates, us...)
+			log.Infof(noCtx, "pre-installed NRI plugin %q synchronization success", plugin.name())
+		}
+		return updates, errors.Join(errs...)
```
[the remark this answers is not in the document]
Iceber: If I understand @mikebrow correctly, syncPlugins returns successful updates and failed plugins&errors, and it's up to syncFn to decide whether or not to ignore a particular wrong plugin

klihub:
> In r.next 1.1 I'd like us to consider adding plugin config to describe ordering(after:plugin)/dependencies(requires:plugin)/runtime required(MUST) type information that we would use to order the plugin bring up and whether or not a plugin is required to run a container/pod which would give us a good reason, for example, to not report CRI plugin success and thus block pods on the node. Until then I'd prefer to log errors and continue for failing plugins vs bring the runtime down when there is an unknown / unhandled issue with a plugin.
> 
> We simply can't know if the plugin is required always or if it is only required for certain nodes or if the error is related to some unknown resource device dependency that is only met on some of the worker nodes. This gives the admin the ability to test out deployment of new plugins on a running machine, remove the failing ones report logs etc.

Sounds like a good plan to me.

@Iceber Could you update the PR to still log errors from plugin startup or synchronization failures, but return no error so we don't prevent the runtime from starting up.

Iceber:
@mikebrow @klihub updated to print logs only, PTAL, Thanks

klihub (approved):
LGTM. Thank you @Iceber !

PR #46, opened 2023-08-07
Take pkg/hooks from github.com/containers/common

eclipseo:
pkg/hooks was moved from github.com/containers/podman/v4 to github.com/containers/common and it's to much time consuming to keep packaging github.com/containers/podman/v3.


mikebrow:
Pls sign commits with real name and email.  Cheers!

plugins/hook-injector/hook-injector.go
```diff
@@ -30,7 +30,7 @@ import (
 
 	"github.com/containerd/nri/pkg/api"
 	"github.com/containerd/nri/pkg/stub"
-	"github.com/containers/podman/v3/pkg/hooks"
+	hooks "github.com/containers/common/pkg/hooks/1.0.0"
```
klihub, line 33: AFAICT this does not compile if imported like this. I think it should be without the final 'tag', simply as

```
        hooks "github.com/containers/common/pkg/hooks"
```

We're mimicking [cri-o](https://github.com/cri-o/cri-o/blob/2faaecdaf9e3fbbcfe5f640774b610218f6175eb/internal/lib/container_server.go#L11) here.

klihub (changes requested):
Please fix the new import path and update your commit (to be signed) according to @mikebrow 's instructions.

mikebrow:
@eclipseo wave..

plugins/hook-injector/hook-injector.go
eclipseo: Ok done that.

eclipseo:
Ok fixed the import path and signed the commit.

thaJeztah:
I gave it a quick go to carry the change in https://github.com/containerd/nri/pull/55

changes compared to this PR;

rebase (we merged the "remove containerd dependency"
removed changes from other plugin go mods (keeping the PR focussed on just this change)
updated to containers/common v0.56.0 (which uses runtime spec v1.1.0 instead of -rc.3)

klihub:
Replaced by #55.

PR #47, opened 2023-08-08
.github: add test build to CI workflow.

klihub:
Add a job to do a simple test build as part of the CI workflow. This should make sure that PRs don't break building any of the sample plugins.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,6 +33,30 @@ jobs:
         with:
           working-directory: src/github.com/containerd/nri
 
+  test-build:
+    name: Check buildability
+    runs-on: ubuntu-22.04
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -33,6 +33,30 @@ jobs:
...
 
+  test-build:
+    name: Check buildability
+    runs-on: ubuntu-22.04
+    timeout-minutes: 5
+
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          path: src/github.com/containerd/nri
```
fuweid, line 44: I think we can delete gopath setting in the follow-up pull requests because we already use go.mod.

fuweid (approved):
LGTM

PR #48, opened 2023-08-24
Add support for rlimits

samuelkarp:
This PR adds support for adjusting rlimits and provides a sample plugin.  I was able to test this using a patched version of containerd.


klihub (commented):
Thank you ! Looks pretty good. I'll read through it again in the morning just in case, but my impression is that it is very much ready to go in.

klihub (approved):

kad (approved):
/lgtm

Nice to see additional use cases and example plugins.

plugins/ulimit-adjuster/README.md
```diff
@@ -0,0 +1,36 @@
+## ulimit Adjuster Plugin
+
+This sample plugin can adjust ulimits for containers using pod annotations.
+
+### Annotations
+
+ulimits are annotated using the key
+`ulimits.nri.containerd.io/container.$CONTAINER_NAME`, which adjusts ulimits
+for `$CONTAINER_NAME`.  The ulimit names are the valid names of Linux resource
```
SergeyKanzhelev, line 9: are there scenarios where customer may want to set a single per-pod override that will apply to all containers?

samuelkarp: Possibly, but I don't have a specific use-case for that right now.  I was looking to just build the simplest thing for now; we can always add functionality later if it's desired.

plugins/ulimit-adjuster/README.md
```diff
@@ -0,0 +1,36 @@
...
+- type: RLIMIT_NOFILE
+  soft: 1024
+  hard: 4096
+- path: RLIMIT_MEMLOCK
+  soft: 1073741824
+  hard: 1073741824
+  ...
+```
+
+All fields are mandatory (`soft` and `hard` will be interpreted as 0 if
```
SergeyKanzhelev, line 25: is there relationship between those? Should hard default to soft if not defined?

plugins/ulimit-adjuster/adjuster.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+	ulimits := make([]ulimit, 0)
+	if err := yaml.Unmarshal([]byte(val), &ulimits); err != nil {
+		return nil, err
+	}
+	for i := range ulimits {
+		u := ulimits[i]
+		typ := strings.TrimPrefix(strings.ToUpper(u.Type), rlimitPrefix)
+		if _, ok := valid[typ]; !ok {
+			log.G(ctx).WithField("raw", u.Type).WithField("trimmed", typ).Debug("failed to parse type")
```
SergeyKanzhelev, line 155: maybe output `typ` to debug output

samuelkarp: `WithField("raw", u.Type).WithField("trimmed", typ)` that outputs `typ`.

plugins/ulimit-adjuster/README.md
samuelkarp: Hard should be >= soft.  I can add a validation for that.

samuelkarp: Validation added in 5ecea04.

plugins/ulimit-adjuster/README.md
SergeyKanzhelev: thank you! If there will be a scenario with some global setting for a Pod that will require all containers to adjust limits, it will make sense to have a single override. I am also not aware of scenarios that would need this.

SergeyKanzhelev:
lgtm

bobbypage:
lgtm

+1 to add support for rlimits in NRI, it will be helpful for folks who need to configure rlimits in k8s environments where rlimits are not natively supported in pod specs.

dmcgowan (approved):

PR #53, opened 2023-09-13
gha: remove GOPATH and workingdir, update actions/setup-go@v4, actions/checkout@v4

thaJeztah:


.github/workflows/ci.yml
```diff
@@ -14,48 +14,39 @@ jobs:
...
 
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
         with:
-          path: src/github.com/containerd/nri
           fetch-depth: 25
 
       - uses: containerd/project-checks@v1
-        with:
-          working-directory: src/github.com/containerd/nri
```
klihub, line 34: Isn't this the culprit for the DCO check failure ?

thaJeztah:
Looks like the containerd "project" action depends on GOPATH in some way; need to check why (and if we can fix that)

klihub:
> Looks like the containerd "project" action depends on GOPATH in some way; need to check why (and if we can fix that)

Actually it is not the DCO check that failed but the file header check. And looking at the [action](https://github.com/containerd/project-checks/blob/eae3c6e2e734bc70dc79b3d4c3bec73f0f99eeaa/action.yml#L42) it looks like it requires the `working-directory` input to be set. Now I wonder how come it did not fail with an error message complaining about that. Looks like it went berserk instead and scanned all files it could then complained about the lack of the containerd licence header.

thaJeztah:
Yes, I spotted that working-dir "required" property in the project-check repository; so I think that could be related indeed.

Guess I need to look more into that; in the meantime, I reverted the `GOPATH`, `path`, and `working-dir` changes for the project-checks (but kept it for all the other changes). I think that should probably unblock this PR

klihub (approved):
LGTM.

dmcgowan (approved):

PR #54, opened 2023-09-13
make plugins/ulimit-adjuster a separate module

thaJeztah:
Also switch it to use the new log module, instead of depending on containerd


klihub (approved):
LGTM

thaJeztah:
@samuelkarp @estesp ptal ☺️

samuelkarp (approved):
Thanks @thaJeztah!

thaJeztah:
Thanks! I'll rebase the other one, and then we can look if that's a good way forward to cut the circular dependency (at least for this module that is)

PR #57, opened 2023-09-22
NRI network

pfl:
Implementation of NRI network messages. The messages added are:
- NetworkConfigurationChanged when the CNI config has been read from disk
- PreSetupNetwork before CNI is set up
- PostSetupNetwork after successful CNI configuration
- Pre- and PostNetworkDeleted once the pod network is removed


pfl:
Removed 'WIP' labels as this PR has progressed into something decently useful.

README.md
```diff
@@ -204,6 +204,84 @@ The following pieces of container metadata are available to plugins in NRI:
...
+## Pod Network Lifecycle Events
+
+NRI plugins can subscribe to the following pod network events:
+
+  - Pod network configuration change
+  - Pod network pre-creation before CNI plugins are called
+  - Pod network post-creation after CNI plugins have been called
+  - Pod network removal
+
+### Pod Network Configuration Change
```
aojea, line 216: As a person that worked developing and maintaining different CNI plugins, the CNI word evolved from referring to the CNI spec to refer to component that implements the spec (Calico, Cilium, ovn-kubernetes, kindnet, ...) 

The signatue of the existing CNI calls in containerd are

> err = netPlugin.Status()
> err = netPlugin.Remove(ctx, id, path, opts...)
> result, err := netPlugin.Setup(ctx, id, path, opts...)

if those 3 methods are added as 3 hooks of the NRI plugin (so it can be actually replaced by the NRI plugin), an existing CNI implementation can move to a more rich and flexible model provided by NRI and move on from the legacy interface that has a lot of constrains (strong typing, shell, json, ...)  https://github.com/containernetworking/cni/issues?q=is%3Aissue+is%3Aopen+grpc

README.md
```diff
@@ -204,6 +204,84 @@ The following pieces of container metadata are available to plugins in NRI:
...
+</details>
+
+### PreSetupNetwork
+
+The PreSetupNetwork event is sent immediately before CNI is invoked to set
+up the pod network in its namespace. In addition to the PodSandbox object,
+arguments to PreSetupNetwork includes an array of strings containing all JSON
+formatted CNI configuration files associated with the pod.
+
+The reply data consists of CNI capabilities and 'args' arguments. Any returned
```
aojea, line 253: "CNI capabilities" is not a grea API, mainly because is a convention https://github.com/containernetworking/cni/blob/main/CONVENTIONS.md#current-conventions

README.md
aojea: These days the "CNI plugins" need to run connected to the apiserver to perform API operations, they can not longer run as standalone binaries, Cilum https://github.com/cilium/cilium/blob/d1128cf948979de7433385b2101a98404d99c64f/pkg/client/client.go#L32-L41
and OVN-kubernets https://github.com/ovn-org/ovn-kubernetes/blob/2916102d231ce9596e23c9880103225db2caa65e/go-controller/pkg/cni/cnishim.go#L37-L49 are alread using client/server model, with a binary that is implementing the CNI interface that uses a socket to connect to the "CNI plugin" , the NRI plugin can make this a native integration

maiqueb: +1 on the native integration. 

And I also agree that a "modern" plugin requires knowledge of API server objects (to various extents).

Generalizing this integration would definitely be beneficial.

pfl: > The signatue of the existing CNI calls in containerd are
> 
> > err = netPlugin.Status()
> > err = netPlugin.Remove(ctx, id, path, opts...)
> > result, err := netPlugin.Setup(ctx, id, path, opts...)
> 
> if those 3 methods are added as 3 hooks of the NRI plugin (so it can be actually replaced by the NRI plugin), an existing CNI implementation can move to a more rich and flexible model provided by NRI and move on from the legacy interface

If these three network methods were to be implemented and only these three new methods were used to set up a network, does that imply the contents of /etc/cni/net.d/ is no longer neither needed nor available, as e.g. a netPlugin.GetConfig() equivalent would not exist? Similarly, cniNetConfSyncer also disappears, as the network config is now in the hands of an NRI plugin? Or does the runtime still keep track of files in /etc/cni/net.d/, and keep sending the opts arguments to Setup() and Remove(), as is the case now?

pfl: > result, err := netPlugin.Setup(ctx, id, path, opts...)

Does the 'opts' argument here turn into json marshalled protobuf map<string, bytes> and unmarshalled in the other end as string, interface{} mappings in Go, or did you have some other format in mind?

README.md
pfl: Looking at the above comments, we need plain 'opts' transported back and forth here?

README.md
aojea: > as e.g. a netPlugin.GetConfig() equivalent would not exist? Similarly, cniNetConfSyncer also disappears, as the network config is now in the hands of an NRI plugin? Or does the runtime still keep track of files in /etc/cni/net.d/, and keep sending the opts arguments to Setup() and Remove(), as is the case now?

heh, that is the 1M dollar question

I think CNI solves a problem and works well solving that problem, but is hard to extend. If we think about this as CNI 2.0 , that is very similar per the issues linked https://github.com/containernetworking/cni/issues?q=is%3Aissue+is%3Aopen+grpc , it will logically to think about it as exclusive, so people should choose between v1 or v2, but never both at same time, I do not know if that is even possible though

pfl:
Let's revisit this PR once KNI has taken shape. At that point we will learn whether any CNI specific plugins will be needed or if plugins interested in network activity listen to KNI instead.

MikeZappa87:
@pfl want to reopen this?

pfl:
Once KNI messaging gets specified, NRI plugins may perhaps hook into KNI directly. If needed, a non-KNI NRI based setup should be as close to a KNI one as possible in order to avoid excessive coding and added architecture. Right now I'm unfortunately (de)allocated to some DRA work until mid Q3. Thus the pause in this PR.

Issue #59, opened 2023-10-26
NRI 0.1.0 can not get the annotations of sandbox


sctb512:
After I read the code of [clearcfs example](https://github.com/containerd/nri/blob/main/examples/clearcfs/main.go#L46C2-L46C2), I found the `r.Spec.Annotations` is only works for container but not pod sandbox. I need to pass some parameters through the pod's annotations.

Although the NRI 0.5.0, which is integrated for containerd 1.7.0, is released, but NRI 0.1.0 (containerd 1.6.x) is also used as the mainly version.

Actual:
```
Oct 25 19:26:43 10.37.7.188 containerd[4178092]: time="2023-10-25T19:26:43+08:00" level=info msg="
id: e5239948de1b030f3fc2051f5a802c35e7d880516e8873f4600bd65c58a6b251, 
sandbox id: 7ced537e0d9605fb8b01cec5acb6fec6331a63974d8d2a1f79caba6755178b84. 
Namespaces: map[
	ipc:/proc/4178623/ns/ipc 
	mount: 
	network:/proc/4178623/ns/net 
	pid:/proc/4178623/ns/pid 
	uts:/proc/4178623/ns/uts
], 
cgroups path: /k8s.io/e5239948de1b030f3fc2051f5a802c35e7d880516e8873f4600bd65c58a6b251, 
annonation: map[string]string{
	"io.kubernetes.cri.container-name":"nginx", 
	"io.kubernetes.cri.container-type":"container", 
	"io.kubernetes.cri.image-name":"nginx:latest", 
        "io.kubernetes.cri.sandbox-id":"7ced537e0d9605fb8b01cec5acb6fec6331a63974d8d2a1f79caba6755178b84", 
        "io.kubernetes.cri.sandbox-name":"nginx-sandbox", 
	"io.kubernetes.cri.sandbox-namespace":"default", 
	"io.kubernetes.cri.sandbox-uid":"hdishd83djaidwnduwk28bcsb"
},
"
```

Expected:
```
We can get the annonations of pod sandbox.
```
The definitation of `r.Spec` is here:

```go
// Spec for the container being processed
type Spec struct {
	// Resources struct from the OCI specification
	//
	// Can be WindowsResources or LinuxResources
	Resources json.RawMessage `json:"resources"`
	// Namespaces for the container
	Namespaces map[string]string `json:"namespaces,omitempty"`
	// CgroupsPath for the container
	CgroupsPath string `json:"cgroupsPath,omitempty"`
[...]

PR #60, opened 2023-10-26
add sandbox Annotations to NRI Request

sctb512:
Currently, containerd only passes `pod.Labels` to the NRI (v0.1.0) plugin, but not `pod.Annotations`. In some situations we prefer to set `Pod.Spec.Annotations` instead of `Pod.Spec.Labels`.

This PR provides support for passing sandbox annotations to the NRI plugin.

Related to #59


samuelkarp:
I'm not sure if we want to add new functionality to the 0.1.x series here.  The release/0.1 branch was really created just for dependency cleanup.

sctb512:
> I'm not sure if we want to add new functionality to the 0.1.x series here. The release/0.1 branch was really created just for dependency cleanup.

Should we create a new branch to add new features to `v0.1.0`? The `v0.1.0` is now the main product version, which was integrated with containerd 1.6.x.

samuelkarp:
> Should we create a new branch to add new features to v0.1.0?

I don't know if we want to add new features to v0.1.x.

> The v0.1.0 is now the main product version, which was integrated with containerd 1.6.x.

It's not clear whether we'd want to incorporate this into containerd 1.6.  containerd 1.6 is an [LTS](https://containerd.io/releases/#support-horizon) with some limited exceptions for adding new features, but:

> LTS releases can also accept feature backports to support new Kubernetes releases. The default action has to be reject it though, for long-term stability. This is still negotiable when the feature is a hard dependency for a new release of Kubernetes.

This change does not appear to be a hard dependency for a new release of Kubernetes.

sctb512:
> This change does not appear to be a hard dependency for a new release of Kubernetes.

Yes, it's not a hard dependency for a new release of Kubernetes. 

However, the NRI for containerd 1.7.x has been redesigned. So, it seems hard to add new features to NRI v0.1.x. Perhaps updating containerd to 1.7.x is the only way to use the new features of NRI?

samuelkarp:
> Perhaps updating containerd to 1.7.x is the only way to use the new features of NRI?

Yes

sctb512:
> > Perhaps updating containerd to 1.7.x is the only way to use the new features of NRI?
> 
> Yes

Got it. Thanks.

PR #66, opened 2024-01-29
socketpair_unix: avoid double close(), set FD_CLOEXEC

champtar:
We were calling the close() syscall multiple time
with the same fd number, leading to random issues like
closing containerd stream port.

In newLaunchedPlugin() we have:
```
sockets, _ := net.NewSocketPair()
defer sockets.Close()
conn, _ := sockets.LocalConn()
peerFile := sockets.PeerFile()
defer func() {
  peerFile.Close()
  if retErr != nil {
    conn.Close()
  }
}()
cmd.Start()
```

so we were doing:
```
close(local) (in LocalConn())
cmd.Start()
close(peer) (peerFile.Close())
close(local) (sockets.Close())
close(peer) (sockets.Close())
```

If the NRI plugin that we launch with cmd.Start() is not cached or
the system is a bit busy, cmd.Start() gives a large enough window
for another goroutine to open a file that will get the same fd number
as local was, thus being closed by accident.

Fix the situation by storing os.Files instead of ints, so that closing
multiple times just returns an error.

Also set FD_CLOEXEC on the sockets so we don't leak them.

Fixes 1da2cdfebfcecfb6cb0c93fc20fa3822d918f00b

### DIFF
--- a/pkg/net/socketpair_unix.go
+++ b/pkg/net/socketpair_unix.go
@@ -22,76 +22,74 @@ import (
 	"fmt"
 	"net"
 	"os"
-
-	syscall "golang.org/x/sys/unix"
 )
 
-const (
-	local = 0
-	peer  = 1
-)
-
-// SocketPair contains the file descriptors of a connected pair of sockets.
-type SocketPair [2]int
+// SocketPair contains the os.File of a connected pair of sockets.
+type SocketPair struct {
+	local, peer *os.File
+}
 
 // NewSocketPair returns a connected pair of sockets.
 func NewSocketPair() (SocketPair, error) {
-	fds, err := syscall.Socketpair(syscall.AF_UNIX, syscall.SOCK_STREAM, 0)
+	fds, err := newSocketPairCLOEXEC()
 	if err != nil {
-		return [2]int{-1, -1}, fmt.Errorf("failed to create socketpair: %w", err)
+		return SocketPair{nil, nil}, fmt.Errorf("failed to create socketpair: %w", err)
 	}
 
-	return fds, nil
+	filename := fmt.Sprintf("socketpair-#%d:%d", fds[0], fds[1])
+
+	return SocketPair{
+		os.NewFile(uintptr(fds[0]), filename+"[0]"),
+		os.NewFile(uintptr(fds[1]), filename+"[1]"),
+	}, nil
 }
 
-// LocalFile returns the socketpair fd for local usage as an *os.File.
-func (fds SocketPair) LocalFile() *os.File {
-	return os.NewFile(uintptr(fds[local]), fds.fileName()+"[0]")
+// LocalFile returns the local end of the socketpair as an *os.File.
+func (sp SocketPair) LocalFile() *os.File {
+	return sp.local
 }
 
-// PeerFile returns the socketpair fd for peer usage as an *os.File.
-func (fds SocketPair) PeerFile() *os.File {
-	return os.NewFile(uintptr(fds[peer]), fds.fileName()+"[1]")
+// PeerFile returns the peer end of the socketpair as an *os.File.
+func (sp SocketPair) PeerFile() *os.File {
+	return sp.peer
 }
 
 // LocalConn returns a net.Conn for the local end of the socketpair.
-func (fds SocketPair) LocalConn() (net.Conn, error) {
-	file := fds.LocalFile()
+// This closes LocalFile().
+func (sp SocketPair) LocalConn() (net.Conn, error) {
+	file := sp.LocalFile()
 	defer file.Close()
 	conn, err := net.FileConn(file)
 	if err != nil {
-		return nil, fmt.Errorf("failed to create net.Conn for %s[0]: %w", fds.fileName(), err)
+		return nil, fmt.Errorf("failed to create net.Conn for %s: %w", file.Name(), err)
 	}
 	return conn, nil
 }
 
 // PeerConn returns a net.Conn for the peer end of the socketpair.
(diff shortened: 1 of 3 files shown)


klihub:
> not tested with containerd / windows version also needs fixing

Yeah, we haven't really had the time to impement and test NRI for Windows, yet. That said, on Windows, SocketPair emulates a socketpair using two ends of a pre-connected TCPv4 socket, which are then stored as sys.Handle's. Shouldn't that protect us from a similar double-close error ?

klihub (approved):
This looks fine to me.

champtar:
@klihub just pushed a version fixing windows and moving all the common functions in socketpair.go

pkg/net/socketpair_unix.go
```diff
@@ -20,78 +20,27 @@ package net
...
-}
-
-// PeerConn returns a net.Conn for the peer end of the socketpair.
-func (fds SocketPair) PeerConn() (net.Conn, error) {
-	file := fds.PeerFile()
-	defer file.Close()
-	conn, err := net.FileConn(file)
+	syscall.ForkLock.RLock()
+	defer syscall.ForkLock.RUnlock()
+	fds, err := unix.Socketpair(unix.AF_UNIX, unix.SOCK_STREAM, 0)
```
fuweid, line 33: would you mind to use `unix.SOCK_CLOEXEC` here? so that we don't have to lock.

pkg/net/socketpair_windows.go
```diff
@@ -55,15 +47,14 @@ func emulateWithPreConnect() (SocketPair, error) {
 		err error
 	)
 
+	syscall.ForkLock.RLock()
+	defer syscall.ForkLock.RUnlock()
```
fuweid, line 51: I think the lock is too big here. would you please use a wrap to handle each syscall?

REF: https://github.com/golang/go/blob/d7df7f4fa01f1b445d835fc908c54448a63c68fb/src/net/sock_windows.go#L31

fuweid:
Hi @champtar would you please file a ticket to track window part and focus on linux part in this pull request? I don't have too much knowledge on windows platform actually. we can ask other maintainers to take over this. Thanks

klihub:
> Hi @champtar would you please file a ticket to track window part and focus on linux part in this pull request? I don't have too much knowledge on windows platform actually. we can ask other maintainers to take over this. Thanks

+1 I would also leave Windows out of this now and focus on fixing the bug on un*x, since we anyway don't have all the necessary functionality in place for Windows. Sorry if my earlier question/comment was misleading and could be interpreted otherwise.

pkg/net/socketpair_unix.go
champtar: unix.SOCK_CLOEXEC doesn't exists for darwin for example, this is the generic way

champtar:
Pushed back the unix only changes, I'll open a second PR once this one is merged

pkg/net/socketpair_unix.go
champtar: I'm fine doing a linux only version, just tell me

fuweid: Let's focus on linux right now and add issue to track it for darwin. Thanks!

pkg/net/socketpair_unix.go
```diff
@@ -22,36 +22,43 @@ import (
...
 // LocalFile returns the socketpair fd for local usage as an *os.File.
 func (fds SocketPair) LocalFile() *os.File {
-	return os.NewFile(uintptr(fds[local]), fds.fileName()+"[0]")
+	return fds.local
 }
 
 // PeerFile returns the socketpair fd for peer usage as an *os.File.
 func (fds SocketPair) PeerFile() *os.File {
-	return os.NewFile(uintptr(fds[peer]), fds.fileName()+"[1]")
+	return fds.peer
```
klihub, line 61: Ditto here. Isn't this also changing the semantics of PeerConn() so that it can only be called once with subsequent calls failing ? Should at least be documented.

pkg/net/socketpair_windows.go
champtar: Will be done in windows PR

pkg/net/socketpair_unix.go
```diff
@@ -22,36 +22,43 @@ import (
...
+	return SocketPair{
+		os.NewFile(uintptr(fds[0]), filename+"[0]"),
+		os.NewFile(uintptr(fds[1]), filename+"[1]"),
+	}, nil
 }
 
 // LocalFile returns the socketpair fd for local usage as an *os.File.
 func (fds SocketPair) LocalFile() *os.File {
-	return os.NewFile(uintptr(fds[local]), fds.fileName()+"[0]")
+	return fds.local
```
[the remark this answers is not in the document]
champtar: Before this PR, a second call to LocalConn would potentially try to use a completely unrelated file. If I don't close LocalFile on of the test timeout. I'll improve documentation. I can also return nil after first *File call if preferred.

pkg/net/socketpair_unix.go
champtar: I've added a linux and a unix version

pkg/net/socketpair_unix.go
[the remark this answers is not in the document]
champtar: Added a small comment in the doc `// This closes LocalFile().`

champtar:
@klihub hopefully final version

klihub (approved):

pkg/net/socketpair_windows.go
champtar: PR for windows: https://github.com/containerd/nri/pull/69

PR #67, opened 2024-01-30
adaptation, stub: allow extra ttrpc client and server options.

klihub:
This patch set adds support for passing extra ttrpc client and server options for both the runtime adaptation and the plugin stub. The primary purpose of this PR is to implement minimal plumbing which will allow the propagation of OpenTelemetry span information from runtimes to plugins, over NRI ttrpc connections,using the instrumentation interceptors implemented by https://github.com/containerd/otelttrpc.

This PR is an alternative to #68. This PR simply opens up the ttrpc client and server used by NRI for extra ttrpc options. This allows one to pass in the necessary options to enable otel trace span propagation (essentially using [github.com/containerd/]otelttrpc).

#68 on the other hand introduces two boolean options which tell NRI whether it internally should set up the necessary ttrpc options to enable otel trace span propagation over ttrpc. That brings in an indirect dependency on [github.com/containerd/]otelttrpc and OpenTelemetry itself.

I wanted to ask feedback about which of these alternative approaches others find more preferable, hence I filed both as draft PRs.

I think the biggest difference between these two approaches are in the blast radius of the resulting dependency changes, since #68 pulls in OpenTelemetry as an indirect dependency via otelttrpc.


cpuguy83:
I think I like this one better?

fuweid:
+1 for this one

klihub:
@cpuguy83 @fuweid ACK, closed #68 and removed the draft status from this one.

fuweid (approved):
LGTM

cpuguy83 (approved):

PR #68, opened 2024-01-30
Add support for propagating OpenTelemetry trace span information over NRI ttrpc connection.

klihub:
This patch set implements extra runtime/adaptation and plugin/stub NRI options which enable the propagation of OpenTelemetry trace span information over ttrpc connections.

This PR is an alternative to #67. This PR introduces two boolean options which tell NRI whether it internally should set up the necessary ttrpc options to enable otel trace span propagation over ttrpc. This brings in an indirect dependency on [github.com/containerd/]otelttrpc and OpenTelemetry itself.

 #67 on the other hand simply opens up the ttrpc client and server used by NRI for extra ttrpc options. This allows one to pass in the necessary options to enable otel trace span propagation (essentially using the same otelttrpc instrumenting ttrpc interceptors as used here), both in NRI enabled runtimes and NRI plugins.

I wanted to ask feedback about which of these alternative approaches others find more preferable, hence I filed both as draft PRs.


klihub:
Closing in favor of #67.

PR #73, opened 2024-03-08
set stop status when exiting.

yylt:
update `stub.started` when stub stop 

Signed-off-by: yylt <yang8518296@163.com>

### DIFF
--- a/pkg/stub/stub.go
+++ b/pkg/stub/stub.go
@@ -430,6 +430,7 @@ func (stub *stub) close() {
 		if stub.srvErrC != nil {
 			<-stub.doneC
 		}
+		stub.started = false
 	})
 }
 


klihub:
Hmm... I suspect you try to make `stub.Stub` re-`Start()`able, right ?

I am not sure if that is a good idea, mostly because I'm not sure at the moment if this is enough to safely be able to do that. If you need to restart the your plugin after a disconnect from the runtime, I think it would be a safer alternative to create a new stub for the plugin and start that one instead.

yylt:
> @yylt I suspect you try to make `stub.Stub` re-`Start()`able, right ?
> 
> I am not sure if that is a good idea, mostly because I'm not sure at the moment if this is enough to safely be able to do that. If you need to restart the your plugin after a disconnect from the runtime, I think it would be a safer alternative to create a new stub for the plugin and start that one instead.

haha... indeed, the intention was to restart, but later it was found that there were still other issues about `channel` with the `start` , so it was changed to using the method of creating a new struct.

If it is still not possible to start after stopping, there is really no need to set it up. 

I'm not sure how things will evolve afterwards, but perhaps this change can be added.

klihub:
> > @yylt I suspect you try to make `stub.Stub` re-`Start()`able, right ?
> > I am not sure if that is a good idea, mostly because I'm not sure at the moment if this is enough to safely be able to do that. If you need to restart the your plugin after a disconnect from the runtime, I think it would be a safer alternative to create a new stub for the plugin and start that one instead.
> 
> haha... indeed, the intention was to restart, but later it was found that there were still other issues about `channel` with the `start` , so it was changed to using the method of creating a new struct.
> 
> If it is still not possible to start after stopping, there is really no need to set it up.
> 
> I'm not sure how things will evolve afterwards, but perhaps this change can be added.

@yylt Is it then okay for you if I close this with a comment summarizing the above conclusion ? If we later decide that we want to make the stub re`Start()`able, we can reopen this and stack any other necessary commits on top to achieve that goal.

PR #84, opened 2024-05-21
Modify hook-injector plugin to monitor directories to match cri-o

AbdelrahmanElawady:
As mentioned in hook-injector plugin README, this plugin achieves CRI-O compatible hooks injection. However, there is a difference between this implementation and CRI-O's implementation as CRI-O watches hook directories for any changes which allows for new hooks to be added to new containers. This is not the case here where this plugin only uses hooks created while the plugin is first running. So, they are not really compatible. [Reference CRI-O code](https://github.com/cri-o/cri-o/blob/5ab50a9b6c6152d8eae4850b2668ff82c0b2b67a/cmd/crio/main.go#L385-L396).

Also, there are some use-cases where services set up their hooks after the cluster and the plugin are up. CRI-O would register these new hooks but again the plugin currently won't. This PR aims to fix that.

We can also make this optional if you don't want to change the current behavior with a flag like `--watch` or something like that.

### DIFF
--- a/plugins/hook-injector/hook-injector.go
+++ b/plugins/hook-injector/hook-injector.go
@@ -135,11 +135,12 @@ func dump(args ...interface{}) {
 
 func main() {
 	var (
-		pluginName string
-		pluginIdx  string
-		opts       []stub.Option
-		mgr        *hooks.Manager
-		err        error
+		pluginName   string
+		pluginIdx    string
+		disableWatch bool
+		opts         []stub.Option
+		mgr          *hooks.Manager
+		err          error
 	)
 
 	log = logrus.StandardLogger()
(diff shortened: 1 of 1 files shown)


mikebrow (commented):
nice.. 

Suggest adding --disableWatch as an option ... might want to consider outputing the paths being monitored in help for the client or as debug out (unless that is already being done in the hooks package)  .. 

Could you dump your console output on the PR?  Would be nice if we had hook tests running in critest or some other integration testing.

Maybe open an issue for a test bucket?

AbdelrahmanElawady:
new flag added and the output looks like this:
```console
$ sudo ./hook-injector -idx 10
INFO   [0000] Created plugin 10-hook-injector (hook-injector, handles CreateContainer) 
INFO   [0000] watching directories "/usr/share/containers/oci/hooks.d /etc/containers/oci/hooks.d" for new changes 
INFO   [0000] Registering plugin 10-hook-injector...       
INFO   [0000] Configuring plugin 10-hook-injector for runtime containerd/1.7.2... 
INFO   [0000] Started plugin 10-hook-injector...
``

AbdelrahmanElawady:
> Maybe open an issue for a test bucket?

Also, created an issue to track hook tests: #86

mikebrow (approved):
LGTM

mikebrow:
@haircommander FYI in case you want to review this one or tag someone

klihub (approved):

haircommander:
LGTM

Issue #80, opened 2024-05-07
RemoveMount does not work


liangjingtao11:
containerd version： v1.7.14
nri version: 0.6.1

i want to remove Specific mount using this code in nri
`for _, m := range container.Mounts {
	if strings.HasSuffix(m.Source, mountSuffix) {
		adjust.RemoveMount(m.Destination)
}`

but after pod created，Nothing has changed. May I ask if I used the wrong way?

liangjingtao11:
nri code:

pkg/adaption/result.go

```
unc (r *result) adjustMounts(mounts []*Mount, plugin string) error {
	if len(mounts) == 0 {
		return nil
	}

	create, id := r.request.create, r.request.create.Container.Id

	// first split removals from the rest of adjustments
	add := []*Mount{}
	del := map[string]*Mount{}
	mod := map[string]*Mount{}
	for _, m := range mounts {
		if key, marked := m.IsMarkedForRemoval(); marked {
			del[key] = m
		} else {
			add = append(add, m)
			mod[key] = m
		}
	}

	// next remove marked mounts from collected adjustments
	cleared := []*Mount{}
	for _, m := range r.reply.adjust.Mounts {
		if _, removed := del[m.Destination]; removed {
			r.owners.clearMount(id, m.Destination)
			continue
		}
		cleared = append(cleared, m)
	}
	r.reply.adjust.Mounts = cleared

	// next remove marked and modified mounts from container creation request
	cleared = []*Mount{}
	for _, m := range create.Container.Mounts {
		if _, removed := del[m.Destination]; removed {
			continue
		}
		if _, modified := mod[m.Destination]; modified {
			continue
		}
		cleared = append(cleared, m)
	}
	create.Container.Mounts = cleared

	// next, apply additions/modifications to collected adjustments
	for _, m := range add {
		if err := r.owners.claimMount(id, m.Destination, plugin); err != nil {
			return err
		}
		r.reply.adjust.Mounts = append(r.reply.adjust.Mounts, m)
	}

	// finally, apply additions/modifications to plugin container creation request
	create.Container.Mounts = append(create.Container.Mounts, add...)

	return nil
}
```


it only append mount from add:  r.reply.adjust.Mounts = append(r.reply.adjust.Mounts, m).  That's why del mount was not passed to containerd

klihub:
Thanks, both for the report and the analysis !

PR #87, opened 2024-05-29
fix nri removeMount does not work bug

liangjingtao11:
nri needs to pass the del mount info to containerd, so that containerd can handle the mount records that need to be deleted correctly

fix: https://github.com/containerd/nri/issues/80

### DIFF
--- a/pkg/adaptation/result.go
+++ b/pkg/adaptation/result.go
@@ -322,6 +322,11 @@ func (r *result) adjustMounts(mounts []*Mount, plugin string) error {
 		r.reply.adjust.Mounts = append(r.reply.adjust.Mounts, m)
 	}
 
+	// need add del mount, containerd needs to process this del mounts
+	for _, m := range del {
+		r.reply.adjust.Mounts = append(r.reply.adjust.Mounts, m)
+	}
+
 	// finally, apply additions/modifications to plugin container creation request
 	create.Container.Mounts = append(create.Container.Mounts, add...)
 


mikebrow:
https://github.com/containerd/nri/blob/main/pkg/api/adjustment.go#L20-L32 for consideration...

pkg/adaptation/result.go
```diff
@@ -322,6 +322,11 @@ func (r *result) adjustMounts(mounts []*Mount, plugin string) error {
 		r.reply.adjust.Mounts = append(r.reply.adjust.Mounts, m)
 	}
 
+	// need add del mount, containerd needs to process this del mounts
```
klihub, line 325: nitpick: I would change 'containerd' in the above comment to 'runtime'. Maybe rephrase to something like ' add deleted mounts, the runtime needs them to properly remove deleted mounts'.

klihub:
@liangjingtao11 It would be nice to have a test case for this.

mikebrow:
@liangjingtao11 do you have time to finish this PR?

pkg/adaptation/result.go
```diff
@@ -322,6 +322,11 @@ func (r *result) adjustMounts(mounts []*Mount, plugin string) error {
...
+	for _, m := range del {
+		r.reply.adjust.Mounts = append(r.reply.adjust.Mounts, m)
+	}
```
klihub, line 328: @liangjingtao11 I think doing it this way breaks cases where a single plugin does a RemoveMount() followed by an AddMount(), both using the same destination. This is the idiomatic way for a plugin to change a mount in the case where that mount might have been injected (or updated) by another plugin earlier in the chain. With your changes, that will end up with a final adjustment which will cause the runtime to remove such a mount from the container.

klihub:
@liangjingtao11 This fix is not sufficient in it's current form as it breaks replacing a mount with another one to the same destination. I think we'd need a bit of additional code to prevent that from happening. [Maybe something like this](https://github.com/klihub/nri/tree/fixes/mount-removal-adjustment), where 8f509e7d0f2e4e9af7f7acb77e9c18f2e882efaf is an updated version of your commit, and 7ea0a47f00f853b4e970b980b1a24b082c0930f0 adds a new test case for mount removal.

klihub:
@liangjingtao11 Ping. Do you have time to address the remaining issues ?

klihub:
Closing in favor of updated PR #107.

PR #91, opened 2024-07-04
stub: support restart after stub stopped

zhaodiaoer:
In some scenarios such as runtime restart or the occurrence of nri request timeout, the ttrpc connections between the plugin and the runtime will be actively closed by the runtime, even the underlying network connection will be closed together. After this, the plugin must need to re-register to the adaptation side, but now the stub object cannot be reused for this; if the running plugin wants to reconnect to the runtime, the only way is to create a new stub for the plugin.

This commit introduce a new stub option `WithAutoRestartDelay`. The plugin developer can use this to build a reuseable stub, which will wait for a confiurable time duration and restart automatically after the ttrpc connection is lost, then plugin will be auto-reregistered to the adaptation.

### DIFF
--- a/pkg/stub/stub.go
+++ b/pkg/stub/stub.go
@@ -137,7 +137,9 @@ type PostUpdateContainerInterface interface {
 
 // Stub is the interface the stub provides for the plugin implementation.
 type Stub interface {
-	// Run the plugin. Starts the plugin then waits for an error or the plugin to stop
+	// Run starts the plugin then waits for the plugin service to exit, either due to a
+	// critical error or an explicit call to Stop(). Once Run() returns, the plugin can be
+	// restarted by calling Run() or Start() again.
 	Run(context.Context) error
 	// Start the plugin.
 	Start(context.Context) error
@@ -255,7 +257,6 @@ type stub struct {
 	rpcs       *ttrpc.Server
 	rpcc       *ttrpc.Client
 	runtime    api.RuntimeService
-	closeOnce  sync.Once
 	started    bool
 	doneC      chan struct{}
 	srvErrC    chan error
@@ -288,7 +289,6 @@ func New(p interface{}, opts ...Option) (Stub, error) {
 		idx:        os.Getenv(api.PluginIdxEnvVar),
 		socketPath: api.DefaultSocketPath,
 		dialer:     func(p string) (stdnet.Conn, error) { return stdnet.Dial("unix", p) },
-		doneC:      make(chan struct{}),
 	}
 
 	for _, o := range opts {
@@ -316,10 +316,10 @@ func (stub *stub) Start(ctx context.Context) (retErr error) {
 	stub.Lock()
 	defer stub.Unlock()
 
-	if stub.started {
+	if stub.isStarted() {
 		return fmt.Errorf("stub already started")
 	}
-	stub.started = true
+	stub.doneC = make(chan struct{})
 
 	err := stub.connect()
 	if err != nil {
@@ -378,10 +378,11 @@ func (stub *stub) Start(ctx context.Context) (retErr error) {
 
 	stub.srvErrC = make(chan error, 1)
 	stub.cfgErrC = make(chan error, 1)
-	go func() {
-		stub.srvErrC <- rpcs.Serve(ctx, rpcl)
-		close(stub.doneC)
-	}()
+
+	go func(l stdnet.Listener, doneC chan struct{}, srvErrC chan error) {
+		srvErrC <- rpcs.Serve(ctx, l)
+		close(doneC)
+	}(rpcl, stub.doneC, stub.srvErrC)
 
 	stub.rpcm = rpcm
 	stub.rpcl = rpcl
@@ -401,6 +402,7 @@ func (stub *stub) Start(ctx context.Context) (retErr error) {
 
 	log.Infof(ctx, "Started plugin %s...", stub.Name())
 
+	stub.started = true
 	return nil
 }
 
@@ -413,24 +415,42 @@ func (stub *stub) Stop() {
 	stub.close()
 }
 
+// IsStarted returns true if the plugin has been started either by Start() or by Run().
+func (stub *stub) IsStarted() bool {
+	stub.Lock()
+	defer stub.Unlock()
+	return stub.isStarted()
+}
+
+func (stub *stub) isStarted() bool {
+	return stub.started
+}
+
+// reset stub to the status that can initiate a new
+// NRI connection, the caller must hold lock.
 func (stub *stub) close() {
-	stub.closeOnce.Do(func() {
-		if stub.rpcl != nil {
-			stub.rpcl.Close()
-		}
-		if stub.rpcs != nil {
-			stub.rpcs.Close()
-		}
-		if stub.rpcc != nil {
-			stub.rpcc.Close()
-		}
-		if stub.rpcm != nil {
-			stub.rpcm.Close()
-		}
-		if stub.srvErrC != nil {
-			<-stub.doneC
-		}
-	})
+	if !stub.isStarted() {
+		return
+	}
+
+	if stub.rpcl != nil {
+		stub.rpcl.Close()
+	}
+	if stub.rpcs != nil {
+		stub.rpcs.Close()
+	}
+	if stub.rpcc != nil {
+		stub.rpcc.Close()
+	}
+	if stub.rpcm != nil {
+		stub.rpcm.Close()
+	}
+	if stub.srvErrC != nil {
+		<-stub.doneC
+	}
+
+	stub.started = false
+	stub.conn = nil
 }
 
 // Run the plugin. Start event processing then wait for an error or getting stopped.
@@ -449,14 +469,11 @@ func (stub *stub) Run(ctx context.Context) error {
 	return err
 }
 
-// Wait for the plugin to stop.
+// Wait for the plugin to stop, should be called after Start() or Run().
 func (stub *stub) Wait() {
-	stub.Lock()
-	if stub.srvErrC == nil {
-		return
+	if stub.IsStarted() {
+		<-stub.doneC
 	}
-	stub.Unlock()
-	<-stub.doneC
 }
 
 // Name returns the full indexed name of the plugin.
@@ -518,7 +535,9 @@ func (stub *stub) register(ctx context.Context) error {
 
 // Handle a lost connection.
 func (stub *stub) connClosed() {
+	stub.Lock()
 	stub.close()
+	stub.Unlock()
 	if stub.onClose != nil {
 		stub.onClose()
 		return


zhaodiaoer:
@klihub @fuweid @mikebrow can anyone help to review this ? thanks.

klihub:
> @klihub @fuweid @mikebrow can anyone help to review this ? thanks.

My first question is that do we really want to initiate a reconnection from the stub itself ? Instead of letting some higher-level application logic decide if, and when, attempting a reconnection is both desirable and safe ?

We have the `WithOnClose()` stub option which should allow applications to detect when a connection to the runtime is gone and then trigger at will a reconnection of their own accord.

zhaodiaoer:
> > @klihub @fuweid @mikebrow can anyone help to review this ? thanks.
> 
> My first question is that do we really want to initiate a reconnection from the stub itself ? Instead of letting some higher-level application logic decide if, and when, attempting a reconnection is both desirable and safe ?
> 
> We have the `WithOnClose()` stub option which should allow applications to detect when a connection to the runtime is gone and then trigger at will a reconnection on their own accord.

I quite approve of the idea of leaving the reconnection work to the higher-level application to handle. However, in the actual attempt, I discovered the following problems that are difficult to solve by the higher-level application logic. Most of them come from the current internal logic of the stub:

1. When the initial `stub.conn` is closed, it is not `nil` and no any logic will reset it to `nil`. Therefore, `stub.connect()` will do nothing. The `multiplex` required for reconnection will be constructed base the old `stub.conn` (the ttrpc connection stage comes after `multiplex` construct). Here, it is required that `stub.conn` is a `stdnet.Conn` implementation that is still usable after being closed. Maybe there are some cunning ways to achieve this, but it is inconsistent with the semantics defined by the `stdnet.Conn` interface.
2. The handler provided by `WithOnClose()` cannot take stub object, so it cannot operate `stub.conn` directly, only can send some notify to other functions to do reconnection work.
3. Last minor problem: just like the `stub.conn` variable mentioned in the 1st problem, once `stub.started` is set to true, there is no chance to reset it to false, new multiplex and new ttrpc connections will not be created.

To summarize the problems described above, most of the current connection handling logic is defined within the stub, and there are not enough exposed public methods to implement custom reconnection logic. If my understanding is incorrect, or if there are currently any proposals for optimizing the design of the plugin <-> runtime connection management, please let me know. I am willing to implement the code for these designs. But before that, if a convenient reconnection handling option can be provided to plugin developers for easy use, it might also be a good choice.

klihub:
> When the initial stub.conn is closed, it is not nil and no any logic will reset it to nil. Therefore, stub.connect() will do nothing

Sorry, I was not specific enough. I did not mean to try reusing/reconnecting the already disconnected stub. I meant to decide/check if and when re-establishing is safe, then create a new stub and connect that one...

klihub:
> > When the initial stub.conn is closed, it is not nil and no any logic will reset it to nil. Therefore, stub.connect() will do nothing
> 
> Sorry, I was not specific enough. I did not mean to try reusing/reconnecting the already disconnected stub. I meant to decide/check if and when re-establishing the connection is safe, then create a new stub and connect that one...

zhaodiaoer:
> > When the initial stub.conn is closed, it is not nil and no any logic will reset it to nil. Therefore, stub.connect() will do nothing
> 
> Sorry, I was not specific enough. I did not mean to try reusing/reconnecting the already disconnected stub. I meant to decide/check whether and when re-establishing the connection is safe, then create a new stub and connect that one...

oh i got you, In fact, currently we do achieve the purpose of reconnection and re-registration plugin by rebuilding the stub object.

klihub:
> > > When the initial stub.conn is closed, it is not nil and no any logic will reset it to nil. Therefore, stub.connect() will do nothing
> > 
> > 
> > Sorry, I was not specific enough. I did not mean to try reusing/reconnecting the already disconnected stub. I meant to decide/check whether and when re-establishing the connection is safe, then create a new stub and connect that one...
> 
> oh i got you, In fact, currently we do achieve the purpose of reconnection and re-registration plugin by rebuilding the stub object.
> 
> I think a rather crucial point here is what the long-term positioning of the `stub` is like. Does it only cover the lifecycle of a single connection or the lifecycle of the plugin?

Well, (be)for(e) answering that question there is a related question to consider: 'what is a plugin ?' or more precisely 'can the plugins lifecycle be different from the lifecycle of the connection of the plugin ?'.

These might sound very philosophical at first, outright ridiculous even, before you start to consider things from the runtime's perspective. On the client side, sure, there is a clear difference between the lifecycles of the plugin and its connection. On the runtime side however, there isn't a clear distinction. Once a connection is closed/lost, the runtime considers the plugin gone forever. If the plugin was an internal one, we would need to restart it to get it connected, up, and running again. That creates a new connection, a new process, and a new struct for the plugin. Clearly a new instance from all aspects. If it was an external plugin, even if it does not exit but reconnects from within the same process, a new connection and a new internal struct  is created in the runtime. Again, a new instance from the runtime's point of view. So from the runtime's perspective the lifecycles of the connection and the plugin are the same...

> If it is an external type of plugin, it is a relatively common requirement for its operation cycle to cover multiple connections. Rebuilding the entire stub object every time the connection is disconnected is a bit counterintuitive.

That is a fair and valid point. I'd still argue that we should make the stub restartable and reconnectable by the plugin, instead of trying to make the stub restart and reconnect itself.

Considering that it is completely unpredictable in what context a 'connection lost' event comes and in what state the plugin itself is (as opposed to the stub) at that moment, only the application can know when it is safe to reconnect and, for instance, what kind of locking or other synchronization need to be taken care of before this can be attempted.

So IMO, we should reset/clear enough of the internal state of the plugin that it should be possible to safely restart/reconnect it. And maybe this should always be done explicitly... perhaps only by an explicit call to `Stop()` ?

So then for a reconnectable plugin
- high level plugin logic would need to install an `OnClose()` handler to notice lost connections
- arrange it so that the stub is `Stop()`ped once connection is lost (not necessarily from the handler)
- `Stop()` would cause the stub to reset enough of its internal state
- eventually the plugin would call `Start()` on the stub again to reconnect, once this is safe to do

zhaodiaoer:
> So then for a reconnectable plugin
> 
> * high level plugin logic would need to install an `OnClose()` handler to notice lost connections
> * arrange it so that the stub is `Stop()`ped once connection is lost (not necessarily from the handler)
> * `Stop()` would cause the stub to reset enough of its internal state
> * eventually the plugin would call `Start()` on the stub again to reconnect, once this is safe to do

Hi! Klihub, thank you very much for your explanation! After these discussion, I think I understand your concern, that is, it's okay to make the stub on the client side restartable, but it's unsafe to let it restart "automatically".

At present, should I first modify the current patch to achieve **“Stop() would cause the stub to reset enough of its internal state”**, I think this is a good start.

klihub:
> > So then for a reconnectable plugin
> > 
> > * high level plugin logic would need to install an `OnClose()` handler to notice lost connections
> > * arrange it so that the stub is `Stop()`ped once connection is lost (not necessarily from the handler)
> > * `Stop()` would cause the stub to reset enough of its internal state
> > * eventually the plugin would call `Start()` on the stub again to reconnect, once this is safe to do
> 
> Hi! Klihub, thank you very much for your explanation! After these discussion, I think I understand your concern, that is, it's okay to make the stub on the client side restartable, but it's unsafe to let it restart "automatically".
> 
> At present, should I first modify the current patch to achieve **“Stop() would cause the stub to reset enough of its internal state”**, I think this is a good start.

Yes, I also think that would be the best to start with.

zhaodiaoer:
> > > So then for a reconnectable plugin
> > > 
> > > * high level plugin logic would need to install an `OnClose()` handler to notice lost connections
> > > * arrange it so that the stub is `Stop()`ped once connection is lost (not necessarily from the handler)
> > > * `Stop()` would cause the stub to reset enough of its internal state
> > > * eventually the plugin would call `Start()` on the stub again to reconnect, once this is safe to do
> > 
> > 
> > Hi! Klihub, thank you very much for your explanation! After these discussion, I think I understand your concern, that is, it's okay to make the stub on the client side restartable, but it's unsafe to let it restart "automatically".
> > At present, should I first modify the current patch to achieve **“Stop() would cause the stub to reset enough of its internal state”**, I think this is a good start.
> 
> Yes, I also think that would be the best to start with.

I modified the implementation according to the conclusion of the previous communication. There are some key modifications:

- make `stub.close()` can be called many times, and all variables that need to be reset when reconnect to runtime will be handled.
- `stub.Run(ctx)` no longer exits when the plugin service exits. instead, it runs in blocking manner and exits only when plugin service returns a **critical error** or the given context done. Before the `Run()` returned, the `stub.Start()` can be called repeatedly.
- fix the potential deadlock issue  in `stub.Wait()`.

Now, there are two ways in which the high level plugin logic uses the `stub`:

1. Call `stub.Run(ctx)` and give appropriate context, users can receive connection lost event via `onClose()` and decide whether to end the `Run()` or restart plugin via call `Start()`. simply, `stub.Start()` can be put in `onClose()` to achieve the effect of auto restart.
2. Call `stub.Start(ctx)` and call `stub.Wait()` explicitly, after `Wait()` returned, users can decide what to do next stop.

But at present, I still have some things that I haven't thought through: During the execution of the `stub.Run(ctx)` function, which error reports from the plugin service should be regarded as **critical error** and cause the `stub.Run(ctx)` function to return? Currently, my approach is to tolerate all errors defined by the ttrpc layer as recoverable. I don't know if this is reasonable.

zhaodiaoer:
@klihub can you give me more opinion, thanks.

pkg/stub/stub.go
```diff
@@ -319,7 +318,12 @@ func (stub *stub) Start(ctx context.Context) (retErr error) {
 	if stub.started {
 		return fmt.Errorf("stub already started")
 	}
-	stub.started = true
+	defer func() {
```
klihub, line 321: Why to do it this way ? AFAICT, there is only one exit path where we return a nil error, so it would be much easier to follow that we only set `stub.started` to `true` only on that path, if that happened in a more direct/less convoluted way. Also, all the other deferred functions are cleanups, so they have the opposite logic: they do something on failure, not on success. Moving this to be a direct assignment to the very end would keep that consistently intact.

pkg/stub/stub.go
```diff
@@ -441,22 +452,33 @@ func (stub *stub) Run(ctx context.Context) error {
...
+		errors.Is(err, ttrpc.ErrStreamClosed)
 }
 
-// Wait for the plugin to stop.
+// Wait for the plugin to stop, should be called after Start() or Run().
 func (stub *stub) Wait() {
-	stub.Lock()
-	if stub.srvErrC == nil {
-		return
+	if stub.started && stub.doneC != nil {
```
klihub, line 479: nit: can `stub.doneC == nil` ever, if `stub.started == true` ?

pkg/stub/stub.go
```diff
@@ -441,22 +452,33 @@ func (stub *stub) Run(ctx context.Context) error {
...
+			return nil
+		case err = <-stub.srvErrC:
+			if isRecoverableErr(err) {
+				log.Warnf(ctx, "Plugin service stopped", "error", err)
+				continue
+			}
+			return err
+		}
 	}
+}
```
klihub, line 467: I don't think this is a good idea. The plugin should be considered running while it is serving requests. Once it stops serving requests it should not be considered running, so `Run()` should return.

This looks like an attempt at a variation of your original scheme where you tried to make the stub internally restart itself. Now the stub tries to hide the fact that it is not running any more.

Instead what we should do IMO, and I think this is also what we discussed earlier, is to make the stub re-`Start()`-able. IOW, change it so, that calling `stub.Stop()` cleans up/resets the internal state of the plugin so, that a `stub.Start()` succeeds again (for externally launched plugins, for others this cannot be done).

klihub:
> > > > So then for a reconnectable plugin
> > > > 
> > > > * high level plugin logic would need to install an `OnClose()` handler to notice lost connections
> > > > * arrange it so that the stub is `Stop()`ped once connection is lost (not necessarily from the handler)
> > > > * `Stop()` would cause the stub to reset enough of its internal state
> > > > * eventually the plugin would call `Start()` on the stub again to reconnect, once this is safe to do
> > > 
> > > 
> > > Hi! Klihub, thank you very much for your explanation! After these discussion, I think I understand your concern, that is, it's okay to make the stub on the client side restartable, but it's unsafe to let it restart "automatically".
> > > At present, should I first modify the current patch to achieve **“Stop() would cause the stub to reset enough of its internal state”**, I think this is a good start.
> > 
> > 
> > Yes, I also think that would be the best to start with.
> 
> I modified the implementation according to the conclusion of the previous communication. There are some key modifications:
> 
> * make `stub.close()` can be called many times, and all variables that need to be reset when reconnect to runtime will be handled.

> * `stub.Run(ctx)` no longer exits when the plugin service exits. instead, it runs in blocking manner and exits only when plugin service returns a **critical error** or the given context done. Before the `Run()` returned, the `stub.Start()` can be called repeatedly.

I don't think this is a good idea and it is not what we discussed earlier.  The plugin should be considered running only when it is serving requests. Once it stops serving requests, it should not be considered running and `Run()` should return.

zhaodiaoer:
> > > > > So then for a reconnectable plugin
> > > > > 
> > > > > * high level plugin logic would need to install an `OnClose()` handler to notice lost connections
> > > > > * arrange it so that the stub is `Stop()`ped once connection is lost (not necessarily from the handler)
> > > > > * `Stop()` would cause the stub to reset enough of its internal state
> > > > > * eventually the plugin would call `Start()` on the stub again to reconnect, once this is safe to do
> > > > 
> > > > 
> > > > Hi! Klihub, thank you very much for your explanation! After these discussion, I think I understand your concern, that is, it's okay to make the stub on the client side restartable, but it's unsafe to let it restart "automatically".
> > > > At present, should I first modify the current patch to achieve **“Stop() would cause the stub to reset enough of its internal state”**, I think this is a good start.
> > > 
> > > 
> > > Yes, I also think that would be the best to start with.
> > 
> > 
> > I modified the implementation according to the conclusion of the previous communication. There are some key modifications:
> > 
> > * make `stub.close()` can be called many times, and all variables that need to be reset when reconnect to runtime will be handled.
> 
> > * `stub.Run(ctx)` no longer exits when the plugin service exits. instead, it runs in blocking manner and exits only when plugin service returns a **critical error** or the given context done. Before the `Run()` returned, the `stub.Start()` can be called repeatedly.
> 
> I don't think this is a good idea and it is not what we discussed earlier. The plugin should be considered running only when it is serving requests. Once it stops serving requests, it should not be considered running and `Run()` should return.

Frankly speaking, this is also the aspect that I have spent a considerable amount of time on but still haven't figured out how to handle better. The positioning of `Run()` is quite similar to that of `Start()`, except that unlike `Start()`, `Run()` can obtain the error returned by the service. However, I don't think this should be the key point to differentiate the two. Therefore, some modifications have been made to enhance the strength of `Run()` in terms of the attributes for managing the running cycle.

Regarding the restart requirement I initially proposed, the goal can now be achieved by using `Start()` and `Wait()` in combination. The modification of `Run()` is not necessary and this part of the modification can be reverted.

pkg/stub/stub.go
zhaodiaoer: you are right, changed

zhaodiaoer:
> I don't think this is a good idea and it is not what we discussed earlier. The plugin should be considered running only when it is serving requests. Once it stops serving requests, it should not be considered running and `Run()` should return.

@klihub How to optimize `Run()` is indeed worthy of long-term consideration. I have rolled back all the modifications about `Run()`. Thanks for your information.

pkg/stub/stub.go
```diff
@@ -137,7 +137,8 @@ type PostUpdateContainerInterface interface {
 
 // Stub is the interface the stub provides for the plugin implementation.
 type Stub interface {
-	// Run the plugin. Starts the plugin then waits for an error or the plugin to stop
+	// Run the plugin in blocked way and wait for critical error return from plugin service.
+	// before this function return, stopped plugin can be restart.
```
klihub, line 141: I think this wording is now a bit unclear (especially the 'blocked way' bit). Could we update this to be more clear ? Maybe using something like 
```
// Run starts the plugin then waits for the plugin service to exit, either due to a critical error or an explicit call to Stop(). Once Run() returns, the plugin can be restarted by calling Run() or Start() again.
```

klihub (changes requested):
@zhaodiaoer This looks now good to me, apart from a bit unclear godoc comment about Run(). I gave this a brief practical test this with a modified `template` plugin which internally `Stop()`ped itself periodically and otherwise sat in a `Run()` loop. It seemed to work fine, although I did not try to torture it much.

Could you still fix/update that unclear comment ?

zhaodiaoer:
> @zhaodiaoer This looks now good to me, apart from a bit unclear godoc comment about Run(). I gave this a brief practical test this with a modified `template` plugin which internally `Stop()`ped itself periodically and otherwise sat in a `Run()` loop. It seemed to work fine, although I did not try to torture it much.
> 
> Could you still fix/update that unclear comment ?

sure, let me do that

pkg/stub/stub.go
zhaodiaoer: already update, thanks

klihub (approved):
LGTM.

zhaodiaoer:
kindly ping @fuweid

pkg/stub/stub.go
```diff
@@ -449,14 +457,11 @@ func (stub *stub) Run(ctx context.Context) error {
 	return err
 }
 
-// Wait for the plugin to stop.
+// Wait for the plugin to stop, should be called after Start() or Run().
 func (stub *stub) Wait() {
-	stub.Lock()
-	if stub.srvErrC == nil {
-		return
+	if stub.started {
```
mikebrow, line 462: pls provide a getstarted state api... stub.Lock() started:=stub.started.. sub.Unlock() return started... at least for use when not locked..

zhaodiaoer: > pls provide a getstarted state api... stub.Lock() started:=stub.started.. sub.Unlock() return started... at least for use when not locked..

If i haven‘t misunderstood, the `getstarted` api is expect to replace all `if stub.started` code? currently many started check exist in context that have hold the lock, if provide a `getstarted` state api now may make deadlock   more likely to occur, maybe we need more consideration about this, and solve it in a new pr?
Or if your concern just the `if stub.started` without lock in `Wait()`, check `stub.started` is just a protection for not use `Wait()` before the first `Start()` or `Run()`, or else `Wait()` may try to read nil `stub.doneC`: 

- if call `Wait()` before the first `Start()` and `Run()`: the `stub.started` is must `false` and `Wait()` will not wait for the nil `stub.doneC`
- if call `Wait()` after the first `Start()` or `Run()`: there is no chance `stub.doneC` will change back to `nil`, so wait for it is safely

klihub: For your locking vs. context of call problem you could go the conventional way of sliptting it into two functions: 

```
// IsStarted returns true if the plugin has been started either by Start() or by Run().
func (s *Stub) IsStarted() bool {
    s.Lock()
    defer s.Unlock()
    return s.isStarted()
}

func (s *Stub) isStarted() bool {
    return s.started()
}
```

Then you can simply replace your current internal `if s.started` checks in an already locked context with corresponding `if s.isStarted()` checks.

klihub:
@fuweid If you have some spare cycles, PTAL.

pkg/stub/stub.go
mikebrow: this ^

zhaodiaoer: OK and Done.

pkg/stub/stub.go
```diff
@@ -401,6 +401,7 @@ func (stub *stub) Start(ctx context.Context) (retErr error) {
 
 	log.Infof(ctx, "Started plugin %s...", stub.Name())
 
+	stub.started = true
```
fuweid, line 404: Line 398 is still working? We can fix it in the follow-up.

```
if err = <-stub.cfgErrC; err != nil {
		return err
	}

```

pkg/stub/stub.go
```diff
@@ -316,10 +316,10 @@ func (stub *stub) Start(ctx context.Context) (retErr error) {
 	stub.Lock()
 	defer stub.Unlock()
 
-	if stub.started {
+	if stub.isStarted() {
 		return fmt.Errorf("stub already started")
 	}
-	stub.started = true
+	stub.doneC = make(chan struct{})
```
fuweid, line 322: In Line 382, could we wait for it if `retErr` isn't nil? 
The goroutine uses reference so that new `Start` would refresh `stub.doneC` and we may have two goroutines to close one channel. We can consider to drain the goroutine if `retErr` != nil.

```
go func() {
		stub.srvErrC <- rpcs.Serve(ctx, rpcl)
		close(stub.doneC)
	}()
```

pkg/stub/stub.go
zhaodiaoer: In my understanding and testing, is not affected by this update, what is the concern?

pkg/stub/stub.go
fuweid: Sorry for late reply.

If `Start` returns error, the following goroutine could be still running and it doesn't reach the `close(stub.doneC)`. And then new `stub.Start()` call will create a new `stub.doneC` channel. If the previous goroutine calls the `close(stub.doneC)` after `stub.doneC = make(chan struct{})`, the `stub.doneC` can be closed twice and then panic. 

```go
go func() {
		stub.srvErrC <- rpcs.Serve(ctx, rpcl)
		close(stub.doneC)
}()
```

For the above goroutine, we can consider to pass var instead of using reference directly.

```go
go func(rpcl xxx, doneC chan struct{}, srvErrC chan error) {
       srvErrC <- rpcs.Serve(ctx, rpcl)
       close(doneC)
}(rpcl, stub.doneC, stub.srvErrC)
```

To be honest, it's hard to reproduce without adding failpoint(sleep). However, we should ensure logic is clean and easy to reason.

zhaodiaoer: Thanks for your explanation! you are right, there definitely exists a chance that double goroutine closes one `stub.doneC`;
Does define a separate function like `stub.serve(...)` better?

fuweid: > Does define a separate function like stub.serve(...) better?

It sounds good to me. You can also use anonymous function as it is right now.

pkg/stub/stub.go
zhaodiaoer: I think before we need to let `stub.Configure()` may been called more than once, and these code are must run after `stub.register()`, we can let it remain as it is for the time being, the currently implementation of stub's restart is depend that `stub.register()` is executed every time after restart .

pkg/stub/stub.go
zhaodiaoer: updated.

fuweid (approved):
LGTM

Thanks

mikebrow (approved):
LGTM

PR #94, opened 2024-07-10
Add support for adjusting OOM score adjustment.

### LINKED ISSUE #92: Support adjust oom_Score_adj for NRI plugins
The LinuxContainer structure transferred by the containerd contains the OomScoreAdj field. However, the NRI plugins do not support the adjustment of this field. Theoretically, the OomScoreAdj field should be customized by users to provide more refined QoS control.

The definition of LinuxContainer: 
```
type LinuxContainer struct {
	state         protoimpl.MessageState
	sizeCache     protoimpl.SizeCache
	unknownFields protoimpl.UnknownFields

	Namespaces  []*LinuxNamespace `protobuf:"bytes,1,rep,name=namespaces,proto3" json:"namespaces,omitempty"`
	Devices     []*LinuxDevice    `protobuf:"bytes,2,rep,name=devices,proto3" json:"devices,omitempty"`
	Resources   *LinuxResources   `protobuf:"bytes,3,opt,name=resources,proto3" json:"resources,omitempty"`
	OomScoreAdj *OptionalInt      `protobuf:"bytes,4,opt,name=oom_score_adj,json=oomScoreAdj,proto3" json:"oom_score_adj,omitempty"`
	CgroupsPath string            `protobuf:"bytes,5,opt,name=cgroups_path,json=cgroupsPath,proto3" json:"cgroups_path,omitempty"`
}
```

The definition of LinuxContainerAdjustment:
```
type LinuxContainerAdjustment struct {
	state         protoimpl.MessageState
	sizeCache     protoimpl.SizeCache
	unknownFields protoimpl.UnknownFields

	Devices     []*LinuxDevice  `protobuf:"bytes,1,rep,name=devices,proto3" json:"devices,omitempty"`
	Resources   *LinuxResources `protobuf:"bytes,2,opt,name=resources,proto3" json:"resources,omitempty"`
	CgroupsPath string          `protobuf:"bytes,3,opt,name=cgroups_path,json=cgroupsPath,proto3" json:"cgroups_path,omitempty"`
}
```
Obviously, losing the ability to oom adjust is not reasonable.

zxj874478410:
The OomScoreAdj field is added to the LinuxContainerAdjustment structure of the NRI plugin. This field records the value of oom_score_adj that needs to be adjusted during CreateContainer. Added the appropriate set method and the method to write back to the NRI response.
Fix #92 

**TEST METHOD：**
We build a NRI plugin with CreateContainer function like this:
```
func (p *plugin) CreateContainer(_ context.Context, pod *api.PodSandbox, ctr *api.Container) (*api.ContainerAdjustment, []*api.ContainerUpdate, error) {
	log.Infof("Creating container %s/%s/%s...", pod.GetNamespace(), pod.GetName(), ctr.GetName())

	//
	// This is the container creation request handler. Because the container
	// has not been created yet, this is the lifecycle event which allows you
	// the largest set of changes to the container's configuration, including
	// some of the later immautable parameters. Take a look at the adjustment
	// functions in pkg/api/adjustment.go to see the available controls.
	//
	// In addition to reconfiguring the container being created, you are also
	// allowed to update other existing containers. Take a look at the update
	// functions in pkg/api/update.go to see the available controls.
	//

	adjustment := &api.ContainerAdjustment{}

	adjustment.SetLinuxOomScoreAdj(121)
	log.Infof("set oom score: %v", 121)
	return adjustment, nil, nil
}
```
Then run this NRI plugin with containerd. Create a k8s pod of apache.
```
[root@master bin]# kubectl get po -A |grep apache
kube-system    apache-66f99dd558-c9bgm          1/1     Running   0               4h5m
```
Run the docker inspect command to query the process ID of the Apache container is **3445531**.

At last, query the oom_score_adj of this pid. It has been set to 121 successfully.
```
[root@master bin]# cat /proc/3445531/oom_score_adj
121
```


pkg/api/api.proto
```diff
@@ -25,17 +25,17 @@ option go_package = "github.com/containerd/nri/pkg/api;api";
 // only covers plugin registration and unsolicited container updates.
 // The rest of the API is defined by the Plugin service.
 service Runtime {
-    // RegisterPlugin registers the plugin with the runtime.
-    rpc RegisterPlugin(RegisterPluginRequest) returns (Empty);
-    // UpdateContainers requests unsolicited updates to a set of containers.
-    rpc UpdateContainers(UpdateContainersRequest) returns (UpdateContainersResponse);
```
klihub, line 31: nit: can we get rid of these unrelated changes ? AFAICT, they just change tabulation.

pkg/api/api.proto
```diff
@@ -25,17 +25,17 @@ option go_package = "github.com/containerd/nri/pkg/api;api";
...
+  // UpdateContainers requests unsolicited updates to a set of containers.
+  rpc UpdateContainers(UpdateContainersRequest) returns (UpdateContainersResponse);
 }
 
 message RegisterPluginRequest {
-    // Name of the plugin to register.
-    string plugin_name = 1;
-    // Plugin invocation index. Plugins are called in ascending index order.
-    string plugin_idx = 2;
+  // Name of the plugin to register.
```
klihub, line 35: nit: can we get rid of these unrelated changes ? AFAICT, they just change tabulation.

pkg/runtime-tools/generate/generate.go
```diff
@@ -309,6 +310,13 @@ func (g *Generator) AdjustCgroupsPath(path string) {
 	}
 }
 
+// AdjustOomScoreAdj adjusts the cgroup pseudofs path in the OCI Spec.
```
klihub, line 313: Please fix this comment. It could be something like:
```
// AdjustOomScoreAdj adjusts the kernel's Out-Of-Memory (OOM) killer score for the container
```

I don't know if most folks are generally aware that the kubelet also actively sets/adjusts the OOM Score adjustment field for containers based on the QoS class (and IIRC memory requests for Burstable containers). If not, maybe also a word of warning here warranted. @samuelkarp WDYT ?

klihub (changes requested):
Thank's for the PR @zxj874478410 ! I have a few nits, but otherwise it looks good to me.

pkg/runtime-tools/generate/generate.go
chenzhiwei: We extended the k8s qos classes to support more workloads, in our implementation, we will override the Kubelet's value when k8s qos class is Burstable, and respect Guaranteed and BestEffort qos classes.

zxj874478410:
> Thank's for the PR @zxj874478410 ! I have a few nits, but otherwise it looks good to me.

Thanks for the review @klihub . I've fixed these faults.

pkg/api/adjustment.go
```diff
@@ -253,6 +253,12 @@ func (a *ContainerAdjustment) SetLinuxCgroupsPath(value string) {
 	a.Linux.CgroupsPath = value
 }
 
+// SetLinuxOomScoreAdj records setting the cgroups path for a container.
```
chenzhiwei, line 256: please also update this comment

zxj874478410: Already update it.

chenzhiwei (approved):
lgtm

klihub (approved):

pkg/runtime-tools/generate/generate.go
```diff
@@ -309,6 +310,14 @@ func (g *Generator) AdjustCgroupsPath(path string) {
 	}
 }
 
+// AdjustOomScoreAdj adjusts the kernel's Out-Of-Memory (OOM) killer score for the container.
+// This may override kubelet's settings for OOM score.
+func (g *Generator) AdjustOomScoreAdj(score *nri.OptionalInt) {
+	if score != nil {
+		g.SetProcessOOMScoreAdj(int(score.Value))
+	}
+}
```
mikebrow, line 319:
```suggestion
} else {
// need to nil out the optional oom here... ?
}
```

mikebrow (commented):
wondering if we should be able to nil out the oooscoreadj

pkg/api/adjustment.go
```diff
@@ -253,6 +253,12 @@ func (a *ContainerAdjustment) SetLinuxCgroupsPath(value string) {
 	a.Linux.CgroupsPath = value
 }
 
+// SetLinuxOomScoreAdj records setting the kernel's Out-Of-Memory (OOM) killer score for a container.
+func (a *ContainerAdjustment) SetLinuxOomScoreAdj(value int) {
```
zxj874478410, line 257: Maybe we could change that like this:
```suggestion
	func (a *ContainerAdjustment) SetLinuxOomScoreAdj(value *int) {
```
In this way, we can unset/ nil the oomscoreadj with the change of generate.go. @mikebrow

pkg/runtime-tools/generate/generate.go
```diff
@@ -309,6 +310,14 @@ func (g *Generator) AdjustCgroupsPath(path string) {
 	}
 }
 
+// AdjustOomScoreAdj adjusts the kernel's Out-Of-Memory (OOM) killer score for the container.
+// This may override kubelet's settings for OOM score.
+func (g *Generator) AdjustOomScoreAdj(score *nri.OptionalInt) {
+	if score != nil {
+		g.SetProcessOOMScoreAdj(int(score.Value))
+	}
```
zxj874478410, line 318: Thanks for the suggestion! After we set OomScoreAdj to nil, We can handle the nil here.
I try to call the func g.initConfigProcess() by calling func g.SetProcessOOMScoreAdj(0), and then set g.Config.Process.OOMScoreAdj to nil.

```suggestion
	} else {
		g.SetProcessOOMScoreAdj(0)
		g.Config.Process.OOMScoreAdj = nil
	}
```

pkg/api/adjustment.go
mikebrow: right .. let's go that route I think.

pkg/runtime-tools/generate/generate.go
mikebrow: sgtm  let's try it / test it out

zxj874478410: We tested it with the following code
```	
oomScoreAdj:= 555
adjustment.SetLinuxOomScoreAdj(&oomScoreAdj)
adjustment.SetLinuxOomScoreAdj(nil)
```
The second set unset the oomscoreadj with the api. Finally, the oom_score_adj is set to 996 by kubelet.

> cat /proc/2213925/oom_score_adj
996
This is work! @mikebrow

pkg/runtime-tools/generate/generate_suite_test.go
```diff
@@ -322,6 +344,15 @@ func withMemorySwap(v int64) specOption {
 	}
 }
 
+func withOomScoreAdj(v int) specOption {
```
mikebrow, line 347: nit... also make v an optional pointer to int here in this with test func and

spec.Process.OOMScoreAdj = Int(value) // using Int(value) from ./options.go to optionally allocate a pointer to normalized copy of value


and add a test case that uses nil down in the suite test  pls :)

zxj874478410: v has become a pointer. The test case to nil down ha been added :)

mikebrow (approved):
LGTM

pkg/runtime-tools/generate/generate_suite_test.go
mikebrow: thx!****

zxj874478410:
@klihub @mikebrow @kzys Just rebased code and fixed conflicts, please take a look, thanks.

zxj874478410:
> LGTM

zxj874478410:
@klihub @mikebrow Sorry for conflict again. Just fixed conflicts, please take a look, thanks.

klihub:
@zxj874478410 This fix is not sufficient in it's current form as it breaks replacing a mount with another one to the same destination. I think we'd need a bit of additional code to prevent that from happening. [Maybe something like this](https://github.com/klihub/nri/tree/fixes/mount-removal-adjustment), where 8f509e7d0f2e4e9af7f7acb77e9c18f2e882efaf is an updated version of your commit, and 7ea0a47f00f853b4e970b980b1a24b082c0930f0 adds a new test case for mount removal.

zxj874478410:
> @zxj874478410 This fix is not sufficient in it's current form as it breaks replacing a mount with another one to the same destination. I think we'd need a bit of additional code to prevent that from happening. [Maybe something like this](https://github.com/klihub/nri/tree/fixes/mount-removal-adjustment), where [8f509e7](https://github.com/containerd/nri/commit/8f509e7d0f2e4e9af7f7acb77e9c18f2e882efaf) is an updated version of your commit, and [7ea0a47](https://github.com/containerd/nri/commit/7ea0a47f00f853b4e970b980b1a24b082c0930f0) adds a new test case for mount removal.

Thanks for the detailed instructions, but does this problem refer to another pr #87 ? This one refers to OOM score adjustment.

klihub:
> > @zxj874478410 This fix is not sufficient in it's current form as it breaks replacing a mount with another one to the same destination. I think we'd need a bit of additional code to prevent that from happening. [Maybe something like this](https://github.com/klihub/nri/tree/fixes/mount-removal-adjustment), where [8f509e7](https://github.com/containerd/nri/commit/8f509e7d0f2e4e9af7f7acb77e9c18f2e882efaf) is an updated version of your commit, and [7ea0a47](https://github.com/containerd/nri/commit/7ea0a47f00f853b4e970b980b1a24b082c0930f0) adds a new test case for mount removal.
> 
> Thanks for the detailed instructions, but does this problem refer to another pr #87 ? This one refers to OOM score adjustment.

No, it refers to this 

> > @zxj874478410 This fix is not sufficient in it's current form as it breaks replacing a mount with another one to the same destination. I think we'd need a bit of additional code to prevent that from happening. [Maybe something like this](https://github.com/klihub/nri/tree/fixes/mount-removal-adjustment), where [8f509e7](https://github.com/containerd/nri/commit/8f509e7d0f2e4e9af7f7acb77e9c18f2e882efaf) is an updated version of your commit, and [7ea0a47](https://github.com/containerd/nri/commit/7ea0a47f00f853b4e970b980b1a24b082c0930f0) adds a new test case for mount removal.
> 
> Thanks for the detailed instructions, but does this problem refer to another pr #87 ? This one refers to OOM score adjustment.

Argh... Sorry about that. Yes, you are absolutely right, it is about PR #87. Time to renew my glasses...

klihub (approved):

mikebrow (approved):
LGTM

pkg/runtime-tools/generate/generate.go
samuelkarp: This code path appears to be the source of https://github.com/containerd/nri/issues/117.  @mikebrow, do we need to support a plugin both setting an explicit adjustment and then clearing its own adjustment?

zxj874478410:
这是来自QQ邮箱的假期自动回复邮件。
 
您好，我最近正在休假中，无法亲自回复您的邮件。我将在假期结束后，尽快给您回复。

pkg/runtime-tools/generate/generate.go
mikebrow: nod.. while I think it's a valid path to unset given 0 / 999 etc.. is not unset.. we need to ensure we are not accidentally unsetting a valet oom from kubelet. @klihub FYI

mikebrow: manual:  

``` OOMScoreAdjust=
           Sets the adjustment value for the Linux kernel's
           Out-Of-Memory (OOM) killer score for executed processes.
           Takes an integer between -1000 (to disable OOM killing of
           processes of this unit) and 1000 (to make killing of
           processes of this unit under memory pressure very likely).
           See The /proc Filesystem[6] for details. If not specified
           defaults to the OOM score adjustment level of the service
           manager itself, which is normally at 0.
```
note behavior normally 0 when unset, but is configurable making unset a unique state

PR #98, opened 2024-07-10
api: add support for NRI-native CDI injection

klihub:
Add support for native CDI injection through NRI. With this patch set in place, a plugin can now directly request injection of CDI devices by fully qualified CDI device names. This is considerably simpler than what is currently required from an NRI plugin to perform CDI injection: perform CDI device resolution, then injection to an (almost) empty OCI Spec, inspect the resulting OCI Spec and convert the differences to a set of low-level NRI adjustments for devices, mounts, env. vars. and OCI hooks to effectively do the same through the NRI protocol.

/cc @elezar


pkg/runtime-tools/generate/generate.go
```diff
@@ -103,6 +111,9 @@ func (g *Generator) Adjust(adjust *nri.ContainerAdjustment) error {
 	g.AdjustEnv(adjust.GetEnv())
 	g.AdjustHooks(adjust.GetHooks())
 	g.AdjustDevices(adjust.GetLinux().GetDevices())
+	if err := g.InjectCDIDevices(adjust.GetLinux().GetCDIDevices()); err != nil {
```
elezar, line 114: Question: Should CDI devices be under `Linux`? They are in theory platform independent with their modifications being platform specific.

pkg/runtime-tools/generate/generate.go
```diff
@@ -323,6 +334,20 @@ func (g *Generator) AdjustDevices(devices []*nri.LinuxDevice) {
 	}
 }
 
+// InjectCDIDevices injects CDI devices to the OCI Spec.
```
elezar, line 337: It's not clear how this injects the device into the OCI spec specifically.

plugins/device-injector/README.md
```diff
@@ -1,6 +1,6 @@
 ## Device Injector Plugin
 
-This sample plugin can inject devices and mounts into containers using
+This sample plugin can inject devices, CDI devices and mounts into containers using
```
elezar, line 3:
```suggestion
This sample plugin can inject Linux device nodes, CDI devices, and mounts into containers using
```

pkg/adaptation/result.go
```diff
@@ -90,7 +90,8 @@ func collectCreateContainerResult(request *CreateContainerRequest) *result {
 				Hooks:       &Hooks{},
 				Rlimits:     []*POSIXRlimit{},
 				Linux: &LinuxContainerAdjustment{
-					Devices: []*LinuxDevice{},
+					Devices:    []*LinuxDevice{},
+					CDIDevices: []*CDIDevice{},
```
elezar, line 94: `CDIDevices` should be at the same level as `Mounts` and envs. They are not Linux-specific.

pkg/runtime-tools/generate/generate.go
klihub: Updated comment to be more specific about how actual CDI device injection happens.

plugins/device-injector/README.md
klihub: Updated accordingly.

pkg/adaptation/result.go
klihub: Updated accordingly.

pkg/adaptation/result.go
```diff
@@ -386,6 +390,36 @@ func (r *result) adjustDevices(devices []*LinuxDevice, plugin string) error {
...
+	}
+
+	// Notes:
+	//   CDI devices are opaque references, typically to vendor specific
+	//   devices. They get resolved to actual devices and potential related
+	//   mounts, environment variables, etc. in the runtime. Unlike with
+	//   devices, we only allow CDI devices references to be added to a
+	//   container, not removed. We pass them unresolved to the runtime and
+	//   have them resolved there. Also unlike with devices, we don't include
+	//   CDI device references in a creation requests. However, since there
```
elezar, line 405: What is the scope at which these multiple references are checked.

It is, for example, perfectly fine for multiple containers to request the same device. In the case of a single container requesting the same device multiple times, this could also be seen as an ok use-case that could be solved by some kind of deduplication.

klihub: With NRI and a single NRI request, (potential) multiplicity is present at the NRI plugin level, not the container level. With a few exceptions, a request is always about a single pod or a single container. However, there might be multiple plugins acting on the same request and applying modifications to the container via the NRI-provided interfaces. What we currently check is that only a single NRI plugin injects the same CDI device (using the same by fully qualified CDI device name) to the container. Multiple plugins attempting the same is flagged as a conflict/error.

elezar: OK. Like I said, I'm happy to keep this as is for now, but at some point we may want to consider relaxing this restriction. In the end the CDI modifications are applied by the runtime and since CDI device names are locally unique, the end result should be the same.

elezar:
This change would make https://github.com/containers/nri-plugins/pull/327 redundant. Is there a way to detect that we're working with a runtime / plugin that supports CDI devices natively?

plugins/device-injector/device-injector.go
```diff
@@ -162,6 +190,36 @@ func parseDevices(ctr string, annotations map[string]string) ([]device, error) {
...
+			annotation = []byte(value)
+			break
+		}
+	}
+
+	if annotation == nil {
+		return nil, nil
+	}
+
+	if err := yaml.Unmarshal(annotation, &cdiDevices); err != nil {
```
fuweid, line 216: would you please add one test for multiple values in `annotation`? I don't follow `yaml.Unmarshal` into `[]string` here. Thanks!

plugins/device-injector/README.md
```diff
@@ -26,6 +26,33 @@ The annotation syntax for device injection is
 
 `file_mode`, `uid` and `gid` can be omitted, the rest are mandatory.
 
+### CDI Device Annotations
+
+Devices are annotated using the `cdi-devices.nri.io` annotation key prefix.
```
yylt, line 31: It seems that no additional attributes are needed here; wouldn't using a comma-separated format be more convenient?

klihub: Would definitely be possible but it is a bit subjective which one is more convenient. I prefer keeping device injection and CDI device injection syntactically as close to each other as possible here. Also, the CRI protocol uses a `struct { Name string }` for passing CDI devices to inject and this NRI patch set follows suit. That allows for future extensions with additional per-device attributes. The current strightforward struct-marshalled-to-yaml annotation syntax can be trivially extended to support such extensions (and it is also syntactically forward-compatible). A comma-separated syntax, however, would need to be abandoned in favor of the current syntax if ever such extensions were introduced.

klihub:
> This change would make [containers/nri-plugins#327](https://github.com/containers/nri-plugins/pull/327) redundant.

I rolled this PR actually with that CDI injecting NRI plugin in mind (I mentioned this in the initial review of that plugin PR). The idea was to test/see how we could do native CDI injection through NRI. And, if we're happy enough with the result to merge it, to make the implementation of actual CDI device injection in that plugin much easier.

There are both pros and cons with this approach. The pros are
- no need to track CDI evolution so closely (even a recompile is not necessary when CDI is extended)
- no need to reimplement existing CDI injection logic in terms of NRI adjustments
- does not tie CDI capabilities to NRI capabilities (technically we can allow such container mutation capabilities to CDI which we don't want to allow in NRI)

The cons are maybe that it is a bit superficial, as currently the container mutation capabilities offered by NRI are a superset of what is needed for CDI injection.

> Is there a way to detect that we're working with a runtime / plugin that supports CDI devices natively?

Currently no other means than checking `RuntimeName` and `RuntimeVersion` in the plugin configuration NRI message, then using explicit knowledge about which runtime/version support started appearing in.

fuweid (approved):
LGTM

pkg/runtime-tools/generate/generate.go
```diff
@@ -323,6 +335,23 @@ func (g *Generator) AdjustDevices(devices []*nri.LinuxDevice) {
...
+	if len(devices) == 0 || g.injectCDIDevices == nil {
+		return nil
+	}
+
+	names := []string{}
+	for _, d := range devices {
+		names = append(names, d.Name)
+	}
+
+	return g.injectCDIDevices(g.Config, names)
```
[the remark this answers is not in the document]
klihub: You mean, check that `g.injectCDIDevices != nil` ? That check is at the beginning of the function.

yylt (approved):

plugins/device-injector/device-injector.go
```diff
@@ -104,6 +107,31 @@ func (p *plugin) CreateContainer(_ context.Context, pod *api.PodSandbox, contain
...
 	}
 
+	// inject CDI devices to container
+	cdiDevices, err = parseCDIDevices(container.Name, pod.Annotations)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	if len(cdiDevices) == 0 {
+		log.Infof("%s: no CDI devices annotated...", ctrName)
```
elezar, line 117: Do we really want to log at info for every container that has no CDI devices?

plugins/device-injector/device-injector.go
```diff
@@ -104,6 +107,31 @@ func (p *plugin) CreateContainer(_ context.Context, pod *api.PodSandbox, contain
 		}
 	}
 
+	// inject CDI devices to container
+	cdiDevices, err = parseCDIDevices(container.Name, pod.Annotations)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	if len(cdiDevices) == 0 {
```
elezar, line 116: Does an early return for this case make sense? It would remove the need to have the injection code in an else block.

plugins/device-injector/device-injector_test.go
```diff
@@ -0,0 +1,70 @@
...
+`,
+			},
+			result: []string{
+				"vendor0.com/device=null",
+				"vendor0.com/device=zero",
+			},
+		},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			devices, err := parseCDIDevices("ctr0", tc.annotations)
```
elezar, line 65: Does it make sense to add a test for an annotation not matching the container name? We also don't strictly speaking cover the other two allowed annotation keys.

elezar (commented):
Thanks @klihub 

This looks fine at a high level now. My last comments are effectively nits or notes on unit test coverage.

pkg/adaptation/result.go
mikebrow: yeah some sort of allow plugin overrides (conflicts)

plugins/device-injector/README.md
```diff
@@ -26,6 +26,33 @@ The annotation syntax for device injection is
...
+    cdi-devices.nri.io/container.c1: |
+      - vendor0.com/device=zero
+    cdi-devices.nri.io/container.mgmt: |
+      - vendor0.com/device=all
+```
+
+requests the injection of the vendor0.com/device=null, vendor0.com/device=zero,
+and vendor0.com/device=all CDI devices, to the c0, c1, and mgmt containers of
+the pod.
+
```
mikebrow, line 55: maybe another example showing value with a plurality of devices to one container, maybe one from another vendor

plugins/device-injector/README.md
```diff
@@ -26,6 +26,33 @@ The annotation syntax for device injection is
 
 `file_mode`, `uid` and `gid` can be omitted, the rest are mandatory.
 
+### CDI Device Annotations
+
+Devices are annotated using the `cdi-devices.nri.io` annotation key prefix.
+The key `cdi-devices.nri.io/container.$CONTAINER_NAME` annotates CDI devices
+to be injected into `$CONTAINER_NAME`. The keys `cdi-devices.nri.io` and
+`cdi-devices.nri.io/pod` annotate CDI devices to be injected into all
```
mikebrow, line 34: reasoning for both with and without /pod?

mikebrow (commented):
some comments, cheers!

plugins/device-injector/device-injector.go
klihub: We can't really, because the same function later might still inject mounts if those were annotated.

plugins/device-injector/device-injector.go
```diff
@@ -104,6 +107,31 @@ func (p *plugin) CreateContainer(_ context.Context, pod *api.PodSandbox, contain
...
+		log.Debugf("%s: no CDI devices annotated...", ctrName)
+	} else {
+		if verbose {
+			dump(ctrName, "annotated CDI devices", devices)
```
elezar, line 120: Should this be `cdiDevices`?

plugins/device-injector/device-injector.go
elezar: OK. I see now. I think separate functions would be more readable, but I'm also happy for that to be a follow-up.

plugins/device-injector/device-injector.go
```diff
@@ -90,7 +93,7 @@ func (p *plugin) CreateContainer(_ context.Context, pod *api.PodSandbox, contain
 	}
 
 	if len(devices) == 0 {
-		log.Infof("%s: no devices annotated...", ctrName)
+		log.Debugf("%s: no devices annotated...", ctrName)
```
elezar, line 96: Should we make the same change for `mounts`?

plugins/device-injector/device-injector.go
```diff
@@ -162,6 +190,36 @@ func parseDevices(ctr string, annotations map[string]string) ([]device, error) {
...
 }
 
+func parseCDIDevices(ctr string, annotations map[string]string) ([]string, error) {
+	var (
+		key        string
+		annotation []byte
+		cdiDevices []string
+	)
+
+	// look up effective device annotation and unmarshal devices
```
elezar, line 200: Refactor idea: Getting the list of annotations is common to devices, mounts, and CDI devices and should be factored out into a separate function.

plugins/device-injector/device-injector.go
```diff
@@ -162,6 +190,36 @@ func parseDevices(ctr string, annotations map[string]string) ([]device, error) {
...
+	for _, key = range []string{
+		cdiDeviceKey + "/container." + ctr,
+		cdiDeviceKey + "/pod",
+		cdiDeviceKey,
+	} {
+		if value, ok := annotations[key]; ok {
+			annotation = []byte(value)
+			break
```
elezar, line 208: Is it documented that we only support a single annotation style?

plugins/device-injector/device-injector.go
klihub: Refactored to use separate functions, a common function for picking annotations, and to use an early return to omit a final else-block in the injection code paths.

plugins/device-injector/device-injector.go
klihub: Yes, it should have been. It has been fixed while refactoring to use multiple injection functions.

plugins/device-injector/device-injector.go
klihub: Fixed in the refactored function.

plugins/device-injector/device-injector.go
klihub: Implemented in the refactored version.

pkg/adaptation/adaptation_suite_test.go
```diff
@@ -488,6 +488,13 @@ var _ = Describe("Plugin container creation adjustments", func() {
 		case "rlimit":
 			a.AddRlimit("nofile", 456, 123)
 
+		case "CDI device":
```
elezar, line 491: nit: Seems strange that this is a space-separated string.

pkg/adaptation/adaptation_suite_test.go
```diff
@@ -2008,6 +2025,12 @@ func stripLinuxResourcesUnified(u map[string]string) map[string]string {
 	return u
 }
 
+func stripCDIDevices(a *api.ContainerAdjustment) {
```
elezar, line 2028: Just noticed that this seems a bit different to the other functions here since we don't have a return value.

OK. I think the reason it looks strange is that it's not adjacent to the other `strip*` functions for `ContainerAdjustment`. Could we move it?

pkg/adaptation/api.go
```diff
@@ -78,6 +78,7 @@ type (
 	LinuxMemory              = api.LinuxMemory
 	LinuxDevice              = api.LinuxDevice
 	LinuxDeviceCgroup        = api.LinuxDeviceCgroup
+	CDIDevice                = api.CDIDevice
```
elezar, line 81: Not that's it's too important, but should we consider sorting this block?

pkg/api/adjustment.go
```diff
@@ -129,6 +129,12 @@ func (a *ContainerAdjustment) RemoveDevice(path string) {
 	})
 }
 
+// AddCDIDevice records the addition of the given CDI device to a container.
+func (a *ContainerAdjustment) AddCDIDevice(d *CDIDevice) {
+	a.initLinux()
```
elezar, line 134: I think this can be removed since `CDIDevices` is no longer in the `linux` struct.

pkg/runtime-tools/generate/generate.go
```diff
@@ -325,6 +337,23 @@ func (g *Generator) AdjustDevices(devices []*nri.LinuxDevice) {
 	}
 }
 
+// InjectCDIDevices injects the given CDI devices into the OCI Spec.
+// Devices are given by their fully qualified CDI device names. The
```
elezar, line 341: Could we use `requested` or `specified` instead of `given`?

plugins/device-injector/README.md
klihub: Updated the example to include also injection of multiple CDI devices into a single container.

plugins/device-injector/device-injector_test.go
klihub: Added test for non-matching annotation and the missing tests for the other annotation parsing functions.

pkg/runtime-tools/generate/generate.go
```diff
@@ -103,6 +112,9 @@ func (g *Generator) Adjust(adjust *nri.ContainerAdjustment) error {
 	g.AdjustEnv(adjust.GetEnv())
 	g.AdjustHooks(adjust.GetHooks())
 	g.AdjustDevices(adjust.GetLinux().GetDevices())
+	if err := g.InjectCDIDevices(adjust.GetCDIDevices()); err != nil {
```
elezar, line 115: Let's move this so that we're dealing with the `GetLinux()` adjustments separately / consecutively.

plugins/device-injector/README.md
```diff
@@ -1,16 +1,19 @@
...
+containers using pod annotations.
 
 ### Device Annotations
 
 Devices are annotated using the `devices.nri.io` annotation key prefix.
 The key `devices.nri.io/container.$CONTAINER_NAME` annotates devices to
 be injected into `$CONTAINER_NAME`. The keys `devices.nri.io` and
-`devices.nri.io/pod` annotate devices to be injected into all containers.
+`devices.nri.io/pod` annotate devices to be injected into containers
+without any other, container-specific device annotations. Only one of
```
elezar, line 12: Is the unspecified nature of the ordering not for all three annotations?

plugins/device-injector/device-injector.go
```diff
@@ -67,131 +69,183 @@ type plugin struct {
...
-		return nil, fmt.Errorf("invalid mount annotation %q: %w", key, err)
+		return nil, fmt.Errorf("invalid mount annotation %q: %w", string(annotation), err)
 	}
 
 	return mounts, nil
 }
 
+func getAnnotation(annotations map[string]string, mainKey, ctr string) []byte {
+	for _, key := range []string{
+		mainKey + "/container." + ctr,
```
elezar, line 237: Note: The doc indidcates that the order is non-deterministic, but his range over a slice means that we have a clear order or precedence.

elezar (commented):
Thanks @klihub.

I think I'm down to nits and organisational questions / comments for the most part. Thanks for the cleanups. (for some of them it may be clearer if we have the refactorings as seprate commits / PRs).

plugins/device-injector/README.md
klihub: We've used in some of our existing components a similar annotation syntax/precedence. There the reasoning was to try provide both regularly symmetric annotation syntax ($KEY/pod vs. $KEY/container.$CONTAINER_NAME) and 'convenience' ($KEY interpreted as a shorthand notation/alias for $KEY/pod).

pkg/adaptation/result.go
```diff
@@ -386,6 +390,36 @@ func (r *result) adjustDevices(devices []*LinuxDevice, plugin string) error {
...
+	//   is typically a strong ownership and a single related management entity
+	//   per vendor/class for these devices we do treat multiple injection of
+	//   the same CDI device reference as an error here.
+
+	id := r.request.create.Container.Id
+
+	// apply additions to collected adjustments
+	for _, d := range devices {
+		if err := r.owners.claimCDIDevice(id, d.Name, plugin); err != nil {
+			return err
```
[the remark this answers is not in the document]
klihub: Do you mean to verbatim include that comment here... as a reminder to consider that in the future ?

pkg/adaptation/adaptation_suite_test.go
klihub: Updated to use `"CDI-device"`.

pkg/api/adjustment.go
klihub: True and thanks for spotting that. Removed.

pkg/runtime-tools/generate/generate.go
klihub: Updated, using requested instead.

pkg/runtime-tools/generate/generate.go
klihub: Moved it accordingly.

plugins/device-injector/README.md
klihub: No. I was way too hasty when addressing your original comment in the docs. It is not unspecified at all which key we pick, since we don't iterate the annotation map, but iterate a slice of key according to our preference/precedence logic. Hence, we always use the more specific annotation (IOW, always only using a container-specific one if present), preferring the explicitly articulated `$key/pod` notation over the plain `$key` one. I updated the docs to reflect this.

plugins/device-injector/device-injector.go
klihub: Yup, noticed the same and I already updated the docs accordingly.

pkg/adaptation/api.go
klihub: Yeah, not sure what would be the best for this. Originally I tried to have this is an sort of 'logical hierarchical order', but it's probably not really that now. Lexical sorting would clearly be one well defined order. Anyway, I'd not reshuffle it now in this PR, as that would be an unrelated change.

elezar: Sure, let's revisit this in a follow-up.

pkg/runtime-tools/generate/generate.go
elezar:
```suggestion
// Devices are requested by their fully qualified CDI device names. The
```
(this could also be `specified`).

elezar (approved):
Thanks @klihub.

This looks good from my side. (there is a minor comment string suggestion, but not a blocker at all).

plugins/device-injector/README.md
```diff
@@ -1,16 +1,20 @@
...
+these latter two annotations will be ever taken into account. If both are
+present, `devices.nri.io/pod` is used and `devices.nri.io` is silently
+ignored.
```
mikebrow, line 15:
```suggestion
ignored, otherwise `devices.nri.io`, in the absence of additional suffix text,  is processed as shorthand for the `devices.nri.io/pod` annotation.
```

pkg/adaptation/result.go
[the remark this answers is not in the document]
mikebrow: sort of.. suggesting the consideration of a TODO here, not that explicit text, or we can just open an issue though the way issues get removed when the bots are turned on :-)

mikebrow (approved):
LGTM

PR #103, opened 2024-08-19
adaptation: close plugin if initial synchronization fails.

klihub:
Close the connection to the plugin if initial synchronization fails, like we do for all other failures considered unrecoverable.

### DIFF
--- a/pkg/adaptation/adaptation.go
+++ b/pkg/adaptation/adaptation.go
@@ -450,7 +450,7 @@ func (r *Adaptation) acceptPluginConnections(l net.Listener) error {
 				r.sortPlugins()
 				r.Unlock()
 
-				log.Infof(ctx, "plugin %q connected", p.name())
+				log.Infof(ctx, "plugin %q connected and synchronized", p.name())
 			}
 		}
 	}()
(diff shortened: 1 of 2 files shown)


pkg/adaptation/plugin.go
```diff
@@ -418,6 +418,7 @@ func (p *plugin) synchronize(ctx context.Context, pods []*PodSandbox, containers
 	}
 	rpl, err := p.stub.Synchronize(ctx, req)
 	if err != nil {
+		p.close()
```
mikebrow, line 421: nit.. rename this func to initialSynchronizeOnConnect
same on above log.Info.. "synchronizing plugin on connect %s"

pkg/adaptation/adaptation.go
```diff
@@ -449,9 +449,9 @@ func (r *Adaptation) acceptPluginConnections(l net.Listener) error {
 				r.plugins = append(r.plugins, p)
 				r.sortPlugins()
 				r.Unlock()
-			}
 
-			log.Infof(ctx, "plugin %q connected", p.name())
+				log.Infof(ctx, "plugin %q connected", p.name())
```
mikebrow, line 453:
```suggestion
				log.Infof(ctx, "plugin %q connected and synchronized", p.name())
```

pkg/adaptation/plugin.go
mikebrow: or will it also be used in catchup scenarios ..

mikebrow (approved):
LGTM

fuweid (approved):

PR #108, opened 2024-09-10
fix: Prevent calling the main process exit without WithOnClose()

attlee-wang:
My service is a multi-goroutine daemon service, and nri-plugin is one of the modules of my service. 
In nri-plugin I did not use WithOnClose() to set the exit callback function, I would rebuild a new stub when exiting. I found that when the stup exception occurred, the daemon service main was actually exited. I felt that it was very unfriendly to exit the deamon service due to a goroutine exception
After checking, it is found that there is a problem with the following code：
```
// Handle a lost connection.
func (stub *stub) connClosed() {
	stub.Lock()
	stub.close()
	stub.Unlock()
	if stub.onClose != nil {
		stub.onClose()
		return
	}

	os.Exit(0) // exits the process
}

```
os.Exit(0) exits the process，for the nri module package, using runtime.Goexit() is more elegant. runtime.Goexit() will only exit the current goroutine, not the daemon process.


attlee-wang:
@klihub @mikebrow @fuweid can anyone help to review this ? thanks.

klihub:
> My service is a multi-goroutine daemon service, and nri-plugin is one of the modules of my service. In nri-plugin I did not use WithOnClose() to set the exit callback function, I would rebuild a new stub when exiting. 

Not directly related to this PR, but it might be useful for you: ever since we have merged PR #91, it became possible to restart a stub by  using a `Stop()` followed by a `Start()`. 

> I found that when the stup exception occurred, the daemon service main was actually exited. I felt that it was very unfriendly to exit the deamon service due to a goroutine exception. After checking, it is found that there is a problem with the following code：
> ...

This behavior has been discussed and criticized earlier a few times, and it is fair to say that not without a reason. I have a few question/comments related to this.

1. This original, and admittedly a bit harshly/hastily implemented behavior, was partly in place to make pre-connected/runtime-launched plugins exit once they lose their pre-connected transport (a `socketpair()` which cannot be externally re-established). It would be nice and also fairly easily possible to retain this behavior. We could either save the fact in the original connection setup/adoption code that we took a socketpair and then use that in the connection lost handler, or we could simply re-check the same env. var and use its existence for the same.
2. Cannot you work around this immediate problem in your current code by simply creating your stub with a `WithOnClose()` handler to prevent the stub internally doing a `os.Exit(0)` ? Your handler could either do nothing or, for instance using a channel, signal your upper layer logic to recreate (or Stop()/Start()) the stub ?

pkg/stub/stub.go
```diff
@@ -543,7 +544,7 @@ func (stub *stub) connClosed() {
 		return
 	}
 
-	os.Exit(0)
+	runtime.Goexit() //The exit code can be determined when the coroutine exits
```
klihub, line 547: I'm not sure why a `runtime.Goexit()` would be better here than not doing anything (other than maybe recording the fact that we lost connection) and simply returning instead? If the stub has been started by `stub.Start()` or `stub.Run()` it is sitting in `ttrpc.Server.Serve()` which should return (soon) once we return from here. That should cause `stub.Run()` to return if the caller used it that to start the stub, or otherwise `stub.Wait()` to return if that is used to wait for the stub to stop.

I think one alternative we could do is to
- either record in `connect()` the fact that we've taken a `socketpair` from the env. (IOW. we are running as a pre-connected plugin launched by the runtime)
- and use that stored state to `os.Exit()` here, or check here for the presence of the same env. var and simply use that fact to do the same

klihub: Or I think something like this could work, too:

```diff
diff --git a/pkg/stub/stub.go b/pkg/stub/stub.go
index 88fd2e1..b68344f 100644
--- a/pkg/stub/stub.go
+++ b/pkg/stub/stub.go
@@ -502,6 +502,15 @@ func (stub *stub) connect() error {
                        return fmt.Errorf("invalid socket (%d) in environment: %w", fd, err)
                }
 
+               userOnClose := stub.onClose
+               stub.onClose = func() {
+                       if userOnClose != nil {
+                               userOnClose()
+                       }
+                       log.Infof(noCtx, "Connection from environment lost, exiting...")
+                       os.Exit(0)
+               }
+
                return nil
        } 
```

klihub:
ping @attlee-wang

attlee-wang:
> 2\. `WithOnClose()`您是否不能通过简单地使用处理程序创建存根来防止存根内部执行来解决当前代码中的这个直接问题？如果您既没有使用 运行它，也没有等待它终止，`os.Exit(0)`您的处理程序可以不执行任何操作，或者（例如）使用通道向您的上层逻辑发出信号以重新创建（或 Stop()/Start()）存根？`stub.Run()``stub.Wait()`

@klihub  It's not that I don't use WithOnClose(). I looked at the stub.go package and saw stub.Stop(). I subjectively thought that stub.Stop() could be used. After all, WithOnClose() is just an optional function.

attlee-wang:
> 2. Cannot you work around this immediate problem in your current code by simply creating your stub with a `WithOnClose()` handler to prevent the stub internally doing a `os.Exit(0)` ? Your handler could either do nothing or, for instance using a channel, signal your upper layer logic to recreate (or Stop()/Start()) the stub, if you are neither running it with `stub.Run()` nor waiting for its termination with `stub.Wait()` ?

@klihub It's not that I don't use WithOnClose(). I looked at the stub.go package and saw stub.Stop(). I subjectively thought that stub.Stop() could be used. After all, WithOnClose() is just an optional function.

PR #109, opened 2024-09-20
api: pass configured timeouts to plugins.

klihub:
Pass configured registration and request timeouts to plugins in the `Configure` request.


.github/workflows/ci.yml
```diff
@@ -85,7 +85,7 @@ jobs:
 
     strategy:
       matrix:
-        go: [1.20.x, 1.21.x, 1.22.3]
+        go: [1.21.x, 1.22.3]
```
mikebrow, line 88: the go.mod line 3 go minimum to be g v1.20 but that version is no longer tested.. was there a test issue

mikebrow: same question on the sub-tree package go mod's should they all move up to 1.21

klihub: Yes, I had this hitting us when running tests:

```
Error:
/home/runner/go/pkg/mod/github.com/onsi/gomega@v1.34.0/matchers/support/goraph/bipartitegraph/bipartitegraphmatching.go:4:2: package slices is not in GOROOT (/opt/hostedtoolcache/go/1.20.14/x64/src/slices)
```

It looked testing-related so I simply removed `1.20.x` from the test matrix to get rid of it. But you're right that both containerd (1.7) and cri-o are past 1.20. So maybe we could also bump the minimum requirement to 1.21. And if we bump NRI itself to 1.21 then I think we should do the same for the plugins as well.

zhaodiaoer:
> /cc @zhaodiaoer

LGTM

This is exactly what we need. The plugin owner can use these configurations more conveniently to adjust the context of the internal logic. Thanks.

mikebrow (approved):
LGTM

mikebrow (approved):
still LGTM

klihub:
@fuweid PTAL

dmcgowan (approved):

PR #111, opened 2024-09-26
fixes: use multiple sync messages if we hit ttrpc max message limit.

klihub:
This PR updates the protocol to allow initial plugin sync to use multiple messages. This is necessary if we hit the maximum ttrpc message size limit. That in turn can happen, as the maximum limit for total annotation key 
 and values in K8s is 256k. The adaptation implementation is updated to fall back to multiple messages if necessary. The stub implementation is updated to transparently collect and recombine multipart sync messages into a single one before passing it up the stack to the plugin.


pkg/adaptation/plugin.go
```diff
@@ -412,19 +414,101 @@ func (p *plugin) synchronize(ctx context.Context, pods []*PodSandbox, containers
...
+			log.Debugf(ctx, "oversized message, retrying in smaller chunks")
+		}
 	}
 
 	return rpl.Update, nil
 }
 
+func recalcObjsPerSyncMsg(pods, ctrs int, err error) (int, int, error) {
+	const (
+		minObjsPerMsg = 8
```
mikebrow, line 473: are these pod/container minimum counts actual based on size limit maximums for the fields or just not wanting to loop until we try 1obj?

klihub: It's roughly based on the observation that the maximum annotation size per object is 256k. If you put then 8 of those in a single message (ignoring all other message container which it then negligible), then you get raw data a bit over 2M. Encoded, that should fit easily in to the max message size limit of 4M, so there is not much point in trying to scale back below that.

pkg/adaptation/plugin.go
```diff
@@ -412,19 +414,101 @@ func (p *plugin) synchronize(ctx context.Context, pods []*PodSandbox, containers
...
+	maxLen := e.MaximumLength()
+	msgLen := e.RejectedLength()
+
+	if msgLen == 0 || maxLen == 0 || msgLen <= maxLen {
+		return pods, ctrs, fmt.Errorf("failed to synchronize plugin with split messages")
+	}
+
+	factor := float64(maxLen) / float64(msgLen)
+	if factor > 0.9 {
+		factor = 0.9
```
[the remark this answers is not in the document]
klihub: This is just capping the scale back factor to a maximum of 0.9. We unwrap from the error message both the maximum message size (which is at the moment an unexported constant in ttrpc) and the offending message size. Based on these two, we calculate a really naive/dumb linear scale back factor first, then we make sure it is not more than 0.9. So if our message size is roughly twice the allowed max, we'll (naively) cut the number of containers and pods in half and retry. This 0.9 capping only kicks in, if we are very close to the max, so that we rather voluntarily err on the side of a bit more aggressive scale back and fewer oversized retries than the other way around.

pkg/adaptation/plugin.go
mikebrow: cool, thx!

mikebrow (approved):
LGTM

fuweid (approved):

PR #112, opened 2024-09-27
Makefile: regenerate protobufs only on proto content change

askervin:
make sometimes triggers generating protobufs in github actions due to proto/protobuf timestamp changes. This causes errors in github ci as protobuf files are not ment to be build there and thus tools are missing.

This change changes protobuf generation to happen only if proto file contents have changed after previous generation.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -171,8 +171,14 @@ golangci-lint:
 #
 
 %.pb.go: %.proto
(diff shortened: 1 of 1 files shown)


askervin:
This is to be solved with a separate check in CI.

Issue #117, opened 2024-10-23
OOMScoreAdj is incorrectly cleared even if no plugin performs an adjustment


samuelkarp:
From https://github.com/kubernetes/kubernetes/issues/128251, and see the new failing unit test in https://github.com/containerd/nri/pull/116.

mikebrow:
@klihub FYI

samuelkarp:
Leaving this open to track the remainder of the work:

- [ ] new tag for `github.com/containerd/nri`
- [ ] update containerd to consume the new version

PR #116, opened 2024-10-23
 generate: do not set OOMScoreAdj if no adjustment

samuelkarp:
Existing OOMScoreAdj on the incoming spec should not be cleared.

Fixes https://github.com/containerd/nri/issues/117

PR #118, opened 2024-10-24
adaptation: return nil as no-op adjustment.

klihub:
With no NRI plugins registered or none of the registered plugins requesting changes to a container, we should ideally skip applying NRI adjustments on the runtime side altogether, instead of going through the motions without any effective changes. Having seen #117 (quickly fixed by #116), this is not the case.

This PR is the first part in an attempt to change that. With this PR in place, we don't create an a priori empty adjustment in preparation to process NRI requests. We only create it once a plugin makes an actual change to the container being created. Effectively, this should result in a nil adjustment reply when there are no plugins present, or none of the present plugins touches a container.

The other part is a related change to the [runtime integration code](https://github.com/klihub/containerd/commit/276cb1c6d433083e8bc1312cb2ec14f77cb6c657) to check for this and short-circuit processing.

### DIFF
--- a/pkg/adaptation/result.go
+++ b/pkg/adaptation/result.go
@@ -90,27 +90,7 @@ func collectCreateContainerResult(request *CreateContainerRequest) *result {
 		request: resultRequest{
 			create: request,
 		},
-		reply: resultReply{
-			adjust: &ContainerAdjustment{
-				Annotations: map[string]string{},
-				Mounts:      []*Mount{},
-				Env:         []*KeyValue{},
-				Hooks:       &Hooks{},
-				Rlimits:     []*POSIXRlimit{},
-				CDIDevices:  []*CDIDevice{},
-				Linux: &LinuxContainerAdjustment{
-					Devices: []*LinuxDevice{},
-					Resources: &LinuxResources{
-						Memory:         &LinuxMemory{},
-						Cpu:            &LinuxCPU{},
-						HugepageLimits: []*HugepageLimit{},
-						Unified:        map[string]string{},
-					},
-					Namespaces: []*LinuxNamespace{},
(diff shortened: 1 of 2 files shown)


pkg/adaptation/result.go
```diff
@@ -751,6 +757,73 @@ func (r *result) adjustRlimits(rlimits []*POSIXRlimit, plugin string) error {
...
+}
+
+func (r *result) initAdjustResources() {
+	r.initAdjustLinux()
+	if r.reply.adjust.Linux.Resources == nil {
+		r.reply.adjust.Linux.Resources = &LinuxResources{
+			Memory:         &LinuxMemory{},
+			Cpu:            &LinuxCPU{},
+			HugepageLimits: []*HugepageLimit{},
+			Unified:        map[string]string{},
```
champtar, line 806: We only need to init the fields that are pointers no ?

champtar: Actually we need to init the map at least, maybe not the array

champtar (approved):

pkg/adaptation/result.go
klihub: Yes, maps need to be initialized if you might try adding or updating map entries.

mikebrow (approved):
LGTM  just have to push and cherry the nil check referenced in the PR desc.

mikebrow:
> @mikebrow No, I think not yet. It'll require [something like this](https://github.com/klihub/containerd/commit/276cb1c6d433083e8bc1312cb2ec14f77cb6c657). But I think I know what you're after. It would be much safer to (first rebase, retest and make sure that it is really all we'll need and then) merge that change in containerd (and a similar one in CRI-O, even if they are now checks for something that can't happen), and only then merge this one. Right ?

nod.. thx!

PR #124, opened 2024-11-22
api: add namespace adjustment

tych0:
We are interested in running some parts of a pod in host or totally separate pid and network namespaces, so add an adjustment that allows for that.


pkg/adaptation/result.go
```diff
@@ -1349,6 +1373,14 @@ func (o *owners) claimOomScoreAdj(plugin string) error {
 	return nil
 }
 
+func (o *owners) claimNamespaces(plugin string) error {
+	if other := o.namespaces; other != "" {
+		return conflict(plugin, other, "seccomp policy")
```
dcantah, line 1378:
```suggestion
		return conflict(plugin, other, "namespaces")
```

pkg/runtime-tools/generate/generate_suite_test.go
```diff
@@ -375,6 +375,35 @@ var _ = Describe("Adjustment", func() {
...
+			rg := &rgen.Generator{Config: spec}
+			xg := xgen.SpecGenerator(rg)
+
+			Expect(xg).ToNot(BeNil())
+			Expect(xg.Adjust(adjust)).To(Succeed())
+			Expect(spec).To(Equal(makeSpec(
+				withNamespaces([]rspec.LinuxNamespace{
+					{
+						Type: rspec.PIDNamespace,
+						Path: "/meshuggah.rocks",
```
dcantah, line 401: 😆

dcantah (approved):
LGTM after the error msg fix

klihub (approved):
LGTM. @mikebrow WDYT ?

I suspect this PR (and the #123 seccomp one) might revive the discussion around the possibility to administratively lock down some NRI adjustments via configuration.

mikebrow:
Need more detail on the use case.  What is a "part" of a pod.  E.g. the network namespace, well there is only one network namespace for reported networks of the pod and all of it's containers.. and that network namespace is host, pod, or user namespace pod as directed by kubelet based on the pod spec.  Is this some sort of non-k8s use case for linux distros to support pods such as podman pods?  Need to understand these use cases to understand where and how to manage these security / isolation changes, possibly on a client basis and possibly under a new non k8s.io namespace.

tych0:
Sure,

> What is a "part" of a pod.

Specifically, one container in a pod. The rest of the pod we will leave as is. In fact,

> E.g. the network namespace, well there is currently only one network namespace for the shared networks of the pod and all of it's containers..

it's exactly the network namespace that we want to change here. The rest of the pod will live in the same set of namespaces as it usually does.

>  Is this some sort of non-k8s use case for linux distros

It is non-k8s in the sense that the network namespace we care about is created entirely outside of k8s, and there is no k8s infrastructure for managing it. It is unrelated to any linux-distro specific thing, and has to do with Netflix' network architecture. There is an old Plumber's talk about the specifics here: https://lpc.events/event/11/contributions/932/attachments/908/1764/LPC%202021_%20Talking%20IPv6%20to%20IPv4%20Without%20NAT_2.pdf

tych0:
> it's exactly the network namespace that we want to change here.

Actually it's the pid ns as well. We want to run in the parent pidns of the containers, so that we can see them to do seccomp() operations on them correctly.

mikebrow:
thx for the detail

champtar:
I didn't pay attention to the open PRs and ended up doing another PR to adjust namespaces: #135 
The key differences:
- you pass only the added / modified / deleted namespaces, not the full list
- ownership is per namespace (1 plugin add cgroup, another plugin change network)
- there are some helper functions

I'm fine if we pick this PR in the end, I just want to be able to adjust namespaces :)

As for the security discussion, I though NRI was considered part of the runtime, ie you get NRI you get root.
Today adjusting mounts or devices you can likely already escape to the host, so I don't think adjusting namespaces or seccomp changes anything security wise, it's already wide open.

pkg/runtime-tools/generate/generate.go
```diff
@@ -332,6 +335,16 @@ func (g *Generator) AdjustOomScoreAdj(score *nri.OptionalInt) {
 	}
 }
 
+func (g *Generator) AdjustNamespaces(namespaces []*nri.LinuxNamespace) error {
+	for _, ns := range namespaces {
+		if err := g.AddOrReplaceLinuxNamespace(ns.Type, ns.Path); err != nil {
+			return err
+		}
+	}
```
champtar, line 343: In the commit you say `We are interested in running some parts of a pod in host` but if I understand this correctly you can't remove a namespace, correct ?

klihub:
@samuelkarp @tych0 @dcantah @mikebrow @champtar @kad @etungsten I'd really like to try moving things forward both with these pending PRs and #137. Since I couldn't come up with anything better, I rolled a branch for testing with

- a prototype implementation of configurable adjustment restrictions (described in #137) 
- cherry-picked PRs #123 (seccomp policy adjustment),
- cherry-picked #124 (namespace adjustment)
- also #118 (use nil as no-op adjustment) pulled in as that requires a small change to #123 and #124
- commits to implement adjustment restrictions with tests for namespaces and seccomp policies 
- adjustment setters for namespaces and seccomp policies for consistency with existing stuff

If the approach proposed in #137 is anywhere close to something acceptable, then this branch should show what bits and pieces we'd need to get those controls in, together with (globally) restrictable namespace and seccomp adjustments (as a first step). If you have some extra cycles, PTAL. And if you have 
 any comments regarding #137, please chime in there as well.

champtar:
@klihub I add a very quick look at your branch, I would prefer you to use #135 as it's more fine grained (separate add/modify/delete, per namespace ownership, ...)
In any case can you open a PR so this move forward ?
Small note, I initially just read the email notification and I'm just seeing the link to the code now, better to post new messages than edit

klihub:
Closing in favor of #135 by @champtar, which implements a superset of this functionality.

PR #121, opened 2024-11-06
Add WASM plugin support

saschagrunert:
Just an inspiration how WASM support could work.

PR #125, opened 2024-12-12
Add dedicated plugin type struct

saschagrunert:
Follow-up on: https://github.com/containerd/nri/pull/121#discussion_r1872009094

PTAL @mikebrow

### DIFF
--- a/pkg/adaptation/plugin_type.go
+++ b/pkg/adaptation/plugin_type.go
@@ -1,0 +1,80 @@
+/*
+   Copyright The containerd Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


pkg/adaptation/plugin_type.go
```diff
@@ -0,0 +1,80 @@
...
+
+import (
+	"context"
+
+	"github.com/containerd/nri/pkg/api"
+)
+
+type pluginType struct {
+	wasmImpl api.Plugin
+	grpcImpl api.PluginService
```
klihub, line 27: nit: technically `ttrpcImpl` would be more correct.

pkg/adaptation/plugin_type.go
```diff
@@ -0,0 +1,80 @@
...
+}
+
+func (p *pluginType) isWasm() bool {
+	return p.wasmImpl != nil
+}
+
+func (p *pluginType) isGrpc() bool {
```
klihub, line 34: nit: ditto here, strictly speaking `isTtrpc()` would be the correct name.

pkg/adaptation/plugin_type.go
saschagrunert: Makes sense, fixed as suggested :+1:

klihub (approved):
LGTM

mikebrow (commented):
excellent

PR #134, opened 2025-01-21
pkg/api: Copy Unified map

champtar:
Both FromOCILinuxResources and FromCRILinuxResources were not copying the Unified map, so NRI plugins don't have access to Unified for both the pod or the container.


klihub (approved):
Thank you, champtar ! LGTM.
Tested this with the K8s MemoryQoS feature gate enabled.

fuweid (approved):

PR #135, opened 2025-01-23
Allow to adjust LinuxNamespaces

champtar:
2 use cases I have in mind:
- add cgroup namespace for k8s privileged pods
- enable time namespace (depends at least on https://github.com/opencontainers/runtime-tools/issues/783)


klihub:
@champtar AFAICT, this is an alternative implementation of #124.

champtar:
@klihub indeed ...
Now that I've written it, mine is more fine grained, ie you can add / modify / remove individual namespaces

champtar:
Ci is broken: `/bin/sh: 2: protoc: not found`

champtar:
@klihub as there is some interest in having namespaces adjustment, can you ping the right persons to move this PR or #124 forward ?

klihub:
> @klihub indeed ... Now that I've written it, mine is more fine grained, ie you can add / modify / remove individual namespaces

Currently there is a pair of PRs (#123, #124) with security implications, and it's actively being discussed what additional infra bits we'd need (mostly the ability to lock down selected adjustment capabilities in NRI administratively) so that everybody would feel confident enough about taking those changes in. Your PR is an alternative implementation of #124, so if you have valid uses cases for all of these additional capabilities, could you chime in #124 and point to this PR of yours as an alternative ? 

, so IMO the right thing to do is to

pkg/adaptation/result.go
```diff
@@ -404,6 +411,44 @@ func (r *result) adjustDevices(devices []*LinuxDevice, plugin string) error {
...
+			creatensmap[key] = n
+		}
+		r.reply.adjust.Linux.Namespaces = append(r.reply.adjust.Linux.Namespaces, n)
+	}
+
+	// maps.Values() requires go 1.23+
+	creatensarray := []*LinuxNamespace{}
+	for _, n := range creatensmap {
+		creatensarray = append(creatensarray, n)
+	}
```
djdongjin: we're going to have go 1.24+ soon

https://github.com/containerd/nri/pull/148/files#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6

plugins/default-validator/default-validator.go
```diff
@@ -121,6 +128,37 @@ func (v *DefaultValidator) validateOCIHooks(req *api.ValidateContainerAdjustment
...
+	}
+
+	offenders := ""
+	sep := ""
+
+	if len(owners) > 1 {
+		offenders = "plugins "
+	} else {
+		offenders = "plugin "
+	}
```
djdongjin: nit: maybe simply `offenders := "plugin(s)"` :)

djdongjin (approved):

pkg/adaptation/result.go
klihub: I think this can be updated once we bump go to 1.24+ in NRI main.

plugins/default-validator/default-validator.go
klihub: @djdongjin Updated PR like that.

mikebrow:
let's wait to rebase this on #123

klihub:
> pls rebase...

@mikebrow Rebased and pushed.

pkg/adaptation/result.go
champtar: NRI now requires go 1.24

pkg/api/adjustment.go
```diff
@@ -147,6 +147,20 @@ func (a *ContainerAdjustment) AddCDIDevice(d *CDIDevice) {
 	a.CDIDevices = append(a.CDIDevices, d) // TODO: should we dup d here ?
 }
 
+// AddOrReplaceNamespace records the addition or replacement of the given namespace to a container.
+func (a *ContainerAdjustment) AddOrReplaceNamespace(n *LinuxNamespace) {
+	a.initLinuxNamespaces()
+	a.Linux.Namespaces = append(a.Linux.Namespaces, n) // TODO: should we dup n here ?
```
mikebrow, line 153: What do you mean by this TODO?  oci only allows one namespace per type and the func we are calling AddOrReplaceLinuxNamespace ensures that..

mikebrow (commented):
just one question around the TODO..

pkg/api/adjustment.go
klihub: @mikebrow It's related to the similar comments about the corresponding helpers for mounts and [CDI] devices (as this is done consistently with those).

Basically we append here the user-supplied `n *LinuxNamespace` as such to the set of adjustments, as opposed to appending a copy of it. That (being a pointer) allows the user, in principle, to modify it once it has been recorded as an adjustment.

Since the plugin can only shoot itself in the foot with that, I think no too big harm is done. But I think it's good to have the comments in place, so that if we ever decide to change one, we update all the others consistently.

pkg/adaptation/result.go
klihub: Updated to use `slices.Collect(maps.Values(creatensmap))` instead.

mikebrow (approved):
LGTM

PR #136, opened 2025-01-27
Makefile: symlink-safer tinygo docker builds.

klihub:
Don't let symlinks on $(PWD) break dockerized tinygo builds of the sample WASM plugin so easily.


fuweid (approved):

mikebrow (approved):
LGTM

PR #139, opened 2025-02-04
adaptation: honor context deadline in WASM runtime/plugins.

### LINKED ISSUE #138: WASM plugins do not honor request timeouts / context.
Calling an event handler of  a WASM plugin with an inifinite loop never times out.

klihub:
Don't use the default wazero runtime configuration. Instead use one with `WithCloseOnContextDone(true) in order for calls to WASM plugins to properly obey context deadlines, which otherwise they don't. Also, once a WASM plugin is closed (for instance due to a request processing timeout), mark the plugin closed so it gets properly removed from the list of active plugins.

Fixes #138.

### DIFF
--- a/pkg/adaptation/adaptation.go
+++ b/pkg/adaptation/adaptation.go
@@ -30,6 +30,8 @@ import (
 	"github.com/containerd/nri/pkg/api"
 	"github.com/containerd/nri/pkg/log"
 	"github.com/containerd/ttrpc"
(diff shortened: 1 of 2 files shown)


saschagrunert (approved):
Thank you, great find!

fuweid (approved):

mikebrow (approved):
LGTM

PR #141, opened 2025-02-06
Add UpdatePodSandbox to NRI plugin API

chrishenzie:
This API is called by the container runtime to update pod sandbox resources when they change. This is for the [In-place Pod Resize KEP-1287](https://github.com/kubernetes/enhancements/tree/master/keps/sig-node/1287-in-place-update-pod-resources#cri-changes).

@tallclair
@samuelkarp


pkg/api/api.proto
```diff
@@ -234,7 +251,10 @@ enum Event {
   POST_UPDATE_CONTAINER = 9;
   STOP_CONTAINER = 10;
   REMOVE_CONTAINER = 11;
+  // TODO: Should this be moved to the end (assuming it's not used anywhere)?
```
klihub, line 254: This is used for event mask validation and String()ification, so it should be updated to really be the last one.

aojea:
cc @MikeZappa87 

This may have some interesting properties for networking plugins

askervin:
Now that pods can be updated, should we extend also `SynchronizeResponse` with `repeated PodSandboxUpdate`, too?

pkg/stub/stub.go
```diff
@@ -739,6 +739,11 @@ func (stub *stub) StopContainer(ctx context.Context, req *api.StopContainerReque
 	}, err
 }
 
+// UpdatePodSandbox request handler.
+func (stub *stub) UpdatePodSandbox(_ context.Context, _ *api.UpdatePodSandboxRequest) (*api.UpdatePodSandboxResponse, error) {
+	return nil, errors.New("unimplemented")
```
klihub, line 744: Do you plan to kick this forward later with further PRs or still with update to this PR ?

I'm asking because I was wondering whether we should try to wire up with the rest of the infra a bit more of the stub side handling for new requests/events already at this point ?

In practice we could add the corresponding new interfaces plugins could then choose to implement for receiving the new events:

```golang
// UpdatePodInterface handles UpdatePodSandbox API requests.
type UpdatePodInterface interface {
    // UpdatePodSandbox relays an UpdatePodSandbox request to the plugin.
    // TODO describe the changes the plugin request in response to this.
    UpdatePodSandbox(context.Context, *api.PodSandbox, *api.LinuxResources) (*api.UpdatePodSandboxUpdate, error)
 }

// PostUpdatePodInterface handles PostUpdatePodSandbox API events.
type PostUpdatePodInterface interface {
    // PostUpdatePodSandbox relays a PostUpdatePodSandbox event to the plugin.
    PostUpdatePodSandbox(context.Context, *api.PodSandbox) error
 }
```

Then we could add new/extend existing stub handlers for these new events:

```golang
// UpdatePodSandbox request handler.
func (stub *stub) UpdatePodSandbox(ctx context.Context, req *api.UpdatePodSandboxRequest) (*api.UpdatePodSandboxResponse, error) {
	handler := stub.handlers.UpdatePodSandbox
	if handler == nil {
		return nil, nil
	}
	update, err := handler(ctx, req.Pod, req.Resources)
	return &api.UpdatePodSandboxResponse{
		Update: update,
	}, err
}

...

// StateChange event handler.
func (stub *stub) StateChange(ctx context.Context, evt *api.StateChangeEvent) (*api.Empty, error) {
	var err error
	switch evt.Event {
...
	case api.Event_POST_UPDATE_POD_SANDBOX:
		if handler := stub.handlers.PostUpdateRunPodSandbox; handler != nil {
			err = handler(ctx, evt.Pod)
		}
...
        }

	return &api.StateChangeResponse{}, err
}
```

And update the handler setup code to subscribe for the new events if there are plugin side handlers for them:

```golang
// Set up event handlers and the subscription mask for the plugin.
func (stub *stub) setupHandlers() error {
...
	if plugin, ok := stub.plugin.(RunPodInterface); ok {
		stub.handlers.RunPodSandbox = plugin.RunPodSandbox
		stub.events.Set(api.Event_RUN_POD_SANDBOX)
	}
	if plugin, ok := stub.plugin.(UpdatePodInterface); ok {
		stub.handlers.UpdatePodSandbox = plugin.UpdatePodSandbox
		stub.events.Set(api.Event_UPDATE_POD_SANDBOX)
	}
	if plugin, ok := stub.plugin.(PostUpdatePodInterface); ok {
		stub.handlers.PostUpdatePodSandbox = plugin.PostUpdatePodSandbox
		stub.events.Set(api.Event_POST_UPDATE_POD_SANDBOX)
	}
...
}
```

I probably forgot something, but I think taking care of these would already be pretty close to what we'd need for a working implementation on the stub/plugin side.

chrishenzie: I can make further updates to this PR, I just wanted to get some initial feedback on the API first to make sure there was general consensus

chrishenzie: Updated to include all of these changes + more (to make the adaptation test suite happy). I still have some implementations to fill in but wanted to get everything building again first.

pkg/stub/stub.go
```diff
@@ -826,6 +862,10 @@ func (stub *stub) setupHandlers() error {
 		stub.handlers.RemovePodSandbox = plugin.RemovePodSandbox
 		stub.events.Set(api.Event_REMOVE_POD_SANDBOX)
 	}
+	if plugin, ok := stub.plugin.(PostUpdatePodInterface); ok {
```
klihub, line 865: uber-nit: is there a particular reason not to put this right after UpdatePodInterface check, like you did in the other similar contexts, which does feel more intuitive as it follows the natural pod lifecycle ?

klihub:
> Now that pods can be updated, should we extend `SynchronizeResponse` with `repeated PodSandboxUpdate`, too?

Good question/point. And if we do that, we should also consider adding an (unsolicited) UpdatePodSandboxes next to UpdateContainers in the protocol.

pkg/api/api.proto
```diff
@@ -206,6 +211,18 @@ message StopContainerResponse {
   repeated ContainerUpdate update = 1;
 }
 
+message UpdatePodSandboxRequest {
+  // Pod being updated.
+  PodSandbox pod = 1;
+  // Resources to update.
+  LinuxResources linux_resources = 2;
```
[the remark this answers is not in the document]
chrishenzie: Updated to add an `overhead_linux_resources` field. If you have any preferred field names please let me know.

pkg/stub/stub.go
chrishenzie: I agree placing after UpdatePodSandbox is more intuitive. In this case I tried to follow the ordering used in this method, e.g. [the "post" methods for containers are placed at the end](https://github.com/containerd/nri/blob/v0.9.0/pkg/stub/stub.go#L849-L860), however I'm open to changing this to be more intuitive.

chrishenzie:
> > Now that pods can be updated, should we extend `SynchronizeResponse` with `repeated PodSandboxUpdate`, too?
> 
> Good question/point. And if we do that, we should also consider adding an (unsolicited) UpdatePodSandboxes next to UpdateContainers in the protocol.

Can the container runtime make changes to the Pod sandbox's resources? I thought that was managed (at least for now) by kubelet.

MikeZappa87:
> > > Now that pods can be updated, should we extend `SynchronizeResponse` with `repeated PodSandboxUpdate`, too?
> > 
> > 
> > Good question/point. And if we do that, we should also consider adding an (unsolicited) UpdatePodSandboxes next to UpdateContainers in the protocol.
> 
> Can the container runtime make changes to the Pod sandbox's resources? I thought that was managed (at least for now) by kubelet.

I am pretty certain if the PodSandboxStatus RPC returns additional labels, annotations that the kubelet doesn't know about a test will fail. I believe I broke that last year :-)

pkg/stub/stub.go
klihub: Ah, then it's fine, makes more sense the way you did it. I thought we were (more globally) consistent but we're apparently not.

klihub:
> > > Now that pods can be updated, should we extend `SynchronizeResponse` with `repeated PodSandboxUpdate`, too?
> > 
> > 
> > Good question/point. And if we do that, we should also consider adding an (unsolicited) UpdatePodSandboxes next to UpdateContainers in the protocol.
> 
> Can the container runtime make changes to the Pod sandbox's resources? I thought that was managed (at least for now) by kubelet.

Not at the moment... And actually considering that pod-level cgroup management is in the kubelet and not in the runtime, I think we try to do this now wrong here. At the NRI level this needs to be an event-type message (no possibility for the plugin to request changes to the pod in response).

And if this is true, it would be better to add it as a new Event enumeration and pass to the stub/plugin on the wire as a StateChange event, just like all the other similar ones.

pkg/adaptation/adaptation.go
```diff
@@ -208,6 +209,17 @@ func (r *Adaptation) RunPodSandbox(ctx context.Context, evt *StateChangeEvent) e
 	return r.StateChange(ctx, evt)
 }
 
+// UpdatePodSandbox relays the corresponding CRI request to plugins.
+func (r *Adaptation) UpdatePodSandbox(_ context.Context, _ *UpdatePodSandboxRequest) (*UpdatePodSandboxResponse, error) {
```
klihub, line 213: After thinking more about this, I think this should be an event: unlike for containers, pod cgroup management is in the kubelet. Therefore there is not much point in allowing a plugin to respond with requested changes to the update. There is nothing the runtime could do to make that happen.

pkg/adaptation/adaptation.go
```diff
@@ -50,6 +50,7 @@ type SyncFn func(context.Context, SyncCB) error
 type SyncCB func(context.Context, []*PodSandbox, []*Container) ([]*ContainerUpdate, error)
 
 // UpdateFn is a container runtime function for unsolicited container updates.
+// TODO: I think I'll need to add one for PodSandboxUpdates.
```
klihub, line 53: I think UpdatePodSandbox needs to be a purely informative event, not allowing a plugin to request changes to the update in response. Consequently there should not be unsolicited updates requested by a plugin either. Then there is no need for such a new opaque update function either.

pkg/api/api.proto
```diff
@@ -206,6 +211,20 @@ message StopContainerResponse {
...
+  PodSandbox pod = 1;
+  // Overhead associated with this pod.
+  LinuxResources overhead_linux_resources = 2;
+  // Sum of container resources for this pod.
+  LinuxResources linux_resources = 3;
+}
+
+message UpdatePodSandboxResponse {
+  // Requested update to pod.
+  PodSandboxUpdate update = 1;
```
klihub, line 225: Considering that pod level cgroup handling is in the kubelet I think this is a mere informative event and the message should reflect it. There is nothing the runtime could do if a plugin decided to request changes to the pod resource update. Then the response message should reflect this, and be simply empty.

pkg/stub/stub.go
```diff
@@ -69,6 +69,14 @@ type RunPodInterface interface {
 	RunPodSandbox(context.Context, *api.PodSandbox) error
 }
 
+// UpdatePodInterface handles UpdatePodSandbox API requests.
+type UpdatePodInterface interface {
+	// UpdatePodSandbox relays an UpdatePodSandbox request to the plugin.
+	// TODO: Add comment on anything the plugin can do in response (if
+	// applicable).
+	UpdatePodSandbox(context.Context, *api.PodSandbox, *api.LinuxResources, *api.LinuxResources) (*api.PodSandboxUpdate, error)
```
klihub, line 77: Considering the previous notes about pod-level cgroup management being done in the kubelet and not the runtime, I think the return part signature should be just an error here: `UpdatePodSandbox(context.Context, *api.PodSandbox, *api.LinuxResources, *api.LinuxResources) error`.

MikeZappa87:
We were hoping to get the Pod Labels/Annotations added to the request message. The reason behind this is that if the label is updated on a K8s pod, the container runtime is never updated with this state thus etcd becomes the source of truth. 

https://github.com/kubernetes/cri-api/blob/82e1e9174de30efdb4f82fbb5251c31e4a7ce34a/pkg/apis/runtime/v1/api.proto#L479

https://github.com/kubernetes/cri-api/blob/82e1e9174de30efdb4f82fbb5251c31e4a7ce34a/pkg/apis/runtime/v1/api.proto#L496

aojea:
> We were hoping to get the Pod Labels/Annotations added to the request message. The reason behind this is that if the label is updated on a K8s pod, the container runtime is never updated with this state thus etcd becomes the source of truth. I think adding this will help the NRI ecosystem move service mesh and possibly other network plugins over to NRI.

If we have labels ( and maybe also annotations) information on the PodSandboxXXXXX hooks, then we remove all the need for watching the apiserver on network integrations , like cni plugins and service meshes, it will be a really nice addition that solves a long standing problem, the runtime is the source of truth for the Pod lifecycle on the Node in this hooks, components having to go to the apiserver are always going to race, since the state of the apiserver is the desired state that does not necessarily matches the current state on the runtime (this also solve problems when node get disconnected temporary, or have issues to connect to the control plane)

klihub:
@chrishenzie @askervin Looking at my comments from yesterday it's probably warranted to try and give a bit more coherent summary which does not look like a diary entry from an H.P. Lovecraft story:

- my understanding is that UpdatePodSandbox simply informs NRI plugins about what will happen
- therefore the plugin can't request changes to the reported update and the protocol/API should not imply otherwise
- consequently the response message should not include resources
- let's go with the dedicated new RPC call suggested here (as opposed to using StateChange on the wire)

klihub:
> We were hoping to get the Pod Labels/Annotations added to the request message. The reason behind this is that if the label is updated on a K8s pod, the container runtime is never updated with this state thus etcd becomes the source of truth. I think adding this will help the NRI ecosystem move service mesh and possibly other network plugins over to NRI.
> 
> Links to the fields.
> 
> https://github.com/kubernetes/cri-api/blob/82e1e9174de30efdb4f82fbb5251c31e4a7ce34a/pkg/apis/runtime/v1/api.proto#L479
> 
> https://github.com/kubernetes/cri-api/blob/82e1e9174de30efdb4f82fbb5251c31e4a7ce34a/pkg/apis/runtime/v1/api.proto#L496
> 
> cc: @mikebrow @aojea

@MikeZappa87 @aojea @askervin 

Looking at the CRI protocol definition changes [in the KEP](https://github.com/kubernetes/enhancements/tree/master/keps/sig-node/1287-in-place-update-pod-resources#cri-changes), it looks to me as if those `"annotations"` in `UpdatePodSandboxResourcesRequest` were blindly copied from `UpdateContainerResourcesRequest` simply replacing `container` with `sandbox` in the comments. Currently they are defined/described as

>    // Unstructured key-value map holding arbitrary additional information for
>   // sandbox resources updating. This can be used for specifying experimental
>   // resources to update or other options to use when updating the sandbox.
>    map<string, string> annotations = 4;

This seems to imply to me different semantics than updating the the pods annotations. Rather optionally annotating (how to apply) the updates instead.

Looking at the runtime implementations for `UpdateContainerResources`, I think the corresponding container resource update annotations are simply ignored and seem more like a protocol relic than something actually in use.

If we'd prefer not to carry over (and maybe only semi-accidentally) a relic to the new `UpdatePodSandboxResources` CRI RPC call, but instead would like to use it for actually updating pod annotations, it would be good to bring this up in the KEP. And probably also mention the desired addition of labels as well.

pkg/api/api.proto
chrishenzie: Would you prefer this uses `Empty{}`, or an empty `UpdatePodSandboxResponse`? I figure the latter is more extensible in case pod-level cgroup handling moves more to the container runtime in the future.

pkg/adaptation/adaptation.go
chrishenzie: Is there a likelihood that pod-level cgroup management could shift to the container runtime in the future? If so I think it would be more intuitive for this to not be just an event but to also have its own response type (which can be empty for now per your other feedback).

It looks like this method signature would fit in with [the approach in your branch](https://github.com/klihub/nri/commit/dfd20ac81115ba329c79e0e7d262d8a51448c087) so I'd be inclined to not make any changes here.

pkg/api/api.proto
```diff
@@ -445,6 +466,18 @@ message ContainerEviction {
...
+// Requested update to an already created pod sandbox.
+message PodSandboxUpdate {
+  string pod_sandbox_id = 1;
+  LinuxPodSandboxUpdate linux = 2;
+  bool ignore_failure = 3;
+}
+
+// Updates to (linux) resources.
+message LinuxPodSandboxUpdate {
+  LinuxResources resources = 1;
```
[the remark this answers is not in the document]
chrishenzie: Yes definitely, thanks for catching

pkg/adaptation/adaptation.go
chrishenzie: Acknowledged, removed this

pkg/stub/stub.go
chrishenzie: Fixed this, thanks

pkg/adaptation/adaptation.go
```diff
@@ -208,6 +208,25 @@ func (r *Adaptation) RunPodSandbox(ctx context.Context, evt *StateChangeEvent) e
 	return r.StateChange(ctx, evt)
 }
 
+// UpdatePodSandbox relays the corresponding CRI request to plugins.
+func (r *Adaptation) UpdatePodSandbox(ctx context.Context, req *UpdatePodSandboxRequest) (*UpdatePodSandboxResponse, error) {
```
chrishenzie, line 212: @klihub Does this implementation match your expectations (e.g. still treat as an event but with request / response types included)?

pkg/api/api.proto
klihub: Yes and that's why I'd prefer to use a dedicated `UpdatePodSandboxResponse`, if other reviewer will agree with that. It might be a bit inconsistent compared to the rest of the pod handling RPC calls and messages for the time being. But I'd like to anyway get rid of `Empty` (and the usage of `StateChange` on the wire) altogether before we tag an NRI 1.0.0 version. It's just better to hold that change back until all those pending PRs are merged which also touch the wire protocol, to avoid unnecessary conflicts.

klihub:
> @klihub agree, annotations are good for experimentation but then become an ugly API and extension point creating tightly coupling between the implementations ... we need to think about something better, I will check with the DRA folks, since this is about resources we may model this problem as a DRA problem that has better semantics to support what we want

Well, I meant something much simpler with my comment (and I think being able to propagate pod annotation/label updates to the runtime could be useful). All I wanted to say is that if we want to use the annotations in that message to update pod annotations, then we should bring this up in the KEP discussion. And also change the related comment in the protocol description, because it looks like something which was simply copied from the existing `UpdateContainerResourcesRequest` and the comments imply different usage/semantics.

pkg/adaptation/adaptation.go
klihub: This looks now a bit mixed. It is my fault not yours, as my comments so far have been bit of an unclear and unorganized mess. I apologize for that.

So about this version. This defines a new RPC call (UpdatePodSandbox) and protocol messages for it (UpdatePodSandbox{Request,Response}) on the wire in pkg/api/api.proto. But then on the runtime side (`pkg/adapatation') UpdatePodSandbox is implemented using the existing `StateChange` call. On the plugin/stub side however, it expects the newly defined RPC call, instead of an Event_UPDATE_POD_SANDBOX StateChange event. BTW, this is the reason why the commented out test fails for UpdatePodSandbox (but the one for PostUpdatePodSandbox would work).

It would be better/clearer to use the newly added RPC call instead of StateChange. That would require
- updating Adaptation.UpdatePodSandbox to not use StateChange (similar to UpdateContainer implementation, but there is no need to collect any result)
- implementing something like plugin.updatePodSandbox(context.Context, *UpdatePodSandboxRequest) (*UpdatePodSandboxResponse, error)
- and implementing pluginType.UpdatePodSandbox (called by the previous one) which simply does the selective call to either a wasm or a ttrpc plugin implementation
With these you should be able to enable the minimal tests you added and they should pass.

There is one more thing that bothers me a bit and I'd like @askervin and @kad to chime in regarding that. With the current KEP proposal, UpdatePodSandbox inevitably will end up being a post-UpdatePodSandbox, at least semantically. That renders our explicit PostUpdatePodSandbox technically useless now. If an error returned from a plugin for UpdatePodSandbox could prevent the update then in the case of multiple active plugins, PostUpdatePodSandbox would carry meaningful information. But that is not really the case. @askervin, @kad Any thoughts on this ?

askervin: Indeed the [KEP#CRI Changes](https://github.com/kubernetes/enhancements/tree/master/keps/sig-node/1287-in-place-update-pod-resources#cri-changes) states about the order:

> The Kubelet will call UpdatePodSandboxResources after it has reconfigured the pod-level cgroups. This ordering is consistent with pod creation, where the Kubelet configures the pod-level cgroups before calling RunPodSandbox.

The consistency does not seem to extend beyond the order: Kubelet cancels pod/container creation if RunPodSandbox returns an error from CRI, but:

> For now, the UpdatePodSandboxResources call will be treated as best-effort by the Kubelet. This means that in the case of an error, Kubelet will log the the error but otherwise ignore it and proceed with the resize.

I would like to interpret _For now_ so that at some point an error detected in a container runtime on resizing the pod may very well have a real effect. Stopping the pod, restoring pod resources, or retrying to resize to "this is the best the lower layers can do" values reported by the runtime, reporting resize status `Infeasible`... whatever the policy of handling the resize error would be in the future, it seems to be "log the error" for now. And therefore, at minimum, it would be great if NRI plugins could return an error to be logged.

Regarding splitting CRI UpdatePodSandboxResources into UpdatePodSandbox (that is already effectively a post-event as in the first quote above) and separate PostUpdatePodSandbox in NRI. I don't see use for PostUpdatePodSandbox, consistently to not having a PostRunPodSandbox event either. All updates are expected to be in place already at UpdatePodSandbox. This is not the case in similar-looking UpdateContainer because many NRI plugins can adjust the container, and PostUpdateContainer is needed to get all plugins on the same page with the final state of the updated container.

tallclair:
Following along the discussion of annotations, I think we should probably just drop the annotations field from the `UpdatePodSandboxResources` request for now. We are not currently using them for the analogous container request, and we don't support updating extended resources today. I would prefer to hold off on adding these until we have a specific use case that calls for them.

> If we'd prefer not to carry over (maybe even only semi-accidentally) a relic to the new UpdatePodSandboxResources CRI RPC call, but instead would like to use it for actually updating pod annotations, it would be good to bring this up in the KEP. And probably also mention the desired addition of labels as well.

I don't think tacking these onto the `UpdatePodSandboxResources` request is the right approach. If we want to pass updates to these annotations through the CRI, we should consider a dedicated request, or a generic `UpdatePodSandbox` request. For now, I'd like to keep the use case of `UpdatePodSandoxResources` more targeted.

felipeagger (approved):

pkg/adaptation/adaptation.go
klihub: Yes, I agree. I think we should stick to the normal RPC  signatures on the wire and just have the returned UpdatePodSandboxResponse struct be empty. In the stub/plugin interface (the handlers) though, it would probably be better to only have an returned error in the signature of the handler and add the the empty returned &UpdatePodSandboxResponse{} in the stub. WDYT ?

klihub: > Is there a likelihood that pod-level cgroup management could shift to the container runtime in the future? If so I think it would be more intuitive for this to not be just an event but to also have its own response type (which can be empty for now per your other feedback).

I know some folks who'd like to at least give some serious consideration for that to happen in the long run.

klihub:
@chrishenzie I think we'd still need to get rid of the usage of StateChange on the adaptation side implementation of UpdatePodSandbox. Maybe, something like this:  https://github.com/klihub/nri/commit/7b1b3e021e0871475f9bdf591570e8a834660ac9

pkg/adaptation/adaptation.go
mikebrow: can see this going both ways with a switch readOnly/readWriteOverlayRequests

mikebrow:
> We were hoping to get the Pod Labels/Annotations added to the request message. The reason behind this is that if the label is updated on a K8s pod, the container runtime is never updated with this state thus etcd becomes the source of truth. I think adding this will help the NRI ecosystem move service mesh and possibly other network plugins over to NRI.
> 
> Links to the fields.
> 
> https://github.com/kubernetes/cri-api/blob/82e1e9174de30efdb4f82fbb5251c31e4a7ce34a/pkg/apis/runtime/v1/api.proto#L479
> 
> https://github.com/kubernetes/cri-api/blob/82e1e9174de30efdb4f82fbb5251c31e4a7ce34a/pkg/apis/runtime/v1/api.proto#L496
> 
> cc: @mikebrow @aojea

Annotations is a good pattern to explore the thought that we should have a set of client owner annotations reported through the cri apis to the CRI client and another set of annotations that overlay (add) to the set managed under the client at the device layer. We could report the additional annotations to the client in a new field or in the info map.

mikebrow:
> Following along the discussion of annotations, I think we should probably just drop the annotations field from the `UpdatePodSandboxResources` request for now. We are not currently using them for the analogous container request, and we don't support updating extended resources today. I would prefer to hold off on adding these until we have a specific use case that calls for them.
> 
> > If we'd prefer not to carry over (maybe even only semi-accidentally) a relic to the new UpdatePodSandboxResources CRI RPC call, but instead would like to use it for actually updating pod annotations, it would be good to bring this up in the KEP. And probably also mention the desired addition of labels as well.
> 
> I don't think tacking these onto the `UpdatePodSandboxResources` request is the right approach. If we want to pass updates to these annotations through the CRI, we should consider a dedicated request, or a generic `UpdatePodSandbox` request. For now, I'd like to keep the use case of `UpdatePodSandoxResources` more targeted.

Nod... IMO any need to keep "extra" annotations, and/or ENVs, between the plugins and devices should kept in the container runtimes pod/container store as extras and either not reported to the client or if reported then as additional fields and/or placed temporarily in info maps..

MikeZappa87:
> > Following along the discussion of annotations, I think we should probably just drop the annotations field from the `UpdatePodSandboxResources` request for now. We are not currently using them for the analogous container request, and we don't support updating extended resources today. I would prefer to hold off on adding these until we have a specific use case that calls for them.
> > > If we'd prefer not to carry over (maybe even only semi-accidentally) a relic to the new UpdatePodSandboxResources CRI RPC call, but instead would like to use it for actually updating pod annotations, it would be good to bring this up in the KEP. And probably also mention the desired addition of labels as well.
> > 
> > 
> > I don't think tacking these onto the `UpdatePodSandboxResources` request is the right approach. If we want to pass updates to these annotations through the CRI, we should consider a dedicated request, or a generic `UpdatePodSandbox` request. For now, I'd like to keep the use case of `UpdatePodSandoxResources` more targeted.
> 
> Nod... IMO any need to keep "extra" annotations, and/or ENVs, between the plugins and devices should kept in the container runtimes pod/container store as extras and either not reported to the client or if reported then as additional fields and/or placed temporarily in info maps.. Trying to avoid said overrides/additions being unmanaged through oci hooks.

We should keep in mind that K8s and the container runtime may have different states for the labels as you can apply a label to a running pod and the container runtime is never updated.

klihub (approved):
LGTM.

fuweid (approved):

PR #145, opened 2025-02-28
api,adaptation: allow setting container command line.

klihub:
Allow plugins to set or update container command line arguments.


pkg/api/adjustment.go
```diff
@@ -80,6 +82,17 @@ func (a *ContainerAdjustment) RemoveEnv(key string) {
...
+func (a *ContainerAdjustment) SetArgs(args []string) {
+	a.Args = slices.Clone(args)
+}
+
+// UpdateArgs overrides the container command with the given arguments.
+// It won't fail if another plugin has already set the command line.
+func (a *ContainerAdjustment) UpdateArgs(args []string) {
+	a.Args = append([]string{""}, args...)
+}
+
```
etungsten, line 95: I'm somewhat having a hard time understanding the difference between `Set` and `Update` here. What does prepending an empty string do in the case of `UpdateArgs`?

etungsten (approved):
Thanks for the quick turnaround on this!

Tagging @jadams41 to take a look as well.

pkg/api/adjustment.go
mikebrow: The way I read it... In adjust args he's using a pre-pended null string in args index 0.. to clear the adjustment owning id and cuts that nil out of the slice this vs just failing if two plugins change the command..  IOW use updateargs if you don't want to fail when you are second and use setargs if you do..

mikebrow (approved):
LGTM

pkg/api/adjustment.go
klihub: Yes, it is exactly as @mikebrow says.

fuweid (approved):

Issue #140, opened 2025-02-04
Support standard golang compiler for WASM plugins


saschagrunert:
Tracking a feature for supporting the standard golang compiler (big go) for WASM plugins.

Refers to https://github.com/knqyf263/go-plugin/issues/58#issuecomment-2558130178

saschagrunert:
[This change is required](https://go-review.googlesource.com/c/go/+/604316) which will land in go 1.24.

saschagrunert:
Looks like https://github.com/knqyf263/go-plugin/pull/71 does the trick.

knqyf263:
I merged https://github.com/knqyf263/go-plugin/pull/71. It would be appreciated if you can test it. Once we confirm it works, I'll cut a new version of go-plugin.

saschagrunert:
I tested it with CRI-O and https://github.com/containerd/nri/pull/148 and it does not seem to be able to load the plugin:

```
INFO[2025-03-11T14:24:52.57275786+01:00] Found WASM plugin: /opt/nri/plugins/10-wasm   file="adaptation/plugin.go:106"
WARN[2025-03-11T14:24:53.989058924+01:00] failed to start pre-installed NRI plugin "wasm": failed to configure plugin: wasm error: unreachable
wasm stack trace:
        .runtime.abort(i32) i32
        .runtime.fatalpanic(i32) i32
        .runtime.gopanic(i32) i32
        .runtime.sigpanic(i32) i32
        .github.com_containerd_nri_pkg_api._plugin_configure(i32) i32
        .plugin_configure(i32,i32) i64  file="adaptation/adaptation.go:396"
```

knqyf263:
What command did you run? I'm not familiar with NRI. I have no idea how I can test it.

PR #148, opened 2025-03-11
Drop tinygo requirement

saschagrunert:
Fixes https://github.com/containerd/nri/issues/140


knqyf263:
`main` is not called anymore.
https://github.com/containerd/nri/blob/82c65a7af0df946a547cd934370e950bb289d9ee/plugins/wasm/plugin.go#L27-L29

You need to use `init()` instead.
https://github.com/knqyf263/go-plugin/blob/7819d6147c952d109df10990fe555029aff6c3bc/examples/host-function-library/plugin/plugin.go#L13-L18

Note that `main()` is still needed for Go to compile Wasm modules.

saschagrunert:
> `main` is not called anymore with a WASI reactor.
> 
> https://github.com/containerd/nri/blob/82c65a7af0df946a547cd934370e950bb289d9ee/plugins/wasm/plugin.go#L27-L29
> 
> You need to use `init()` instead. https://github.com/knqyf263/go-plugin/blob/7819d6147c952d109df10990fe555029aff6c3bc/examples/host-function-library/plugin/plugin.go#L13-L18
> 
> Note that `main()` is still needed for Go to compile Wasm modules.

Yes, thank you now it works when vendoring the NRI into CRI-O. :+1:

saschagrunert:
@knqyf263 CI is green now, do we want to cut a new go-plugin release first?

go.mod
```diff
@@ -1,10 +1,10 @@
 module github.com/containerd/nri
 
-go 1.22.0
+go 1.24.1
```
klihub, line 3: We need to be a bit careful with this and check if this is (already) fine (at this point) for both for `containerd/main` and `containerd/release/2.0`. Ideally also for 1.7.x, but I'd say updating that anyway might require rolling a maintenance branch with hand-picked cherry-picks and we probably don't need/want to push WASM support for 1.7.x anyway.

knqyf263:
@saschagrunert Thanks for testing it out. I'll cut a new release.

go.mod
aojea: but this is forcing everyone importing this library to move to 1.24, no?

saschagrunert: > but this is forcing everyone importing this library to move to 1.24, no?

Yes.

We could think about wrapping a `// +build go1.24` tag, but this would make a feature dependent on a certain golang version. It would also require more boilerplate.

mikebrow: nod, and the MUST be go 1.24.1 in the go.mod file is forcing everyone that includes this to be dependent on 1.24.1 or higher. This change by golang to make go the go dependency a MUST is crashing all backwards compatibility testing across the board. Typically you chase it down and someone thought it cool to use a new log feature just introduced.

aojea: I see, I was more asking about if was a deliberate change or an oversight, it looks is the only way, I didn;t mean to push in one or other direction

klihub:
@saschagrunert Is this something that would be important to get behind an NRI tag, so we could pull it in on the CRI-O side ?

If it is, then maybe one alternative would be to branch off a 0.9 release branch and merge this there, but keep if off main until all parties involved/affected are content with moving to go 1.24 as the minimum version.

go.mod
samuelkarp: Per the community call today, we are not going to upgrade containerd 2.1 to Go 1.24.  This leaves us with either option (2) or (3) above.

My preference as-is would be option 3, disabling WASM support entirely until we have moved to Go 1.24 (which will happen for containerd 2.2).  Unless there is some way to have the new WASM support sans-tinygo support old tinygo-based plugins?

saschagrunert:
> If it is, then maybe one alternative would be to branch off a 0.9 release branch and merge this there, but keep if off main until all parties involved/affected are content with moving to go 1.24 as the minimum version.

I think that's a valuable solution, yes. We will move to go 1.24 in the future in any case.

A build tag will not work with the go module requirements if I think about it, because go modules are not tag aware. Means we need the go 1.24 requirement in any case.

saschagrunert:
@klihub do you have a new base branch where I could rebase on?

klihub:
> @klihub do you have a new base branch where I could rebase on?

 No, sorry, not yet.

samuelkarp:
I think this can be rebased on `main` now.  containerd will also move to 1.24 for our next release (2.2).

klihub:
> @saschagrunert Now we could rebase this and try to get it merged. I tried to do that, but I was hasty with my push (to your original source branch, argh... sorry 🙈) and now I can't figure out what I screwed up compared to your [saved original branch](https://github.com/klihub/nri/commits/devel/saschagrunert/tinygo.orig), or if it is something extra that the additions since the base branch of that tree would require, since the WASM plugin compiles in your original but not in the updated version.

@saschagrunert I think I managed to figure it out now... will update with a fixed branch.

aojea (approved):

klihub (approved):

mikebrow (commented):
The matrix for go still includes 1.23.9 but during the setup for that version in switches to 1.24.3 instead.. thus we end up using 1.24.3 twice in CI..
```
Prepare all required actions
Getting action download info
Download immutable action package 'actions/setup-go@v5'
Run ./.github/actions/install-go
Run actions/setup-go@v5
Setup go version spec 1.23.9
Found in cache @ /opt/hostedtoolcache/go/1.23.9/x64
Added go to the path
Successfully set up Go version 1.23.9
go: downloading go1.24.3 (linux/amd64)
go version go1.24.3 linux/amd64
```

nit because this can be said for all the plugins.. TODO: need some readme docs for the WASM implementation... linking to https://github.com/knqyf263/go-plugin docs if nec..; explanation of special pkg/api; something for the plugin; and a test would be nice.

saschagrunert:
> The matrix for go still includes 1.23.9 but during the setup for that version in switches to 1.24.3 instead.. thus we end up using 1.24.3 twice in CI..

I removed that version from the matrix. 

> nit because this can be said for all the plugins.. TODO: need some readme docs for the WASM implementation... linking to https://github.com/knqyf263/go-plugin docs if nec..; explanation of special pkg/api; something for the plugin; and a test would be nice.

Yeah I can follow-up on docs and tests.

mikebrow (approved):
LGTM

PR #150, opened 2025-03-18
stub: don't return nil response without an error.

klihub:
If we get an event for which we have no handler, return an empty response with a nil error, instead of a nil response with a nil error. Doing the latter would result in a ttrpc marshalling error being returned, something like this `rpc error: code = Unknown desc = ttrpc: marshal called with nil`.


marquiz (commented):
Thanks @klihub 
/lgtm

fuweid (approved):

mikebrow (approved):
LGTM

Issue #153, opened 2025-03-28
[Proposal] Remove the content about 0.1.0


Iceber:
0.1.0 represents a historical design and stage for NRI, but some packages related to it still appear in the main branch.

These include:
* ./examples: Plugin examples for NRI 0.1.0
* ./skel: NRI 0.1.0 running framework
* ./types: NRI 0.1.0 API package
* ./client: NRI 0.1.0 client package
* ./README-v0.1.0.md

We may be able to delete or move these packages to ensure that the latest main branch is clean.

We could:
1. Delete ./examples and ./skel. If users really need them, they can use the NRI package from the old tag.
2. Move ./README-0.1.0.md, ./types, and ./client-go to ./plugins/v010-adapter.

Or, **we don't delete anything, just move them to the ./0.1.0 directory**. I've provided a branch to demonstrate the modified effect. https://github.com/Iceber/nri/tree/move_0.1.0

klihub:
@Iceber @mikebrow I think it would make sense to either remove the original v0.1.0 bits or move them out of the way. Other than the in-repo v010-adapter plugin which needs access to the old API bits, removing these bits would be straightforward. So maybe moving the old bits out of the way under a 0.1.0 is the best thing we can do now, like you propose in your branch.

Considering that starting from v1.7, containerd does not depend on the v0.1.0 bits, another possibility would be to just move them under something like plugins/v010-adapter/0.1.0 and consume them from that location in the adapter plugin. This could be of some inconvenience to anybody who, unbeknownst to us, is still building and using v0.1.0 plugins, but I find that highly unlikely. We could then eventually deprecate and remove the v010-adapter plugin altogether, which would then also eliminate the original v0.1.0 bits.

mikebrow:
move is fair and deprecate could be discussed.. there were a few using these 0.1.0 style..

PR #154, opened 2025-04-01
move packages related to v0.1.0 to the plugins/v010-adapter

Iceber:
issue: https://github.com/containerd/nri/issues/153

* Delete ./examples and ./skel. If users really need them, they can use the NRI package from the old tag.
* Move ./README-0.1.0.md, ./types, and ./client-go to ./plugins/v010-adapter.
* Removed the Background section from the README.md. It primarily described the differences compared to NRI v0.1.0.


README.md
```diff
@@ -23,26 +23,6 @@ The goal is to enable NRI support in the most commonly used OCI runtimes,
 [containerd](https://github.com/containerd/containerd) and
 [CRI-O](https://github.com/cri-o/cri-o).
 
-## Background
```
klihub, line 26: Should we still keep a revisited short 'Background' or 'Introduction' section which briefly tells in a few sentences what NRI is ?

klihub: Well, we have the Goal section. Maybe that's good enough then.

klihub (approved):
LGTM.

plugins/v010-adapter/0.1.0/README.md
```diff
@@ -0,0 +1,6 @@
+# Node Resource Interface v0.1.0
+
+*This version of NRI is supported through the included v010-adapter plugin.*
+
+Refer to the original documentation: [NRI v0.1.0](https://github.com/containerd/nri/blob/v0.9.0/README-v0.1.0.md).
+If you're still using *nri v0.1.0*, We recommend refactoring to the latest NRI.
```
mikebrow, line 6: bring in the background v.0.1.0 paragraph from the readme.md and add links to the now removed v010 example/skel

.... other 0.1.0 files useful for v0.1.0 cll invoke style NRI plugins..
[v0.1.0 clearcfs example](https://github.com/containerd/nri/blob/v0.9.0/examples)
[v0.1.0 skel example](https://github.com/containerd/nri/blob/v0.9.0/skel)

README.md
mikebrow, line 26: I'd leave the background section in remove the first paragraph .. move the second v0.1.0 paragraph to the new 0.1.0 readme.. and revise the 3rd with a more forward presence.. vs the change to now perspective.

mikebrow (commented):
looking good just a few more changes I think

README.md
Iceber: What do you think of adjusting the third paragraph of the reservation to read as follows:
```readme
## Background
Plugins in NRI are daemon-like entities. A single instance of a plugin is responsible for handling the full stream of NRI events and requests. A unix-domain socket is used as the transport for communication. 

NRI is defined as a formal, protobuf-based 'NRI plugin protocol' which is compiled into ttRPC bindings. This should result in improved communication efficiency with lower per-message overhead, and enable straightforward implementation of stateful NRI plugins.
```

Since the Background section no longer contains content related to 0.1.0, the section name seems a bit odd. It reads more like a description of the Plugin.
Do you think it would be better to either make it a subsection within the Goal section or just move it directly into the Goal section?
eg.
```readme
## Goal
....

### Plugins
Plugins in NRI are daemon-like entities. A single instance of a plugin is responsible for handling the full stream of NRI events and requests. A unix-domain socket is used as the transport for communication. 
...
```

mikebrow: both options work for me :-)

samuelkarp:
@Iceber Do you plan to address @mikebrow's comments?

Iceber:
@mikebrow @samuelkarp I am sorry for not handling it in a timely manner.

I have now updated it, please take a look.

mikebrow (approved):
LGTM

PR #156, opened 2025-04-07
.github: update project-checks, use local install-go action in CI workflow.

klihub:
This patch series
- updates `project-checks` to use latest tagged version
- adopts local `install-go` action from containerd
- updates the CI workflow to use same golang versions for linting and tests as containerd 1.7
- updates `golangci-lint` version

These version bumps should fix failing project-checks in CI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,10 +14,6 @@ jobs:
     timeout-minutes: 5
 
     steps:
-      - uses: actions/setup-go@v5
-        with:
-          go-version: 1.21.x
-
       - name: Set env
         shell: bash
         # TODO(thaJeztah): remove working-directory, path, and GOPATH once project-checks stops needing them; see https://github.com/containerd/nri/pull/53/commits/872fb0ce3dce136f3ae67c068ce78607565194ef#r1324366346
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -103,13 +99,13 @@ jobs:
 
     strategy:
       matrix:
-        go: [1.21.x, 1.22.3]
+        go-version: ["1.23.7", "1.24.1"]
```
fuweid, line 102: is v1.24.2 available in CI?

klihub: TBH, I haven't checked. When I filed my latest PR I ran into failures in project-checks' dependency installation which were very much looked like being cause by a too old compiler (1.21.x, for x/sys, with errors about for loops ranging over various types of integers). I wanted to get rid of that, so I looked into what is different between how containerd uses project-checks and NRI. The differences were the project-checks version, the golang toolchain versions, and the usage of the local install-go action. So I decided to update those. Then the pretty old golangci-lint we were using started failing for 1.24.1, so I bumped that too, to the latest version prior to the v2 series (I've seen in other projects bumping to 2.x causing other failures, so I thought that it's better to do that later with a separate PR).

mikebrow: nod 1.23.8 and 1.24.2 are released

mikebrow: merging up to what containerd is using seems fair... keep us out of trouble for forcing them to move up earlier than desired...

mikebrow (approved):
LGTM

fuweid (approved):

PR #157, opened 2025-04-08
api,adaptation,generate: allow adjusting linux net devices

klihub:
This PR adds NRI support for injecting 'linux net devices' a.k.a host side network interfaces into containers, building upon all the work @aojea did to make this possible.

Notes:
To test this in practice, one needs
- `runc` or `crun` with support for `LinuxNetDevice`s in the OCI Spec,
  - I tested against [runc v1.4.0-rc.1-146-g59a5ff14](https://github.com/opencontainers/runc/commit/59a5ff14a2c1f6beb74982a9c03e31c5fb49859d)
- a patched runtime,
  - I tested with this [patched containerd](https://github.com/klihub/containerd/tree/devel/main/nri-linux-net-device-injection)
- some way to inject linux net devices into containers,
  - I tested with the [updated device-injector plugin](https://github.com/klihub/nri/tree/devel/linux-net-devices/plugins/device-injector) from this PR
- and some test workload
  - I tested with a few host side VLANs and a test pod annotated for the device injector plugin
```yaml
apiVersion: v1
kind: Pod
metadata:
  name: net-dev-test
  annotations:
    network-devices.noderesource.dev/container.c0: |
      - hostIf: ens2.100
        Name: netdev0
    network-devices.noderesource.dev/container.c1: |
      - hostIf: ens2.101
        Name: netdev1
spec:
  containers:
  - name: c0
    image: busybox:latest
    command:
      - sh
      - -c
      - sleep 3600
    resources:
      requests:
        cpu: 100m
        memory: 100M
      limits:
        cpu: 100m
        memory: 100M
    imagePullPolicy: IfNotPresent
  - name: c1
    image: busybox:latest
    command:
      - sh
      - -c
      - sleep 3600
    resources:
      requests:
        cpu: 1
        memory: 100M
      limits:
        cpu: 1
        memory: 100M
    imagePullPolicy: IfNotPresent
```

```release-note
Allow adjusting Linux network devices via NRI
```


aojea:
❤️

pkg/runtime-tools/generate/generate.go
```diff
@@ -354,6 +355,19 @@ func (g *Generator) AdjustDevices(devices []*nri.LinuxDevice) {
 	}
 }
 
+// AdjustLinuxNetDevices adjusts the linux net devices in the OCI Spec.
+func (g *Generator) AdjustLinuxNetDevices(devices map[string]*nri.LinuxNetDevice) error {
+	for k, v := range devices {
+		if key, marked := nri.IsMarkedForRemoval(k); marked {
+			g.RemoveLinuNetDevice(key)
```
aojea:
```suggestion
			g.RemoveLinuxNetDevice(key)
```

pkg/runtime-tools/generate/generate.go
```diff
@@ -527,6 +541,21 @@ func (g *Generator) SetLinuxResourcesBlockIO(blockIO *rspec.LinuxBlockIO) {
...
+// AddLinuxNetDevice adds a new Linux net device.
+func (g *Generator) AddLinuxNetDevice(hostDev string, device *nri.LinuxNetDevice) {
+	if device == nil {
+		return
+	}
+	g.initConfigLinuxNetDevices()
+	g.Config.Linux.NetDevices[hostDev] = device.ToOCI()
+}
+
+// RemoveLinuNetDevice removes a Linux net device.
```
aojea:
```suggestion
// RemoveLinuxNetDevice removes a Linux net device.
```

pkg/runtime-tools/generate/generate.go
```diff
@@ -527,6 +541,21 @@ func (g *Generator) SetLinuxResourcesBlockIO(blockIO *rspec.LinuxBlockIO) {
...
+func (g *Generator) RemoveLinuNetDevice(hostDev string) {
```
aojea:
```suggestion
func (g *Generator) RemoveLinuxNetDevice(hostDev string) {
```

aojea:
This is fantastic @klihub , can we undraft it for final review

pkg/runtime-tools/generate/generate.go
klihub: Ouch. Thank you for spotting that ! I think I need new glasses... or a left brain half.

pkg/runtime-tools/generate/generate.go
klihub: Updated.

pkg/runtime-tools/generate/generate.go
klihub: Updated.

klihub:
> This is fantastic @klihub , can we undraft it for final review

We can try to do that to get it reviewed. @mikebrow WDYT?

But I think we won't be able to get this merged before the necessary new bits in runtime-spec gets behind a tag. Both CRI-O and containerd main/HEAD is at runtime-spec@v1.2.1 and I think we usually tend to stick to a tagged version in both even in main.

aojea:
ok, that is fair

klihub:
> @klihub yeah, I was using the comment as a breadcrumb , so next time I check I can track teh history, otherwise I forget
> 
> I was not trying to push for merging before that, sorry for the confusion

@aojea No prob / no offense taken. I just asked to make sure I understand correctly where we are and whether we can try to move this forward yet.

pkg/api/api.proto
```diff
@@ -377,6 +377,7 @@ message LinuxContainer {
   LinuxIOPriority io_priority = 6;
   SecurityProfile seccomp_profile = 7;
   LinuxSeccomp seccomp_policy = 8;
+  map<string, LinuxNetDevice> net_devices = 9;
```
chrishenzie: What are the keys, and why is a map chosen vs. `repeated`?

aojea: Because network interface names are unique, so there is no need to validate or fail at runtime, the keys are the interface name

plugins/device-injector/device-injector.go
```diff
@@ -286,6 +300,51 @@ func parseIOPriority(ctr string, annotations map[string]string) (*ioPrio, error)
...
+	if verbose {
+		dump(containerName(pod, ctr), "annotated network devices", devices)
+	}
+
+	for _, d := range devices {
+		a.AddLinuxNetDevice(d.HostIf, &api.LinuxNetDevice{
+			Name: d.Name,
+		})
+		if !verbose {
+			log.Infof("%s: injected network device %q -> %q...", containerName(pod, ctr),
```
chrishenzie, line 323: nit: This is unrelated since I see this is an established pattern, but I'm not following why we would conditionally write info logs. I would think we would always info log, and then conditionally decide whether or not to additionally dump verbose output above

chrishenzie (approved):

plugins/device-injector/device-injector.go
klihub: True. It is a bit strange not to log that info in verbose mode. We can fix all instances of that wrong pattern in a separate PR.

klihub: @chrishenzie #246

mikebrow (approved):
LGTM

aojea:
yay

👏

PR #164, opened 2025-04-23
ci: update linter config and add dependabot settings

ErikJiang:
- replace deprecated linter name "vet" with "govet" in .golangci.yml
- add dependabot configuration to automatically update:
  - go modules (weekly)
  - gitHub actions (weekly)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,14 @@
+version: 2
+updates:
+  - package-ecosystem: "gomod"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    groups:
+      golang-x:
+        patterns:
+          - "golang.org/x/*"
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -5,7 +5,7 @@ linters:
     - gofmt
     - goimports
     - ineffassign
-    - vet
+    - govet
     - unused
     - misspell
     - revive


.github/dependabot.yml
```diff
@@ -0,0 +1,14 @@
...
+updates:
+  - package-ecosystem: "gomod"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    groups:
+      golang-x:
+        patterns:
+          - "golang.org/x/*"
+  - package-ecosystem: "github-actions"
```
klihub, line 11: Regarding this we need to keep in mind that we are primarily a 'library package' and one which is a direct dependency for containerd and CRI-O. The general consideration regarding bumping verisons, apart from our sample plugins, are different for us than for packages which primarily generate standalone binaries. We should not greedily/eagerly bump packages which are in the intersection of our, containerd's or cri-o's direct dependencies, not above what is the current the oldest/lowest version in the union of cri-o and containerd.

ErikJiang: Makes sense. Closing this PR as automatic dependency updates may not be suitable for a core library that containerd and CRI-O depend on.

klihub: @ErikJiang Sorry if my initial comment came through as discouraging or unclear.

I didn't mean to say that there is no value or point in dependabot (or imply that the PR should be necessarily closed). I merely tried to describe why I think we can't just go ahead and always bump all of our dependencies unconditionally to the latest available versions (and how much SNR would we then get with that configuration). And I was also wondering (as I'm not familiar with configuring dependabot itself) if we could, for instance, limit the scope first. Maybe start with security alerts...

PR #160, opened 2025-04-10
api,adaptation,generate: allow setting kernel scheduling policy attributes.

klihub:
Allow plugins to set container kernel scheduling policy attributes.

This is the corresponding Process.Scheduler OCI Spec field this PR allows adjusting: https://github.com/opencontainers/runtime-spec/blob/524fc0e1b8ab0180e2fc9abd31837a0f4ed1fd6b/specs-go/config.go#L92

This is how Scheduler is defined in the OCI Spec: https://github.com/opencontainers/runtime-spec/blob/524fc0e1b8ab0180e2fc9abd31837a0f4ed1fd6b/specs-go/config.go#L38

And here are the related scheduler policy and flag definitions: https://github.com/opencontainers/runtime-spec/blob/524fc0e1b8ab0180e2fc9abd31837a0f4ed1fd6b/specs-go/config.go#L880

The related NRI types are all straightforward protobuf defined message representations of the corresponding OCI Spec types. 

```release-note
Allow setting kernel scheduling policy attributes via NRI
```

PR #166, opened 2025-04-25
Enable adjusting memory policy from NRI plugins

askervin:
This PR implements 
- container (default) Linux kernel memory policy adjustment
- container memory policies adjustment by annotation in the device-injector plugin

See set_mempolicy(2) for more details about Linux kernel memory policies.

This PR can be tested using
- this [updated containerd branch](https://github.com/klihub/containerd/tree/devel/main/nri-memory-policy) pointing to this PR
- runc version recent enough to contain https://github.com/opencontainers/runc/commit/3b75374cc797aed1a9ca933e32a97699881d6140
- the updated device-injector plugin from this PR
- and for instance this test pod
```yaml
apiVersion: v1
kind: Pod
metadata:
  name: alpine-test
  annotations:
    memory-policy.noderesource.dev/container.c0: |
      mode: MPOL_PREFERRED
      nodes: "1,2"
      flags: [ MPOL_F_STATIC_NODES ]
    memory-policy.noderesource.dev/container.c1: |
      mode: MPOL_INTERLEAVE
      nodes: "0,3"
      flags: [ MPOL_F_STATIC_NODES ]
spec:
  containers:
  - name: c0
    image: alpine:latest
    imagePullPolicy: IfNotPresent
    command:
      - sh
      - -c
      - echo c0 $(sleep inf)
    resources:
      requests:
        cpu: 1
        memory: '100M'
      limits:
        cpu: 9
        memory: '100M'
  - name: c1
    image: alpine:latest
    imagePullPolicy: IfNotPresent
    command:
      - sh
      - -c
      - echo c1 $(sleep inf)
    resources:
      requests:
        cpu: 1
        memory: '100M'
      limits:
        cpu: 9
        memory: '100M'
```

with something like this

```shell
$ kubectl apply -f test-pod.yaml
$ kubectl exec -c c0 alpine-test -- /bin/sh -c "apk add numactl-tools; numactl -s"
...
policy: [32769]
preferred node: physcpubind: 0 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 
cpubind: 0 1 2 3 
nodebind: 0 1 2 3 
membind: 0 1 2 3 
preferred: 1 2
$ kubectl exec -c c1 alpine-test -- /bin/sh -c "apk add numactl-tools; numactl -s"
...
policy: [32771]
preferred node: physcpubind: 0 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 
cpubind: 0 1 2 3 
nodebind: 0 1 2 3 
membind: 0 1 2 3 
preferred: 0 3
```

```release-note
Enable adjusting Linux memory policy from NRI plugins
```


pkg/api/adjustment.go
```diff
@@ -147,6 +147,24 @@ func (a *ContainerAdjustment) AddCDIDevice(d *CDIDevice) {
...
+	a.initLinuxMemoryPolicy()
+	a.Linux.MemoryPolicy.Nodes = nodes
+}
+
+// SetLinuxMemoryPolicyFlags records setting memory policy flags for a container.
+func (a *ContainerAdjustment) SetLinuxMemoryPolicyFlags(flags []string) {
+	a.initLinuxMemoryPolicy()
+	a.Linux.MemoryPolicy.Flags = flags
+}
+
```
klihub, line 167: If the dominant usage pattern of this is going to be calling at least both `SetLinuxMemoryPolicy{Mode,Nodes}()`, then we could also consider adding a `func (a *ContainerAdjustment) SetLinuxMemoryPolicy(mode, nodes string, flags ...string)`. (Or, in reference to my other review comment, using enum types instead of strings if we end up opting for that...)

pkg/api/api.proto
```diff
@@ -446,6 +446,14 @@ message LinuxContainerAdjustment {
...
   string cgroups_path = 3;
   OptionalInt oom_score_adj = 4;
+  LinuxMemoryPolicy memory_policy=6;
+}
+
+// Memory policy of a container being created.
+message LinuxMemoryPolicy {
+  string mode = 1;
+  string nodes = 2;
+  repeated string flags = 3;
```
klihub, line 456: We have two comparable PRs open, #155 and #160,. They are very much comparable, apart from the fact that her you are also in the process of adding the necessary changes/bits to opencontainers/runtime-spec, opencontainers/runtime-tools, and runc, whereas thoseadd controls for already existing OCI Spec features to NRI. But this difference is irrelevant in the scope of this PR (but would be more relevant in the scope of the corresponding OCI Spec PR).

Anyway, in those PR we don't use `string`s for things like kernel supported policies/`mode` or `flags` for those policies. Instead we define an explicit enumeration with the supported values, and IIRC the OCI Spec is doing the same for the controls in those PRs, by defining a custom string types and then consts for it. That is less flexible in case the kernel gains new modes/policies, but it provides a way to flag usage errors early rather than late (when someone tries to use an unknown policy or flag instead of when runc/crun tries to apply it to the container).

I'm not sure whether one would be a clear winner approach here, but I suspect that unless the kernel is gaining new policies/modes or flags frequently, the inflexibility disadvantage of using enums is negligible, while it is more self-documenting (no need to guess whether mode should be `MPOL_INTERLEAVE`, `INTERLEAVE`, or `interleave`, for instance).

klihub: Hey, and I checked and that's what you also have (custom string types and const 'enums') in the OCI Spec additions, so maybe we could follow suit with the other PRs.

askervin: My reasoning for having memory policy modes and flags as strings in the OCI spec was following what seemed to be the practice there. LinuxSchedulerPolicy, LinuxPersonalityDomain and flags, Seccomp stuff, syscalls, etc. are all strings yet they could be integers. It made sense to me, as VM-based runtimes might even use different kernels than the container runtime.

pkg/api/adjustment.go
askervin: Ended up separating all three (Mode, Nodes, Flags) into separate functions because MPOL_DEFAULT and MPOL_LOCAL require that Nodes must be not be defined. Or more exactly, the nodemask must be 0.

Of course, this does not prevent having `SetLinuxMemoryPolicy(mode, nodes string, flags ...string)` where nodes just must be empty string.

For type safety, I'd actually like to use SetLinuxMemoryPolicy(MemoryPolicyModeType, string, ...MemoryPolicyFlagType) here, but as this api does not import runtime-spec, so it would look a bit strange to have it for this case only.

Anyway, I like your "flags... string" better than "flags []string".

Let's discuss tomorrow face-to-face what would be the best combo out of all these.

pkg/api/api.proto
klihub: > My reasoning for having memory policy modes and flags as strings in the OCI spec was following what seemed to be the practice there. LinuxSchedulerPolicy, LinuxPersonalityDomain and flags, Seccomp stuff, syscalls, etc. are all strings yet they could be integers. It made sense to me, as VM-based runtimes might even use different kernels than the container runtime.

Well, in the OCI Spec/runtime-spec golang package they are not direct, native string types, but custom types with a string as the underlying type, with set of predefined consts that limit what you can (easily) pass in as value (unless you shotgun-typecast arbitrary strings to the custom type). For all of the other existing custom types, IOPrioClass, LinuxSchedulerPolicy, LinuxSchedulerFlags, maybe a few others as well, on the NRI side of the abstraction we opted to use an enum, which AFAICT is as close as you can get to the custom type+predefined const values as you can get with a proto message. You can see these aforementioned abstraction [here for I/O priority](https://github.com/containerd/nri/pull/155/files#diff-7e50ccc58c7b575d610d60547ab0f4e3062a2e80f8bf24c1f198f360e478be9b) and here for [scheduler policy attributes](https://github.com/containerd/nri/pull/160/files#diff-7e50ccc58c7b575d610d60547ab0f4e3062a2e80f8bf24c1f198f360e478be9b).

So basically the essence of my comment and main point was here that, provided we think those other abstractions are fine, and if we want to be consistent within the realm of NRI (which TBH might very well be a tall order when it comes to bits that I have added because nowadays I feel like the only thing I can do consistently is being inconsistent) then we at least should consider using protobuf enums for the suitable memory policy bits as well.

askervin: One question regarding prefixing enum values in `api.proto`.

Following the pattern in above PRs, we'll get:

```proto
// Memory policy of a container being created.
message LinuxMemoryPolicy {
  MpolMode mode = 1;
  string nodes = 2;
  repeated MpolFlag flags = 3;
}

enum MpolMode {
    MPOL_MODE_DEFAULT = 0;
    MPOL_MODE_PREFERRED = 1;
    MPOL_MODE_BIND = 2;
    MPOL_MODE_INTERLEAVE = 3;
    MPOL_MODE_LOCAL = 4;
    MPOL_MODE_PREFERRED_MANY = 5;
    MPOL_MODE_WEIGHTED_INTERLEAVE = 6;
}
```

From which protobuf generates `api.pb.go` lines:

```golang
type MpolMode int32

const (
	MpolMode_MPOL_MODE_DEFAULT             MpolMode = 0
	MpolMode_MPOL_MODE_PREFERRED           MpolMode = 1
	MpolMode_MPOL_MODE_BIND                MpolMode = 2
	MpolMode_MPOL_MODE_INTERLEAVE          MpolMode = 3
	MpolMode_MPOL_MODE_LOCAL               MpolMode = 4
	MpolMode_MPOL_MODE_PREFERRED_MANY      MpolMode = 5
	MpolMode_MPOL_MODE_WEIGHTED_INTERLEAVE MpolMode = 6
)
```

...and then a NRI plugin that modifies memory policy would do something like:

```golang
ca.SetLinuxMemoryPolicy(api.MpolMode_MPOL_MODE_PREFERRED, "0-1", api.MpolFlag_MPOL_FLAG_STATIC_NODES)
```

@klihub, would you see problems if we'd use unprefixed enums (like DEFAULT instead of MPOL_MODE_DEFAULT), because protobuf seems to generate type prefix to the enums anyway? Double prefixes are a bit clumsy, but I don't know if dropping prefixes would cause name collisions elsewhere.

askervin: @klihub, after sleeping over the weekend, I think I should name MpolMode enum values exactly as in linux/mempolicy.h. Then I can use generated MpolMode_name and _value functions for converting NRI wire values into OCI strings directly. Probably something you suggested already. :D Sorry about the noise.

klihub: Yes, if we want to avoid that double-prefix due to enum values getting 'namespaced' by their enum type, then I think calling the enum `MpolMode` then have the enum values without the `MPOL_MODE_` prefix maybe sounds to most appealing/natural, so they then would end up being `MpolMode_DEFAULT` instead of `MpolMode_MPOL_MODE_DEFAULT`. It would be close enough to the corresponding original.

An maybe slightly ugly alternative would be to call the enum type MPOL_MODE instead, so have something like 

```
 // Memory policy of a container being created.
message LinuxMemoryPolicy {
  MPOL_MODE mode = 1;
  string nodes = 2;
  repeated MpolFlag flags = 3;
}

enum MPOL_MODE {
    DEFAULT = 0;
    PREFERRED = 1;
    BIND = 2;
    INTERLEAVE = 3;
    LOCAL = 4;
    PREFERRED_MANY = 5;
    WEIGHTED_INTERLEAVE = 6;
}
```

Then if we want to offset it a bit and allow the enum type to be called MpolMode, we could add `pkg/api/memory-policy.go` (which I think anyway will be needed to provide `ToOCI` and `FromOCI` helpers) with alias like

```
// MpolMode is an alias for MPOL_MODE which we prefer to use in hand-written code.
type MpolMode = MPOL_MODE
```

We can then use MpolMode all over in our hand-written code instead of MPOL_MODE but have the MPOL_MODE_ prefix for the enum valies. If I had to chose, probably this would look nicer to my eyes. But this is just my subjective opinion. Someone else might consider not having an alias a better alternative. 

I haven't done this in my PRs, so I have double prefixes now. I could try doing something similar for the I/O Priority PR (https://github.com/containerd/nri/pull/155) where the naming is similar and rather consistent.

But unfortunately for the scheduling parameters PR (https://github.com/containerd/nri/pull/160) I think I can't have something like that...

klihub: @askervin Okay, I tried how this would look like with the I/O-prio PR (#155) mentioned above. Here is the commit on top of the current PR: https://github.com/klihub/nri/commit/9b5a7ed13471b1c383b55e205a38fbe4ce24ba8d

Dunno... Clearly doable... but I don't know if it really makes sense to have kludgery this way, and if we'd be able to do it consistently in the future if/when we add more similar features.

I suspect that the easiest to stay consistent about is one of two alternatives. Either a straightforward naive conversion of the corresponding OCI Spec types (which is what you have here and what I have in #155 and #160 which then results in some cases having repetitive prefixes). Or omit repetitive prefixes wrt. the enum type name from the enum values. So then here we'd call the enum type MpolMode and omit the repetitive MPOL_MODE_ prefix from the values, ending up with MpolMode_DEFAULT instead of MPOL_MODE_DEFAULT, which is close enough.

plugins/device-injector/device-injector.go
```diff
@@ -395,6 +409,70 @@ func parseScheduler(ctr string, annotations map[string]string) (*scheduler, erro
...
+
+	amp, err := pol.ToNRI()
+	if err != nil {
+		log.Errorf("%s: invalid memory policy annotation: %v",
+			containerName(pod, ctr), err)
+		return fmt.Errorf("invalid memory policy: %w", err)
+	}
+
+	a.SetLinuxMemoryPolicy(amp.Mode, amp.Nodes, amp.Flags...)
+	if !verbose {
```
chrishenzie: nit: Per https://github.com/containerd/nri/pull/246, I would prefer to always log info lines

plugins/device-injector/device-injector.go
```diff
@@ -55,6 +55,10 @@ const (
 	schedulerKey = "scheduling-policy.noderesource.dev"
 	// Deprecated: Prefix of the key used for scheduler attribute adjustment.
 	oldSchedulerKey = "scheduling-policy.nri.io"
+	// Prefix of the key used for memory policy adjustment.
+	memoryPolicyKey = "memory-policy.noderesource.dev"
+	// Deprecated: Prefix of the key used for memory policy adjustment.
+	oldMemoryPolicyKey = "memory-policy.nri.io"
```
chrishenzie, line 63: If this is a new feature, is there such a thing as an "old policy key"? This sounds like something that only existed before we changed domains, and I wonder if it's better to not propagate this further by allowing the old domain for new annotations

plugins/device-injector/device-injector.go
```diff
@@ -395,6 +409,70 @@ func parseScheduler(ctr string, annotations map[string]string) (*scheduler, erro
...
+	}
+
+	if err := yaml.Unmarshal(annotation, policy); err != nil {
+		return nil, fmt.Errorf("invalid memory policy annotation %q: %w", string(annotation), err)
+	}
+
+	if policy.Mode != "" {
+		policy.Mode = strings.ToUpper(policy.Mode)
+		if !strings.HasPrefix(policy.Mode, "MPOL_") {
+			policy.Mode = "MPOL_" + policy.Mode
```
chrishenzie: nit: It looks like we accept two valid values for each memory policy. Is this a pattern we've established already? I'm wondering if it would just be easier to accept only one value, so then the validation logic is cleaner.

Also, where do we validate the value for `policy.Mode`? What if a user sets it to an arbitrary string?

plugins/device-injector/device-injector.go
klihub: Updated accordingly.

plugins/device-injector/device-injector.go
klihub: This is consistent with all the other recent additions we've done since introducing the new key.

plugins/device-injector/device-injector.go
klihub: Mode and flags are validated in `memoryPolicy.ToNRI()`. Nodes we don't, letting instead runc/crun validate and fail the request if necessary.

klihub (approved):

plugins/device-injector/README.md
mikebrow: - adjustments to memory policy

mikebrow (commented):
The device injector plugin readme is getting a bit long.. WDYT about links from the top.. to each of these plugin supported adjustment types?

TODO... this and a few of the other plugins are getting useful enough that they should probably be mentioned/linked in the main readme.

klihub:
> The device injector plugin readme is getting a bit long.. WDYT about links from the top.. to each of these plugin supported adjustment types?

@mikebrow Do you mean that a link from each bullet list point at the top of `plugins/device-injector/README.md` to the related more detailed description in the same `README.md` ?

> TODO... this and a few of the other plugins are getting useful enough that they should probably be mentioned/linked in the main readme.

:+1: 

Yes, and another TODO is that I promised @samuelkarp to roll a PR for renaming `device-injector` to something more closely reflecting the current state of affairs, maybe calling it `container-adjuster`.

plugins/device-injector/README.md
klihub: Updated.

plugins/device-injector/README.md
```diff
@@ -158,6 +159,43 @@ instead of a single container. Also many of the settings are not namespaced at
...
+
+- MPOL_DEFAULT
+- MPOL_PREFERRED
+- MPOL_BIND
+- MPOL_INTERLEAVE
+- MPOL_LOCAL
+- MPOL_PREFERRED_MANY
+- MPOL_WEIGHTED_INTERLEAVE
+
+You can omit the MPOL_ prefix and can use lowercase at will.
```
chrishenzie: Is this accurate? I can't find in the device-injector implementation where we trim the prefix and support lowercase

pkg/api/memorypolicy.go
```diff
@@ -0,0 +1,90 @@
...
+
+	mpolFlags := make([]MpolFlag, len(flags))
+	for i, flag := range flags {
+		mpolFlags[i] = MpolFlag(MpolFlag_value[string(flag)])
+	}
+	return mpolFlags
+}
+
+// ToOCI returns the OCI MemoryPolicyClass corresponding to the given
+// IOPrioClass.
```
chrishenzie: Should it be this?
> to the given `MpolMode`

plugins/device-injector/README.md
mikebrow: hope not will be like optional CAP_ prefix all over again :)

mikebrow: https://github.com/containerd/nri/pull/166#discussion_r2060555580 might've been in there originally?

klihub: No, not any more. Support for that was one of the first things asked (by you IIRC) to be removed in the reviews, but forgot to update the docs. Thank you for spotting this !

mikebrow (approved):
LGTM

Issue #137, opened 2025-01-28
Configurable Restrictions of Selected NRI Features


klihub:
### Background

An increasing number of PRs or feature requests from the community propose adding new container mutation capabilities with direct security implications to NRI. See #123, #124, and #135 for some examples.

Based on the review discussions of these PRs it looks like
1. there is a desire to extend NRI with such capabilities
2. the ability to administratively lock down such capabilities would help accepting them
3. some existing capabilities in NRI could be put under such control (OCI hook injection)

### Proposed Enhancement

Extend the current implementation of NRI with the notion of restrictions.

Restrictions allow one to disable a subset of the container mutation capabilities present in NRI. Restrictions are defined in the configuration. Initially a single global set of restrictions are defined and they apply to all plugins. This can later be updated to a more fine-grained model with further improvements, if desired.

Restrictions are communicated to an NRI plugin during registration. The plugin can then report and choose not to start up if the configured restrictions prevented it from functioning properly. Initially OCI hook injection would be subject to restrictions. For backward compatibility with the current implementation, OCI hooks would default to an unrestricted configuration.

NRI request/response processing is updated to check a plugin's response for potential restriction violations before applying the requested mutations to containers. If restrictions are violated, the plugin is forcibly disconnected with an error message.

#### Protocol Additions

Define the set of possible restrictions. Initially this only consist of OCI hook injection.

```proto
message Restrictions {
    // If set to true, adjustment of OCI hooks is disabled.
    bool oci_hooks = 1;
}
```

Include configured restrictions also in the plugin configuration request.

```proto
message ConfigureRequest {
  // Any plugin-specific data, if present among the NRI configuration.
  string config = 1;
  // Name of the runtime NRI is running in.
  string runtime_name = 2;
  // Version of the runtime NRI is running in.
  string runtime_version = 3;
  // Configured registration timeout in milliseconds.
  int64 registration_timeout = 4;
  // Configured request processing timeout in milliseconds.
  int64 request_timeout = 5;
  // Extra restrictions imposed by the runtime.
  Restrictions restrictions = 6;
}
```

Add functions for checking container adjustment against restrictions.

```golang
func (r *Restrictions) CheckAdjustment(a *ContainerAdjustment) error {
	if a == nil || r == nil {
		return nil
	}
	if err := r.checkOciHookAdjustment(a); err != nil {
		return err
	}
	return nil
}

func (r *Restrictions) checkOciHookAdjustment(a *ContainerAdjustment) error {
[...]

PR #169, opened 2025-05-14
[pluggable-validation / 1]: protoize conflict detection

klihub:
This PR is the first in a stack which splits up #163 into multiple PRs, which collectively implement [configurable restrictions](https://github.com/containerd/nri/issues/137) using the [proposed pluggable validation API](https://github.com/containerd/nri/issues/137#issuecomment-2673008626) and a builtin default validator plugin.

This PR reworks conflict detection using protobuf messages to provide raw information for validation.

/cc @chrishenzie


chrishenzie (approved):
LGTM

pkg/api/api.proto
```diff
@@ -513,3 +513,63 @@ message OptionalBool {
 message OptionalFileMode {
   uint32 value = 1;
 }
+
+// CompoundFieldOwners tracks 'plugin ownership' of compound fields, typically
+// maps or slices.
+message CompoundFieldOwners {
+    map<string, string> owners = 1;
```
samuelkarp, line 520: Bringing over https://github.com/containerd/nri/pull/163#discussion_r2070947490

This needs a comment describing what is stored in the `map<string, string> owners`.  From reading through `owners.go`, it looks like:

* The key of the map is the key of the compound value (such as env var name, mount destination, annotation name)
* The value of the map is the name of the owning plugin

pkg/api/owners.go
```diff
@@ -0,0 +1,687 @@
...
+		if clearer == plugin {
+			return nil
+		}
+		other = clearer
+	}
+
+	return f.Conflict(field, plugin, other)
+}
+
+func (f *FieldOwners) ClaimCompound(field int32, key, plugin string) error {
```
samuelkarp, line 347: This doesn't appear to be called outside `owners.go` or even outside of `FieldOwners` methods.  Can it be made private?

pkg/api/owners.go
```diff
@@ -0,0 +1,687 @@
...
+	if err := f.IsCompoundConflict(field, key, plugin); err != nil {
+		return err
+	}
+
+	f.Compound[field].Owners[key] = plugin
+	return nil
+}
+
+func (f *FieldOwners) ClaimSimple(field int32, plugin string) error {
```
samuelkarp, line 356: This doesn't appear to be called outside `owners.go` or even outside of `FieldOwners` methods.  Can it be made private?

pkg/api/owners.go
```diff
@@ -0,0 +1,687 @@
...
+	m, ok := f.Compound[field]
+	if !ok {
+		return "", false
+	}
+
+	plugin, ok := m.Owners[key]
+	return plugin, ok
+}
+
+func (f *FieldOwners) SimpleOwner(field int32) (string, bool) {
```
samuelkarp, line 546: This doesn't appear to be called outside `owners.go` or even outside of `FieldOwners` methods.  Can it be made private?

pkg/api/owners.go
```diff
@@ -0,0 +1,687 @@
...
+func (f *FieldOwners) ClearArgs(plugin string) {
+	f.ClearSimple(Field_Args.Key(), plugin)
+}
+
+func (f *FieldOwners) Conflict(field int32, plugin, other string, qualifiers ...string) error {
+	return fmt.Errorf("plugins %q and %q both tried to set %s",
+		plugin, other, qualify(field, qualifiers...))
+}
+
+func (f *FieldOwners) CompoundOwner(field int32, key string) (string, bool) {
```
samuelkarp, line 532: This doesn't appear to be called outside `owners.go` or even outside of `FieldOwners` methods.  Can it be made private?

pkg/api/owners.go
```diff
@@ -0,0 +1,687 @@
...
+	m, ok := f.Compound[field]
+	if !ok {
+		m = NewCompoundFieldOwners()
+		f.Compound[field] = m
+	}
+
+	m.Owners[key] = MarkForRemoval(plugin)
+}
+
+func (f *FieldOwners) ClearSimple(field int32, plugin string) {
```
samuelkarp, line 503: This doesn't appear to be called outside `owners.go` or even outside of `FieldOwners` methods.  Can it be made private?

pkg/api/owners.go
```diff
@@ -0,0 +1,687 @@
...
+
+func (f *FieldOwners) ClaimOomScoreAdj(plugin string) error {
+	return f.ClaimSimple(Field_OomScoreAdj.Key(), plugin)
+}
+
+func (f *FieldOwners) ClaimRlimit(typ, plugin string) error {
+	return f.ClaimCompound(Field_Rlimits.Key(), typ, plugin)
+}
+
+func (f *FieldOwners) ClearCompound(field int32, key, plugin string) {
```
samuelkarp, line 493: This doesn't appear to be called outside `owners.go` or even outside of `FieldOwners` methods.  Can it be made private?

pkg/api/owners.go
```diff
@@ -0,0 +1,687 @@
...
+	}
+
+	m.Owners[key] = MarkForRemoval(plugin)
+}
+
+func (f *FieldOwners) ClearSimple(field int32, plugin string) {
+	f.Simple[field] = MarkForRemoval(plugin)
+}
+
+func (f *FieldOwners) ClearAnnotation(key, plugin string) {
```
samuelkarp, line 507: Bringing over https://github.com/containerd/nri/pull/163#discussion_r2070959731

> These `Clear` functions probably deserve a bit of a comment explaining the situations in which they should be called. It's not immediately obvious why we'd clear ownership for an existing field.

It looks like the generic answer is "when an adjustment deletes the field".

pkg/api/owners.go
```diff
@@ -0,0 +1,687 @@
...
+	return &CompoundFieldOwners{
+		Owners: make(map[string]string),
+	}
+}
+
+func (f Field) Key() int32 {
+	return int32(f)
+}
+
+func (f Field) FieldName() string {
```
samuelkarp, line 681: This looks unused.

pkg/api/strip.go
```diff
@@ -0,0 +1,343 @@
...
+	}
+
+	empty := true
+
+	if len(a.Annotations) == 0 {
+		a.Annotations = nil
+	} else {
+		empty = false
+	}
+	if len(a.Mounts) == 0 {
```
samuelkarp, line 48: Bringing over https://github.com/containerd/nri/pull/163#discussion_r2085319341

> `Mount` contains another `repeated string`, so we may need to have a strip function for that too.

go.mod
```diff
@@ -1,9 +1,12 @@
 module github.com/containerd/nri
 
-go 1.22.0
+go 1.23.0
+
+toolchain go1.23.4
```
samuelkarp, line 5: Do we need this update as part of this PR?

pkg/api/owners.go
klihub: Made it private.

pkg/api/owners.go
klihub: Made it private.

pkg/api/owners.go
klihub: Made it private.

pkg/api/owners.go
klihub: Made it private.

pkg/api/owners.go
klihub: None of these need to be called directly by a plugins. These are used internally in pkg/adaptation for ownership tracking and conflict detection. Added clarifying comments to the beginning of owners.go to that effect.

pkg/api/owners.go
klihub: Thanks for spotting that. Removed.

go.mod
klihub: Hmm... looks like we don't need it. With all the golang bumps reverted, everything still builds, golangci-lint is clean and all tests pass. Weird... I could swear at one point I had to bump. Anyway, thanks for questioning. Reverted the version bumps.

samuelkarp (approved):

samuelkarp:
> In the scope of this PR, I can (and like to) take a look at if we could get completely rid of Strip() for adjustments and updates.

If you're okay with it, my preference would be to merge this as-is to unblock the other two PRs, and then handle that in a follow-up.

klihub:
> > In the scope of this PR, I can (and like to) take a look at if we could get completely rid of Strip() for adjustments and updates.
> 
> If you're okay with it, my preference would be to merge this as-is to unblock the other two PRs, and then handle that in a follow-up.

Yes, of course. That's also fine.

mikebrow (approved):
LTGM

> It should be possible if we change result collection so that, instead of deep-initializing full skeletal container adjustment and update structs at the beginning, it would initialize on demand the necessary structure bits as and when needed.

SGTM  +++1 for follow-up

PR #170, opened 2025-05-14
[pluggable-validation / 2]: implement core adjustment validation.

klihub:
This PR is stacked on #169.

This PR is the second in a stack which splits up #163 into multiple PRs, which collectively implement [configurable restrictions](https://github.com/containerd/nri/issues/137) using the [proposed pluggable validation API](https://github.com/containerd/nri/issues/137#issuecomment-2673008626) and a builtin default validator plugin.

This PR defines the validation API and implements the core validation logic for container adjustments.

/cc @chrishenzie


pkg/adaptation/plugin.go
```diff
@@ -675,6 +680,26 @@ func (p *plugin) StateChange(ctx context.Context, evt *StateChangeEvent) (err er
 	return nil
 }
 
+func (p *plugin) ValidateContainerAdjustment(ctx context.Context, req *ValidateContainerAdjustmentRequest) error {
+	if !p.events.IsSet(Event_VALIDATE_CONTAINER_ADJUSTMENT) {
```
chrishenzie, line 684: nit: Could we make use of `isContainerAdjustmentValidator()` above since we're doing the same thing?

pkg/stub/stub.go
```diff
@@ -808,6 +815,22 @@ func (stub *stub) StateChange(ctx context.Context, evt *api.StateChangeEvent) (*
 	return &api.StateChangeResponse{}, err
 }
 
+func (stub *stub) ValidateContainerAdjustment(ctx context.Context, req *api.ValidateContainerAdjustmentRequest) (*api.ValidateContainerAdjustmentResponse, error) {
+	handler := stub.handlers.ValidateContainerAdjustment
+	if handler == nil {
+		return &api.ValidateContainerAdjustmentResponse{}, nil
```
chrishenzie, line 821: In other stub methods we return `nil` when `handler == nil`. Why might we want to return a non-`nil` response in this case?

pkg/adaptation/plugin.go
klihub: Here, I'd prefer to keep, in that regard, the boilerplate to forward the request to the plugin identical to the other boilerplates. The is*-checker is used in a different context, where it is more readable to hide the test behind a function with a self-explanatory name.

pkg/stub/stub.go
klihub: Hmm, I don't think we return anywhere a nil response when the plugins side `handler == nil`. We always return an empty-valued response in those cases, just like here.

pkg/adaptation/plugin.go
klihub: If there is a strong feeling against having both, then I think I'd rather get rid of isContainerAdjustmentValidator() and write it out in the other context.

pkg/stub/stub.go
chrishenzie: /facepalm Nevermind, I'm not sure where I was looking when I saw this and I can't `grep` for it in the repo either, I guess I hadn't had enough coffee. Closing

pkg/adaptation/adaptation.go
```diff
@@ -588,12 +633,25 @@ func (r *Adaptation) sortPlugins() {
...
+		return r.validators[i].idx < r.validators[j].idx
+	})
 	if len(r.plugins) > 0 {
 		log.Infof(noCtx, "plugin invocation order")
 		for i, p := range r.plugins {
 			log.Infof(noCtx, "  #%d: %q (%s)", i+1, p.name(), p.qualifiedName())
 		}
 	}
+	if len(r.validators) > 0 {
+		log.Infof(noCtx, "validator plugin invocation order")
```
samuelkarp, line 646: Bringing over https://github.com/containerd/nri/pull/163#discussion_r2087686495

> I thought we were planning to invoke validators in parallel? (I don't think that has to be part of _this_ PR, but I think we shouldn't promise a specific invocation order. Maybe we shouldn't sort these at all, or explicitly randomly sort them?)

samuelkarp (commented):
This mostly looks good, just one comment about validator plugin invocation order and also not going to approve this until after https://github.com/containerd/nri/pull/169 is merged.

pkg/adaptation/adaptation.go
klihub: True, we did. I updated the PR to run validation in parallel.

One more related thing I need to check though: is it goroutine-safe to access the same request locally in the builtin plugin and at the same time pass it over ttrpc to an external plugin from another goroutine. My assumption is that protoimpl.{SizeCache,MessageState} are set/used on the receiver side only, but this needs to be verified.

chrishenzie: Are there any places where we might need to update comments to explain that validation happens in parallel? (Just wondering if customers might try creating plugins with low `plugin_idx` and wonder why they don't receive priority)

samuelkarp:
@klihub Can you rebase this branch on `main` so the two already-merged commits disappear from this PR?

pkg/adaptation/adaptation.go
```diff
@@ -363,6 +380,40 @@ func (r *Adaptation) updateContainers(ctx context.Context, req []*ContainerUpdat
...
+
+	req.AddResponse(rpl)
+	req.AddOwners(result.owners)
+
+	errors := make(chan error, len(r.validators))
+	wg := sync.WaitGroup{}
+
+	for _, p := range r.validators {
+		wg.Add(1)
+		go func(p *plugin) {
```
samuelkarp, line 399: Just a note for the future that we might want to revisit this slightly and think about whether we want to limit the concurrency a bit; might become an issue if there are many validator plugins.  No need to change as part of this PR.

samuelkarp (approved):

pkg/adaptation/adaptation.go
klihub: Yes, I think it will be a good idea to emphasize in the documentation that during the validation phase there is no 'plugin ordering by index' as validating plugins are run in parallel.

But regarding users wondering "why they don't get priority", I think they should not. Any presumed ordering/priority would be relevant only if it was possible to force validation both to fail and to succeed. But our semantics is deliberately such that the latter is not possible.

pkg/adaptation/adaptation.go
klihub: > Are there any places where we might need to update comments to explain that validation happens in parallel? (Just wondering if customers might try creating plugins with low `plugin_idx` and wonder why they don't receive priority)

@chrishenzie Yes, I think it will be a good idea to emphasize in the documentation that during the validation phase there is no 'plugin ordering by index' as validating plugins are run in parallel.

But regarding users wondering "why they don't get priority", I think they should not. Any presumed ordering/priority would be relevant only if it was possible to force validation both to fail and to succeed. But our semantics is deliberately such that the latter is not possible.

klihub:
> True, we did. I updated the PR to run validation in parallel and to not log/imply any particular validation invocation order.
> 
> One more related thing I need to check though: is it goroutine-safe to access the same request locally in the builtin plugin and at the same time pass it over ttrpc to an external plugin from another goroutine. My assumption is that protoimpl.{SizeCache,MessageState} are set/used on the receiver side only, but this needs to be verified.

@samuelkarp @chrishenzie Okay, so while adding test cases for the default validator in #171 it became obvious that running validation with multiple plugins and at least one of them being a builtin one (the default validator) is not parallel/goroutine safe. But the reason is not protobuf itself. Rather it is simply that in the current implementation a query/lookup of a field owner of a particular container is (counterintuitively) not a read-only operation if the container has not been adjusted at all. In that case it has no FieldOwners entry at all and that then gets implicitly created during the lookup (with a write to a map). I fixed this in a separate commit in #171 prior to adding the test cases.

pkg/adaptation/adaptation.go
mikebrow: true and users of the api may want ordering regardless of our scheme...   E.g. I will only approve after ___ approval, if my check is also valid.  Allocate claim foo before claim foo'...

mikebrow: typically one would have a set operating in parallel at a level where the higher levels must finish first?  built-in, set-a.. set b with a dependency in set-a..

mikebrow (approved):
LGTM

pkg/adaptation/adaptation.go
samuelkarp: > E.g. I will only approve after ___ approval, if my check is also valid

The validations are a logical AND, so if any fails the container is rejected.  I don't think you can model "I will only approve after $foo approval" since if $foo rejects the container is rejected anyway, and if $foo succeeds then you'd also succeed your validation?

PR #173, opened 2025-05-22
don't exit on close

aojea:
It is not a good practice for a library to unilaterally decide to terminate the entire program and, since it is imported, is super hard to debug for nri plugins authors that only see the program exit without any obvious reason.

Change-Id: Ic72612d4e8421a2eb301d33dda760765461929d7

### DIFF
--- a/pkg/stub/stub.go
+++ b/pkg/stub/stub.go
@@ -504,7 +504,7 @@ func (stub *stub) Run(ctx context.Context) error {
 
 	err = <-stub.srvErrC
 	if err == ttrpc.ErrServerClosed {
-		return nil
+		log.Infof(noCtx, "ttrpc server closed %s : %v", stub.Name(), err)
 	}
 
 	return err
@@ -591,8 +591,6 @@ func (stub *stub) connClosed() {
 		stub.onClose()
 		return
 	}
-
-	os.Exit(0)
 }
 
 //


klihub:
> all credit to @aktau that help me to find this after debugging this problem for a few days/
> 
> @samuelkarp @klihub I do not if we can surface an error or something if the desire is to signal this to the plugin developers, but the os.Exit is really a bad experience in a library 😅

@aojea This has been reported a few times. And so far I've been reluctant to change it, but my maybe my foggy brain is maybe starting to get the memo...  despite that there is a really simple (but probably extremely badly documented or even totally undocumented) programmatic way to opt out from this default behavior: simply create the stub with an OnClose handler. In its simplest form this option should do it: `WithOnClose(func() {})`. The stub won't exit if the plugin has an OnClose handler.

Maybe this was/is a really bad choice, but the reasoning behind this default behavior has been that if you don't install an OnClose handler, then you are not actively watching for the connection state so probably you won't realize that you lost connection and either give up or try to reconnect.

aojea:
It is not obvious you need to set up an OnClose() hook ...
and if you do not set the hook and the program exits clean it is impossible to debug for people unless  you have a lot of knowledge on debugging golang applications, in this case I'm lucky @aktau guided me to get the stack with `strace` and find this call in the library.

```
pid 74982] 08:30:12.290405 +++ exited with 0 +++
[pid 74981] 08:30:12.290414 +++ exited with 0 +++
 > /dranet(runtime.exit.abi0+0xb) [0x7d16b]
 > /dranet(os.Exit+0x38) [0x107178]
 > /dranet(github.com/containerd/nri/pkg/stub.(*stub).connClosed+0x7c) [0xd1e11c]
 > /dranet(github.com/containerd/nri/pkg/stub.(*stub).Start.func4+0x17) [0xd1d217]
 > /dranet(github.com/containerd/ttrpc.(*Client).run+0x53) [0xb32613]
 > /dranet(github.com/containerd/ttrpc.NewClient.gowrap1+0x25) [0xb314e5]
 > /dranet(runtime.goexit.abi0+0x1) [0x7ba21]
[pid 74980] 08:30:12.290445 +++ exited with 0 +++
[pid 74979] 08:30:12.290453 +++ exited with 0 +++
[pid 74978] 08:30:12.290461 +++ exited with 0 +++
[pid 74977] 08:30:12.290478 +++ exited with 0 +++
[pid 74976] 08:30:12.290498 +++ exited with 0 +++
[pid 74975] 08:30:12.290508 +++ exited with 0 +++
```


In Kubernetes you only see a Pod restart with exist status 0, that makes further debugging impossible , this is really bad for nri plugins in production


> Maybe this was/is a really bad choice, but the reasoning behind this default behavior has been that if you don't install an OnClose handler, then you are not actively watching for the connection state so probably you won't realize that you lost connection and either give up or try to reconnect.

I see, but I think that this is not the right solution since the plugins runs with

```
	// Run starts the plugin then waits for the plugin service to exit, either due to a
	// critical error or an explicit call to Stop(). Once Run() returns, the plugin can be
	// restarted by calling Run() or Start() again.
	Run(context.Context) error
```

I think plugins authors should handle the lifecycle in the Run method that is blocking.

Submitted a new approach, where we return a well known error so plugin author can also assert on the specific error

pkg/stub/stub.go
```diff
@@ -592,7 +593,12 @@ func (stub *stub) connClosed() {
 		return
 	}
 
-	os.Exit(0)
+	// the connection has closed, make a best effort to inform the plugin author about it
+	select {
+	case stub.srvErrC <- ErrNoService:
```
aojea, line 598: hmm, this is not making the plugin to exit?

aojea: @klihub if we loose the connection shouldn't we exit the Run method with an error?

aojea: hmm, it returns 

>  ttrpc: server closed

and Run returns with nil

aojea: > Maybe this was/is a really bad choice, but the reasoning behind this default behavior has been that if you don't install an OnClose handler, then you are not actively watching for the connection state so probably you won't realize that you lost connection and either should give up and exit or try to reconnect.

this contradicts the behavior of Run, so if the connection is closed , Run() exists with nil but the handlers exits the program entirely

pkg/stub/stub.go
```diff
@@ -504,7 +505,7 @@ func (stub *stub) Run(ctx context.Context) error {
 
 	err = <-stub.srvErrC
 	if err == ttrpc.ErrServerClosed {
-		return nil
```
aojea, line 507: if we swallow the error, plugin authors will not be able to implement the logic to handle it and it will be assumed everything exited clean

pkg/stub/stub.go
klihub: > > Maybe this was/is a really bad choice, but the reasoning behind this default behavior has been that if you don't install an OnClose handler, then you are not actively watching for the connection state so probably you won't realize that you lost connection and either should give up and exit or try to reconnect.
> 
> this contradicts the behavior of Run, so if the connection is closed , Run() exists with nil but the handlers exits the program entirely

@aojea Yes, you are right. It's much better to not `os.Exit()` on connection close even if there is no OnClose handler and instead return any error we might get from `Run()`.

klihub (approved):
LGTM.

samuelkarp (approved):

PR #181, opened 2025-07-03
ci: use latest go patch version

saschagrunert:
There is no reason to stick to 1.24.3, means we can re-introduce the go1.24.x tag to allow newer ones in CI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -74,7 +74,7 @@ jobs:
 
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


djdongjin (approved):

klihub:
I'm fine with this. But maybe we should also make sure that in CI we additionally always test with the exact same version than recent enough containerd release branches ? So maybe/probably 2.0, 2.1, and main, which is then 1.24.4 AFAICT...

Issue #152, opened 2025-03-27
[RFC] Plugins authentication proposal


kad:
In the light of the recent discussions about what plugins are allowed to do or not, I think we need to implement mechanism for plugin authentication to the runtime. 

This would allow Runtime to:
- define on per-plugin basis what kind of operations is allowed to be performed to plugin (e.g. disable modifications of hooks, or disable modification of anything else)
- have default security for unauthenticated plugins
- uniquely identify plugin by plugin's public key (similar to Wiregiard peer identification)
- make sure that one plugin does not try to fake other plugin's name during registration

For authentication, I propose to use challenge/response pattern during plugin registration/configuration phases.

Challenge/response data can be encrypted via Curve25519 / ChaCha20-Poly1305 that are available via standard Go libraries, and are quite common asymmetric encryption algorithms, again similar to what is used in Wireguard.

For protocol, we can easily extend current messages, as demonstrated in following sequence diagram:

![Image](https://github.com/user-attachments/assets/46ce29b9-cc2b-44ab-8cec-405ed6680b36)

For protocol changes we can do something similar to this:
```patch
diff --git a/pkg/api/api.proto b/pkg/api/api.proto
index 78fa64d..25bb500 100644
--- a/pkg/api/api.proto
+++ b/pkg/api/api.proto
@@ -26,7 +26,7 @@ option go_package = "github.com/containerd/nri/pkg/api;api";
 // The rest of the API is defined by the Plugin service.
 service Runtime {
     // RegisterPlugin registers the plugin with the runtime.
-    rpc RegisterPlugin(RegisterPluginRequest) returns (Empty);
+    rpc RegisterPlugin(RegisterPluginRequest) returns (RegisterPluginResponse);
     // UpdateContainers requests unsolicited updates to a set of containers.
     rpc UpdateContainers(UpdateContainersRequest) returns (UpdateContainersResponse);
 }
@@ -36,6 +36,13 @@ message RegisterPluginRequest {
     string plugin_name = 1;
     // Plugin invocation index. Plugins are called in ascending index order.
     string plugin_idx = 2;
+    // Plugin authentication public key
+    string plugin_public_key = 3;
+}
+
+message RegisterPluginResponse {
+    // Enum, descibes sets of capabilities that plugin allowed to use.
+    repeated int32 allowed_capabilities = 1;
 }
 
 message UpdateContainersRequest {
@@ -141,12 +148,18 @@ message ConfigureRequest {
   int64 registration_timeout = 4;
[...]

PR #183, opened 2025-07-16
implement plugin authentication, finer-grained restrictions/validation.

klihub:
This PR implements plugin authentication. The sole purpose of authenticated plugins is to allow finer-grained restrictions of selected NRI features than just a global on/off. This PR also updates the builtin default validator to allow finer-grained restrictions using overrides for authenticated plugins. 

In particular the patch series
- defines the wire protocol bits for authentication
- implements a challenge-response authentication mostly based on the ideas in #152
- implements runtime and plugin side support for authentication
- implements overrides for authenticated plugins in the default validator

Here are working trees of updated runtimes with support for authenticated NRI plugins and finer-grained validation using an NRI tree with this PR:
- containerd/main: https://github.com/klihub/containerd/tree/devel/nri-plugin-authentication
- cri-o/main: https://github.com/klihub/cri-o/tree/devel/nri-plugin-authentication


pkg/adaptation/plugin.go
```diff
@@ -428,6 +447,74 @@ func (p *plugin) qualifiedName() string {
...
+
+	algo, err := impl.NewWithEphemeralKeys()
+	if err != nil {
+		log.Errorf(ctx, "rejecting client, failed to set up authentication: %v", err)
+		p.close()
+		return nil, err
+	}
+
+	p.role.algo = algo
+	chal, key, err := p.role.algo.Challenge([]byte("XXtmpXXX"), req.PublicKey)
```
chrishenzie: Judging by the name, I'm assuming this will be replaced? Should this be a random seed?

klihub: Argh... that was already supposed to be replaced by a random seed. Thanks for spotting it !

PR #188, opened 2025-07-26
adaptation: Export result library

chrishenzie:
The result collection and merging logic in the adaptation package is useful for NRI plugins that manage a set of their own sub-plugins under the same binary.

Exports the necessary types and methods to make this logic available to other packages.

The following changes were made:
- The `result` type is now exported as `Result`
- Constructor functions (`collect*`) are now exported (`New*`)
- Methods for applying and retrieving responses are now exported
- Internal usages have been updated to the new exported API

This enables building more complex plugins that can intelligently merge results from multiple sources.

/cc @klihub


klihub (approved):
LGTM.

Out of curiosity, do you already have some concrete use in mind for this ? I'm asking because I can't help wondering in what situation for instance a 'pre-collector' using this would really be the right integration point to (hide and) combine multiple plugin logic to a seeminly single...

chrishenzie:
Yes, I have an NRI plugin that serves as a "common" plugin and manages a set of sub-plugins for various internal use cases. The purpose of having a "common" plugin is to reduce the barrier of entry for adding new plugin logic (setting up new repos, build flows, etc.), but isn't a substitute for larger, more complex plugins.

klihub:
> Yes, I have an NRI plugin that serves as a "common" plugin and manages a set of sub-plugins for various internal use cases. The purpose of having a "common" plugin is to reduce the barrier of entry for adding new plugin logic (setting up new repos, build flows, etc.), but isn't a substitute for larger, more complex plugins.
> 
> I'm open to any feedback on better ways to aggregate the changes across a set of "sub-plugins" though (if this library isn't well suited for it).

If you don't expect the 'sub-plugins' to overlap functionally, IOW they don't request adjustments of the same container parameters which could then conflict, then the most straightforward approach might be to simply organize each sub-plugins's code so that the adjustment is done in an (sub-plugin package exported) CreateContainer function which takes an extra `*nri.pkg/api.ContainerAdjustment{}` argument, and uses the provided 'member' functions to add there its own adjustments. Then in the main plugin's CreateContainer you create an empty adjustment and pass it to all sub-plugins. You need to do extra locking if you want to make those calls in parallel from multiple goroutines.

mikebrow (commented):
it is fair, I think, to allow this plugin manager pattern be expressed by a plugin, esp. for the performance and collection of.. benefits..

But can / should we not keep an internal result to coincide with the Exported?  and possibly introduce a WithFunc() pattern for extensions? 

Or is that to heavy a pattern for the adaptation package

PR #191, opened 2025-08-06
Publish reference plugin container images

### LINKED ISSUE #190: [Proposal] Build and publish container images for reference plugins
The reference plugins are useful in testing and development and also used in some real-world deployments. However, currently these reference plugins are not built and published as container images, which is cumbersome for users who want to try them out quickly or run them more permanently in containerized environments.

I propose that we start to publish container images of the reference plugins in ghcr.io.  (e.g., ghcr.io/containerd/nri/plugins/<name>). We could take a piecemeal approach of enabling only selected plugins.

Benefits would include easier adoption and testing of the reference plugins and promoting consistent builds across environments.

Steps/tasks:
1. Write docker files to the (selected) plugins 
1. CI Integration
    - GitHub workflows for building and pushing images
    - publish "unstable" images from the main branch
    - publish tagged images from releases
1. Documentation
    - update README to reference the available container images)
    - possibly also example yaml(s)/snippet(s) for deployment

marquiz:
This is a minimal PR for enabling building and pushing container images for the reference plugins.

Contains two parts:
1. **Add Dockerfile for building images**
    Add a generic dockerfile for building container images of the reference plugins. The build command creates statically linked binaries and the resulting images are minimal, based on scratch base image.
2. **Add github workflow for publishing container images**
    Publish multiarch container images of the reference plugins at ghcr.io/containerd/nri/plugins/<plugin>:<tag>.

    The tag is the git tag name for tagged versions and "unstable" for the main branch.

    The default-validator, network-device-injector and network-logger plugins are skipped in the config as they are also missing from the Makefile.

Fixes #190 

**NOTE:** workflow tested in my personal repo https://github.com/marquiz?tab=packages&repo_name=nri

**TODO:**
- [ ] Documentation


.github/workflows/images.yml
```diff
@@ -0,0 +1,79 @@
...
+          device-injector,
+          differ,
+          hook-injector,
+          logger,
+          #network-device-injector,
+          #network-logger,
+          template,
+          ulimit-adjuster,
+          v010-adapter,
+          wasm,
```
klihub: nit: I don't think we should build an image for the wasm sample plugin as well. The only way to activate is to drop or link it in to the stock pre-installed plugin directory then let the NRI core/the runtime discover it, realize it is a WASM binary and load it as a WASM plugin.

marquiz: Check, will remove in an update. I just brainlessly "enabled" all

klihub:
@marquiz This looks pretty good.

For me the main remaining question is whether this is useful enough already as such, or if we should consider adding either a sample deployment file or a Helm chart for the built plugins. There are some extra bits from the host containerized plugins need to function, so it could be useful to provide a bit more sugar-coating for the sample plugins.

All plugins need access to the NRI socket from the host to be able to run containerized. Additionally some of the plugins need access to extra bits from the host to be able to function. For instance, the `hook-injector` plugin needs access at least to `/etc/containers/oci/hooks.d` for the hook manager to work. The `v010-adapter` needs access to at least to the original `0.1.0`-specific `/etc/nri/conf.json` configuration file and `/opt/nri/bin` plugin binary directory.

marquiz:
Rebased. s/reference plugin/sample plugin/ in PR description and commit messages.

marquiz:
Updated: added CI check to test build images on PRs

.github/workflows/images.yml
```diff
@@ -0,0 +1,83 @@
...
+              fi
+          fi
+          if [ -z "$tag" ]; then
+              echo "ERROR: failed to determine image tag"
+              exit 1
+          fi
+          echo "TAG_NAME=$tag" >> $GITHUB_ENV
+
+      - name: Build and push image
+        uses: docker/build-push-action@v5
```
mikebrow: maybe?
```suggestion
        uses: docker/build-push-action@v6
```

mikebrow (commented):
should we start with multi-arch and use daemon-config "containerd-snapshotter": true

.github/workflows/images.yml
marquiz: Ach, good catch. Fixed

klihub (approved):

PR #192, opened 2025-08-07
Makefile: build all plugins by default.

klihub:
Add missing targets to build also the `network-device-injector` and `network-logger` plugins.


klihub:
> Looks good. One question though, what about the default-validator plugin?

It cannot be built as an external plugin without minimal external boilerplate because it is also used as a built-in plugin, directly linked in to the runtime binary. We used to have the necessary boilerplate and a Makefile target to build it as a standalone sample plugin, but it was decided to drop it in the review rounds before merging the validation PR.

marquiz:
@chrishenzie @mikebrow

chrishenzie (approved):

PR #193, opened 2025-08-07
readme: list all sample plugins

marquiz:
Also add minimal readme for the template and wasm plugins.

The default-validator plugin is left out as it cannot be built as an external plugin (without changes).


marquiz:

mikebrow (approved):
LGTM

PR #196, opened 2025-08-07
Add minimal documentation for plugin images

marquiz:

PR #197, opened 2025-08-07
deployment: add minimal kustomize overlays for deploying plugins

marquiz:


kustomize/device-injector/daemonset.yaml
```diff
@@ -0,0 +1,24 @@
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: nri-plugin-device-injector
+spec:
+  template:
+    spec:
+      containers:
+        - name: plugin
+          image: "ghcr.io/containerd/nri/plugins/device-injector:unstable"
```
klihub, line 10:
```suggestion
          image: logger-image
```

kustomize/device-injector/kustomization.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+namespace: kube-system
+
+resources:
+- daemonset.yaml
+
```
klihub:
```suggestion

images:
- name: "ghcr.io/containerd/nri/plugins/device-injector"
#  newName: "ghcr.io/containerd/nri/plugins/device-injector"
#  newTag: "latest"

```

klihub: @marquiz I'd suggest a placeholder like this (and similar ones for the other plugins) to help those not that familiar with kustomize easily change the image tag.

klihub: And I think it would be a good idea to set some small but reasonable CPU and memory requests among the daemonset resources.

kustomize/device-injector/daemonset.yaml
```diff
@@ -0,0 +1,24 @@
...
+          imagePullPolicy: Always
+          args:
+            - "-idx"
+            - "10"
+            - "-name"
+            - "device-injector"
```
klihub: I think this could be now left out, since you set it in the environment of the image.

marquiz: I must be left out. Otherwise the plugin refuses to start

kustomize/v010-adapter/daemonset.yaml
```diff
@@ -0,0 +1,24 @@
...
+          args:
+            - "-idx"
+            - "99"
+            - "-name"
+            - "v010-adapter"
+          volumeMounts:
+            - name: nri-socket
+              mountPath: /var/run/nri/nri.sock
+      volumes:
+      - name: nri-socket
```
klihub, line 21: This would also require additional bits from the host for the original hookesque v0.1.0 NRI interface to work. Those would be by default the config file `/etc/nri/conf.json` and the old plugin binary directory `/opt/nri/bin`.

kustomize/device-injector/daemonset.yaml
klihub:
```suggestion
```

kustomize/hook-injector/daemonset.yaml
```diff
@@ -0,0 +1,45 @@
...
+    spec:
+      containers:
+        - name: plugin
+          image: "ghcr.io/containerd/nri/plugins/hook-injector:unstable"
+          imagePullPolicy: Always
+          args:
+            - "-idx"
+            - "10"
+            - "-name"
+            - "hook-injector"
```
klihub:
```suggestion
```

kustomize/logger/daemonset.yaml
```diff
@@ -0,0 +1,24 @@
...
+    spec:
+      containers:
+        - name: plugin
+          image: "ghcr.io/containerd/nri/plugins/logger:unstable"
+          imagePullPolicy: Always
+          args:
+            - "-idx"
+            - "10"
+            - "-name"
+            - "logger"
```
klihub:
```suggestion
```

kustomize/network-device-injector/daemonset.yaml
```diff
@@ -0,0 +1,24 @@
...
+    spec:
+      containers:
+        - name: plugin
+          image: "ghcr.io/containerd/nri/plugins/network-device-injector:unstable"
+          imagePullPolicy: Always
+          args:
+            - "-idx"
+            - "10"
+            - "-name"
+            - "network-device-injector"
```
klihub:
```suggestion
```

kustomize/network-logger/daemonset.yaml
```diff
@@ -0,0 +1,24 @@
...
+    spec:
+      containers:
+        - name: plugin
+          image: "ghcr.io/containerd/nri/plugins/network-logger:unstable"
+          imagePullPolicy: Always
+          args:
+            - "-idx"
+            - "10"
+            - "-network-logger"
+            - "network-logger"
```
klihub:
```suggestion
```

kustomize/template/daemonset.yaml
```diff
@@ -0,0 +1,24 @@
...
+    spec:
+      containers:
+        - name: plugin
+          image: "ghcr.io/containerd/nri/plugins/template:unstable"
+          imagePullPolicy: Always
+          args:
+            - "-idx"
+            - "99"
+            - "-name"
+            - "template"
```
klihub:
```suggestion
```

kustomize/ulimit-adjuster/daemonset.yaml
```diff
@@ -0,0 +1,24 @@
...
+    spec:
+      containers:
+        - name: plugin
+          image: "ghcr.io/containerd/nri/plugins/ulimit-adjuster:unstable"
+          imagePullPolicy: Always
+          args:
+            - "-idx"
+            - "99"
+            - "-name"
+            - "ulimit-adjuster"
```
klihub:
```suggestion
```

kustomize/v010-adapter/daemonset.yaml
```diff
@@ -0,0 +1,24 @@
...
+    spec:
+      containers:
+        - name: plugin
+          image: "ghcr.io/containerd/nri/plugins/v010-adapter:unstable"
+          imagePullPolicy: Always
+          args:
+            - "-idx"
+            - "99"
+            - "-name"
+            - "v010-adapter"
```
klihub:
```suggestion
```

kustomize/v010-adapter/daemonset.yaml
```diff
@@ -0,0 +1,24 @@
...
+          volumeMounts:
+            - name: nri-socket
+              mountPath: /var/run/nri/nri.sock
+      volumes:
+      - name: nri-socket
+        hostPath:
+          path: /var/run/nri/nri.sock
+          type: Socket
```
klihub:
```suggestion
          volumeMounts:
            - name: nri-socket
              mountPath: /var/run/nri/nri.sock
            - name: old-nri-config
              mountPath: /etc/nri/conf.json
              readOnly: true
            - name: old-nri-plugins
              mountPath: /opt/nri/bin
              readOnly: true
      volumes:
      - name: nri-socket
        hostPath:
          path: /var/run/nri/nri.sock
          type: Socket
      - name: old-nri-config
        hostPath:
          path: /etc/nri/conf.json
          type: File
      - name: old-nri-plugins
        hostPath:
          path: /opt/nri/bin
          type: DirectoryOrCreate
```

marquiz:
> Can we move these into a `contrib` folder rather than having `kustomize` at the root of the repo?

Moved to contrib/kustomize

kustomize/device-injector/kustomization.yaml
marquiz: > I'd suggest a placeholder like this (and similar ones for the other plugins) to help those not that familiar with kustomize easily change the image tag.

I changed the configs so that the image name in every daemonset.yaml is a placeholder ( `plugin:latest`)  and then we set that with `newName` and `newTat` in the kustomization.yaml. So no placeholder but actually use those.

> And I think it would be a good idea to set some small but reasonable CPU and memory requests among the daemonset resources.

Yes, makes sense. Set request `cpu: 2m` and `memory: "5Mi"` for every plugin.

kustomize/device-injector/daemonset.yaml
marquiz: Fixed

marquiz: Fixed

kustomize/hook-injector/daemonset.yaml
marquiz: Fixed

kustomize/logger/daemonset.yaml
marquiz: Fixed

kustomize/network-logger/daemonset.yaml
marquiz: Fixed

kustomize/network-device-injector/daemonset.yaml
marquiz: Fixed

kustomize/template/daemonset.yaml
marquiz: Fixed

kustomize/ulimit-adjuster/daemonset.yaml
marquiz: Fixed

kustomize/v010-adapter/daemonset.yaml
marquiz: Fixed

kustomize/v010-adapter/daemonset.yaml
marquiz: Changed as suggested (I used the name `nri-v010-{config,plugins}` though)

klihub (approved):
Thank you ! This now LGTM.

contrib/kustomize/device-injector/daemonset.yaml
```diff
@@ -0,0 +1,26 @@
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: nri-plugin-device-injector
+spec:
+  template:
+    spec:
```
mythi: for some of the plugins, `priorityClassName: system-node-critical` might make sense?

marquiz: > for some of the plugins, `priorityClassName: system-node-critical` might make sense?

Perhaps. Any suggestions which? device-injector, hook-injector, network-device-injector, ulimit-adjuster, v010-adapter?

@klihub @samuelkarp @chrishenzie ?

klihub: Yes, for those it could make sense.

marquiz: K, added to those

contrib/kustomize/logger/daemonset.yaml
```diff
@@ -0,0 +1,33 @@
...
+    spec:
+      containers:
+        - name: plugin
+          image: plugin:latest
+          imagePullPolicy: Always
+          args:
+            - "-idx"
+            - "10"
+            - "-events"
+            - "RunPodSandbox,StopPodSandbox,RemovePodSandbox,CreateContainer,PostCreateContainer,StartContainer,PostStartContainer,UpdateContainer,PostUpdateContainer,StopContainer,RemoveContainer"
```
chrishenzie: Why not `"all"`?

contrib/kustomize/logger/daemonset.yaml
```diff
@@ -0,0 +1,33 @@
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: nri-plugin-logger
+spec:
+  template:
+    spec:
+      containers:
+        - name: plugin
```
chrishenzie: nit: Indentation (compared with the indentation used for `volumes` below)

chrishenzie (approved):
Generally LGTM. I have a few nits about the indentation of list items in the yamls but that's no big deal.

Would it be possible to drop a `README` in there that gives a TL;DR on how to use `kustomize` to install these (for folks like myself who are unfamiliar)?

contrib/kustomize/logger/daemonset.yaml
klihub: Because then the plugin image would not be able to start up out of the box. The reason is that we have added new events since the last NRI version update(s) in the runtime(s). So using 'all' would try to register to events currently unknown to the runtimes, until we until we (tag a new NRI release and) bump their NRI dependency.

contrib/kustomize/logger/daemonset.yaml
marquiz: Thx. Fixed.

contrib/kustomize/logger/daemonset.yaml
marquiz: Yeah, with "all" there is no runtime implementation that would be compatible with this (even cri-o main fails)

contrib/kustomize/device-injector/daemonset.yaml
```diff
@@ -0,0 +1,32 @@
...
+      priorityClassName: system-node-critical
+      containers:
+        - name: plugin
+          image: plugin:latest
+          imagePullPolicy: Always
```
mikebrow:
```suggestion
          imagePullPolicy: IfNotPresent
```

contrib/kustomize/device-injector/daemonset.yaml
```diff
@@ -0,0 +1,32 @@
...
+kind: DaemonSet
+metadata:
+  name: nri-plugin-device-injector
+spec:
+  template:
+    spec:
+      priorityClassName: system-node-critical
+      containers:
+        - name: plugin
+          image: plugin:latest
```
mikebrow:
```suggestion
          image: plugin:latest # image and tag "plugin:latest" are placeholders substituted via config
```

contrib/kustomize/device-injector/daemonset.yaml
```diff
@@ -0,0 +1,32 @@
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: nri-plugin-device-injector
+spec:
+  template:
+    spec:
+      priorityClassName: system-node-critical
+      containers:
+        - name: plugin
```
mikebrow:
```suggestion
        - name: plugin # plugin name "plugin" is a placeholder substituted via config
```

contrib/kustomize/device-injector/daemonset.yaml
klihub: IIUC, the default [ck]ustomization updates the daemonset to use the `unstable` image tag, and I thought that was the reason behind using an `Always` `imagePullPolicy`, which would make sense to me (to avoid countless late night WTF?-staring sessions at the terminal when testing).

@marquiz @mikebrow If we change the policy here in the daemonset declaration, can we then also kustomize it to `Always` in the same file where we change the tag to `unstable` ?

contrib/kustomize/device-injector/kustomization.yaml
```diff
@@ -0,0 +1,13 @@
...
+resources:
+  - daemonset.yaml
+images:
+  - name: plugin
+    newName: ghcr.io/containerd/nri/plugins/device-injector
+    newTag: unstable
+labels:
+  - includeSelectors: true
+    pairs:
+      app.kubernetes.io/name: nri-plugin-device-injector
```
klihub:
```suggestion
      app.kubernetes.io/name: nri-plugin-device-injector

patchesStrategicMerge:
- |
  apiVersion: apps/v1
  kind: DaemonSet
  metadata:
    name: nri-plugin-logger
  spec:
    template:
      spec:
        containers:
        - name: plugin
          imagePullPolicy: Always
```

klihub: So maybe with something like this...

contrib/kustomize/device-injector/daemonset.yaml
marquiz: Yeah, the idea was exactly that. The `unstable` is a rolling tag, always following the tip of the main branch. We don't have any "stable" (tagged by version tag) images yet and pull policy Always ensures that you get the latest devel build, not some weeks or months old.

I now dropped the imagePullPolicy field from the daemonset.yaml's alltogether and added a common kustomize component (included in all plugin overlays) that patches the pull policy to Always.

contrib/kustomize/device-injector/kustomization.yaml
marquiz: I added a common component to avoid repeating this in every overlay. (note/nit: patchesStrategicMerge is deprecated, use patches instead)

klihub (approved):
@marquiz Thank you ! LGTM.

klihub:
> Major update: separate overlays for stable and unstable images
...
> 2. Install a specific release
>    ```
>    kubectl create -k "https://github.com/containerd/nri/contrib/kustomize/hook-injector?ref=v0.10.0"
>    ```

@marquiz I think this would be worth including in the documentation additions in PR (#196), because I suspect it will not be obvious for the average user.

contrib/kustomize/device-injector/daemonset.yaml
mikebrow: `always` image pull policy is good for forcing verification of access to private authentication required registries for pods in a multi-tenant situation...

`if not present` image pull policy is good for allowing pods to share container images that have already been cached

`:latest` tag is good for asking to check if there is a newer image on the registry every time

`:version` tags and `SHA`s  are good for requesting a specific version of an image

mikebrow: we can discuss on the community call

mikebrow: kk discussed with klihub just need to mention somewhere in these that this is not for production use .. rather dev testing purposes..

PR #198, opened 2025-08-08
Makefile: simplify plugin build make targets

marquiz:
Use a pattern rule and have a special case for the WASM plugin. Also remove the wildcard so that the plugin(s) are rebuilt e.g. in the case dependencies are updated.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -98,46 +98,14 @@ clean-cache:
 # plugins build targets
 #
 
-$(BIN_PATH)/logger: $(wildcard plugins/logger/*.go)
+$(BIN_PATH)/%: plugins/%/*
 	$(Q)echo "Building $@..."; \
 	cd $(dir $<) && $(GO_BUILD) -o $@ .
 
-$(BIN_PATH)/device-injector: $(wildcard plugins/device-injector/*.go)
-	$(Q)echo "Building $@..."; \
-	cd $(dir $<) && $(GO_BUILD) -o $@ .
-
-$(BIN_PATH)/hook-injector: $(wildcard plugins/hook-injector/*.go)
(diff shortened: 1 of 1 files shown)


klihub (approved):

Makefile
```diff
@@ -98,47 +98,15 @@ clean-cache:
 # plugins build targets
 #
 
-$(BIN_PATH)/logger: $(wildcard plugins/logger/*.go)
+$(BIN_PATH)/%: plugins/%/
```
klihub:
```suggestion
$(BIN_PATH)/%: plugins/%/*.go
```

klihub:
> > I don't think that will happen with this change. I think the opposite will happen (plugins not recompiled even if the plugin sources are changed), since you removed all dependencies (the wildcard-calls were adding dependencies on the plugins' source itself).
> 
> No, works at least for me as intended. I can make it `$(BIN_PATH)/%: plugins/%/*` it that's easier to read

On OSX or Linux ? And with what version of `make` ? For me it does not work with `GNU Make 4.4.1` unless I make the above mentioned changes.

marquiz:
> On OSX or Linux ? And with what version of `make` ? For me it does not work with `GNU Make 4.4.1` unless I make the above mentioned changes.

I don't get it. I have GNU Make 4.4.1 (on Fedora 42) and works like a charm :thinking: Anyway, I added an asterisk if that makes us happy

Makefile
marquiz: We want to depend on all files, e.g. go.mod, not just *go. Changed to `$(BIN_PATH)/%: plugins/%/*`

marquiz:
ping @chrishenzie @mikebrow

PR #200, opened 2025-08-08
plugins: hard code plugin names

marquiz:
Drop the -name argument for plugins, except for the template plugin where the default is set to "template". Ensures consistent name accross environments.

An less invasive alternative would be to change the flag default values.

### DIFF
--- a/plugins/device-injector/device-injector.go
+++ b/plugins/device-injector/device-injector.go
@@ -413,10 +413,9 @@ func dump(args ...interface{}) {
 
 func main() {
 	var (
-		pluginName string
-		pluginIdx  string
-		opts       []stub.Option
-		err        error
+		pluginIdx string
+		opts      []stub.Option
+		err       error
 	)
 
 	log = logrus.StandardLogger()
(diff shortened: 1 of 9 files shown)


marquiz:
@klihub @mikebrow @saschagrunert @chrishenzie

marquiz:
> Cannot we simply omit giving any name ? In that case the stub would use `filepath.Base(os.Args[0]))`, which would be the same what we are hardcoding here. We'd then also need to fix the image generating Dockerfile to not expect all plugins to be called `/bin/plugin` within the image...

IMO this PR is far more simpler (and simplification). We cannot fix the dockerfile. Either need to have separate dockerfile for each plugin, separate docker target/stage for each plugin in the dockerfile or some ugly scripting (dockerfile mangling) in the CI.

Why is it so bad to give the name explicitly (like the differ plugin already does)?

marquiz:
> We could add `ENV NRI_PLUGIN_NAME=${PLUGIN}` to the final image in the Dockerfile and omit setting the plugin name here. The stub defaults to taking it from the env first, then falling back to the binary name otherwise if nothing is given.

Ah, ok, we can do that.

> > Why is it so bad to give the name explicitly (like the differ plugin already does)?
> 
> It is not a biggie and of course we can go with what you suggest. I just thought that it'd be nice to avoid hardcoding and rely on the defaults instead...

We already have `opts = append(opts, stub.WithPluginName("...` in every plugin so I thought just riding that wave (with some simplification). TBH, I don't quite get the name business here `/bin/device-injector -name logger`. I get it for the stub, but not for the plugins.

klihub:
Closing. As discussed above, let's not do this.

PR #202, opened 2025-08-11
stub: allow overriding plugin name with Option

marquiz:
Use the given name instead of erroring out. The command line flag takes precedence over the NRI_PLUGIN_NAME environment variable.


marquiz:
> it would be better to not allow/do this for 'pre-launched' plugins (optional configuration picking/relaying uses the name to pick configuration for the plugin). The easiest way to achieve that is to do an `os.Getenv(api.PluginSocketEnvVar) == ""` check here and otherwise either error out, or give a warning/error but not override the name.

I see the problem. But can we come up with any less kludge-ish way to solve this? What if we just drop the problematic `WithPluginName` option? I still remain my comment in #200 that the naming craziness makes no sense to me.

mikebrow (approved):
LGTM

PR #203, opened 2025-08-11
plugins: drop -name flag from most of the plugins

marquiz:
Drop the -name command line flag from most of the sample plugins. Exceptions are the template plugin and the ulimit-adjuster plugin where the name is inserted into the log context.

Also, the differ plugin now takes it name the standard way, dropping the WithPluginName() opt.

Slightly different approach to #200


klihub (approved):
LGTM.

chrishenzie (approved):

PR #205, opened 2025-08-12
plugins/hook-injector: correct the hook configs

marquiz:
In the OCI hook configureation args[0] will not be passed to the hook script as an arg (by convention args[0] is the command).


marquiz:
@klihub @chrishenzie @mikebrow @samuelkarp

klihub (approved):

chrishenzie (approved):

PR #206, opened 2025-08-13
Bump dependencies

marquiz:
Update outdated deps.


marquiz:
Good point. I changed the main module to bump to the lowest non-CVE-versions. I think plugins can be updated without this limitation...

klihub:
> Good point. I changed the main module to bump to the lowest non-CVE-versions. I think plugins can be updated without this limitation...

Yes, definitely. It only applies to core NRI. Plugins can be freely updated to the latest versions.

PR #210, opened 2025-08-19
extract memory and CPU resource helpers

ErikJiang:
Refactored resource processing logic by extracting `adjustMemoryResource` and `adjustCPUResource` helper functions to eliminate code duplication.

- Added `adjustMemoryResource` function for memory resource operations
- Added `adjustCPUResource` function for CPU resource operations  
- Updated `adjustResources` and `updateResources` to use new helpers

### DIFF
--- a/pkg/adaptation/result.go
+++ b/pkg/adaptation/result.go
@@ -622,6 +622,159 @@ func (r *result) adjustHooks(hooks *Hooks, plugin string) error {
 	return nil
 }
 
+func (r *result) adjustMemoryResource(mem, targetContainer, targetReply *LinuxMemory, id, plugin string) error {
+	if mem == nil {
+		return nil
+	}
+
+	if v := mem.GetLimit(); v != nil {
+		if err := r.owners.ClaimMemLimit(id, plugin); err != nil {
+			return err
+		}
+		targetContainer.Limit = Int64(v.GetValue())
+		if targetReply != nil {
+			targetReply.Limit = Int64(v.GetValue())
+		}
+	}
+	if v := mem.GetReservation(); v != nil {
+		if err := r.owners.ClaimMemReservation(id, plugin); err != nil {
+			return err
+		}
+		targetContainer.Reservation = Int64(v.GetValue())
+		if targetReply != nil {
+			targetReply.Reservation = Int64(v.GetValue())
+		}
+	}
+	if v := mem.GetSwap(); v != nil {
+		if err := r.owners.ClaimMemSwapLimit(id, plugin); err != nil {
+			return err
+		}
+		targetContainer.Swap = Int64(v.GetValue())
+		if targetReply != nil {
+			targetReply.Swap = Int64(v.GetValue())
+		}
+	}
+	if v := mem.GetKernel(); v != nil {
+		if err := r.owners.ClaimMemKernelLimit(id, plugin); err != nil {
+			return err
+		}
+		targetContainer.Kernel = Int64(v.GetValue())
+		if targetReply != nil {
+			targetReply.Kernel = Int64(v.GetValue())
+		}
+	}
+	if v := mem.GetKernelTcp(); v != nil {
+		if err := r.owners.ClaimMemTCPLimit(id, plugin); err != nil {
+			return err
+		}
+		targetContainer.KernelTcp = Int64(v.GetValue())
+		if targetReply != nil {
+			targetReply.KernelTcp = Int64(v.GetValue())
+		}
+	}
+	if v := mem.GetSwappiness(); v != nil {
+		if err := r.owners.ClaimMemSwappiness(id, plugin); err != nil {
+			return err
+		}
+		targetContainer.Swappiness = UInt64(v.GetValue())
+		if targetReply != nil {
+			targetReply.Swappiness = UInt64(v.GetValue())
+		}
+	}
+	if v := mem.GetDisableOomKiller(); v != nil {
+		if err := r.owners.ClaimMemDisableOomKiller(id, plugin); err != nil {
+			return err
+		}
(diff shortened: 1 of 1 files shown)


pkg/adaptation/result.go
```diff
@@ -622,123 +622,144 @@ func (r *result) adjustHooks(hooks *Hooks, plugin string) error {
 	return nil
 }
 
-func (r *result) adjustResources(resources *LinuxResources, plugin string) error {
-	if resources == nil {
+func (r *result) processMemoryResource(mem *LinuxMemory, targetContainer, targetReply *LinuxMemory, id, plugin string) error {
```
chrishenzie: nit: I think I'm partial to the previous "adjust" method naming, which implies it's making adjustments to the result container and reply

chrishenzie: nit: `mem` is also the same type as the following two arguments. Should we remove this redundant `LinuxMemory`?

pkg/adaptation/result.go
```diff
@@ -886,99 +907,12 @@ func (r *result) updateResources(reply, u *ContainerUpdate, plugin string) error
...
-			}
-			resources.Memory.DisableOomKiller = Bool(v.GetValue())
-		}
-		if v := mem.GetUseHierarchy(); v != nil {
-			if err := r.owners.ClaimMemUseHierarchy(id, plugin); err != nil {
-				return err
-			}
-			resources.Memory.UseHierarchy = Bool(v.GetValue())
-		}
+	if err := r.processMemoryResource(u.Linux.Resources.Memory, resources.Memory, &LinuxMemory{}, id, plugin); err != nil {
```
chrishenzie: Without context, feeding in an empty struct that's unused here feels like an anti-pattern. What if we were to split up `processMemoryResource()` so that it only updates `targetContainer`, and then separately we can deep copy the results to `reply`? Thoughts on this approach?

ErikJiang: Thanks for the review! 
I've addressed your suggestions in the latest commit. 
Please let me know if there are any further adjustments needed. 😊

pkg/adaptation/result.go
```diff
@@ -622,124 +623,132 @@ func (r *result) adjustHooks(hooks *Hooks, plugin string) error {
...
+
+	if err := r.adjustMemoryResource(resources.Memory, container.Memory, id, plugin); err != nil {
+		return err
 	}
+	reply.Memory = proto.Clone(container.Memory).(*LinuxMemory)
+
+	if err := r.adjustCPUResource(resources.Cpu, container.Cpu, id, plugin); err != nil {
+		return err
+	}
+	reply.Cpu = proto.Clone(container.Cpu).(*LinuxCPU)
```
klihub: @ErikJiang I think these blind deep copies are not a good idea. Please don't do these. They are altering behavior compared to the original implementation, even if that does not show currently up in the end result.

The original implementation deliberately collected only the requested changes in the reply adjustment. With these changes now if there is any change requested to any of the CPU or memory resources of a container which has non-empty resources, the reply will incorrectly claim that the plugin requested changing every single parameter with the unrequested ones being copies of the original ones. You can see this for yourself easily, if you change the tests so that we don't start with a container with empty resources, for instance using this commit:

```diff
commit 38c50ea6bc12092ee6b9fe8cf53c09ca6a96884c (HEAD)
Author: Krisztian Litkey <krisztian.litkey@intel.com>
Date:   Tue Sep 23 14:57:52 2025 +0300

    WiP: adaptation: demo altered semantics of collected resource adjustments.
    
    Signed-off-by: Krisztian Litkey <krisztian.litkey@intel.com>

diff --git a/pkg/adaptation/adaptation_suite_test.go b/pkg/adaptation/adaptation_suite_test.go
index 4a19c7a..e093ded 100644
--- a/pkg/adaptation/adaptation_suite_test.go
+++ b/pkg/adaptation/adaptation_suite_test.go
@@ -628,6 +628,19 @@ var _ = Describe("Plugin container creation adjustments", func() {
                                                        "argument",
                                                        "list",
                                                },
+                                               Linux: &api.LinuxContainer{
+                                                       Resources: &api.LinuxResources{
+                                                               Cpu: &api.LinuxCPU{
+                                                                       Shares: api.UInt64(111),
+                                                                       Quota:  api.Int64(222),
+                                                                       Period: api.UInt64(333),
+                                                               },
+                                                               Memory: &api.LinuxMemory{
+                                                                       Limit: api.Int64(11111),
+                                                                       Swap:  api.Int64(22222),
+                                                               },
+                                                       },
+                                               },
                                        }
                                )
```

`make test` (or alternatively `ginkgo run ./pkg/adaptation`) should now fail with this extra commit and the changes of this PR in place, while it passes without this PR.

I am pretty sure this is not intentional on your part. It is just a side effect of your latest changes (in response to a previous review comment) where you now proto.Clone()-deep-copy resources from the adjusted container in the request, instead of properly collecting them in the reply. Even if this would not have unintended side-effects on the final adjusted container itself now, it might have in the future if we change semantics, and it already might cause false positives now with custom validators and break stuff.

Please switch your PR back to your original approach where this aspect was not changed.

pkg/adaptation/result.go
klihub: Those `proto.Clone()`s are not a good idea (see my other, more detailed comments). Please revert to your original approach where you properly collected the requested changes only in the reply adjustment.

ErikJiang:
@klihub @chrishenzie Thanks for the feedback. I've removed the `proto.Clone()` and reverted to selectively updating the reply to avoid the side effects you pointed out. The new commit is up for review.

chrishenzie (approved):
LGTM, thanks!

klihub (approved):
@ErikJiang LGTM. Thank you !

PR #211, opened 2025-08-19
Add dockerized target for building proto files

marquiz:
Makes it possible to build api files without contaminating the host system.

### DIFF
--- a/hack/Dockerfile.buildproto
+++ b/hack/Dockerfile.buildproto
@@ -1,0 +1,36 @@
+#   Copyright The containerd Authors.
+
+#   Licensed under the Apache License, Version 2.0 (the "License");
+#   you may not use this file except in compliance with the License.
+#   You may obtain a copy of the License at
+
+#       http://www.apache.org/licenses/LICENSE-2.0
+
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+
+ARG GO_VERSION=1.24
+ARG BASE_DEBIAN_DISTRO="bookworm"
+
+FROM golang:${GO_VERSION}-${BASE_DEBIAN_DISTRO} AS builder
+ARG ARTIFACTS
+
+WORKDIR /go/src
+
+RUN apt-get update && apt-get install -y unzip
+
+RUN --mount=type=cache,target=/go/pkg/mod/ \
+    --mount=src=.,target=. \
+    make install-protoc-dependencies install-ttrpc-plugin install-wasm-plugin install-protoc
+
+RUN --mount=type=cache,target=/go/pkg/mod/ \
+    --mount=src=.,target=.,rw=true \
+    make build-proto && \
+    tar czf /artifacts.tgz ${ARTIFACTS}
+
+FROM scratch AS final
+
+COPY --from=builder /artifacts.tgz .
--- a/Makefile
+++ b/Makefile
@@ -79,6 +79,13 @@ FORCE:
 
 build-proto: $(PROTO_GOFILES)
 
+.PHONY: build-proto-dockerized
+build-proto-dockerized:
+	$(Q)docker build --build-arg ARTIFACTS="$(dir $(PROTO_GOFILES))" --target final \
+		--output type=local,dest=$(RESOLVED_PWD) \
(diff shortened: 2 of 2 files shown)


hack/Dockerfile.buildproto
```diff
@@ -0,0 +1,36 @@
...
+#   limitations under the License.
+
+ARG GO_VERSION=1.24
+
+FROM golang:${GO_VERSION}-bullseye AS builder
+ARG ARTIFACTS
+
+WORKDIR /go/src
+
+COPY . .
```
thaJeztah: As these are only needed during the build, but currently cause the `apt-get` below to always be invalidated, perhaps the source should be mounted as well?

hack/Dockerfile.buildproto
```diff
@@ -0,0 +1,36 @@
...
+
+RUN apt-get update && apt-get install -y unzip
+
+RUN --mount=type=cache,target=/go/pkg/mod/ \
+    make install-protoc-dependencies install-ttrpc-plugin install-wasm-plugin install-protoc
+
+RUN --mount=type=cache,target=/go/pkg/mod/ \
+    make build-proto
+
+RUN tar czf /artifacts.tgz ${ARTIFACTS}
```
thaJeztah: i.e., these both could use a mount for the source. One thing to account for is that (unfortunately) the `make proto-build` defaults to writing to the source location, so it needs the build-context to be built writable (and combined with the `tar` step, as writable mounts are ephemeral);

```console
> [builder 5/6] RUN --mount=type=cache,target=/go/pkg/mod/     --mount=src=.,target=.     make build-proto:
0.317 Generating pkg/api/api.pb.go...
0.425 ./pkg/api/api.pb.go: Read-only file systemmake: *** [Makefile:181: pkg/api/api.pb.go] Error 1
```

Wondering if it'd be possible to make the proto-build accept a destination, which could possibly also avoid having to tar all the files to overwrite them (not just the generated files).



```suggestion
RUN --mount=type=cache,target=/go/pkg/mod/ \
    --mount=src=.,target=. \
    make install-protoc-dependencies install-ttrpc-plugin install-wasm-plugin install-protoc

RUN --mount=type=cache,target=/go/pkg/mod/ \
    --mount=src=.,target=.,rw=true \
    make build-proto && \
    tar czf /artifacts.tgz ${ARTIFACTS}
```

hack/Dockerfile.buildproto
```diff
@@ -0,0 +1,36 @@
...
+
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+
+ARG GO_VERSION=1.24
+
+FROM golang:${GO_VERSION}-bullseye AS builder
```
thaJeztah: This is also needed for the existing Dockerfile; we probably should have a separate build-arg for the debian variant (bullseye may no longer be provided now that trixie is out). Can be something similar to https://github.com/moby/moby/blob/b8897d91db618217dc10a6bf808ed3cf60c35839/Dockerfile#L3-L5

```
ARG GO_VERSION=1.24.6
ARG BASE_DEBIAN_DISTRO="bookworm"
ARG GOLANG_IMAGE="golang:${GO_VERSION}-${BASE_DEBIAN_DISTRO}"
```

hack/Dockerfile.buildproto
thaJeztah: But maybe @crazy-max has some tricks up his sleeve; I recall some of the BuildKit "generated files" dockerfiles also synchronise files (but possibly in a more granular way).

hack/Dockerfile.buildproto
marquiz: Ach :man_facepalming:  Absolutely this way

hack/Dockerfile.buildproto
marquiz: Thanks @thaJeztah :+1:  I was just quickly hacking together with too little thought. Changed as suggested. So much cleaner and better this way.

We can improve later if there are more tricks to apply

hack/Dockerfile.buildproto
marquiz: Sure :+1: Changed as suggested (copied the solution from the reference)

marquiz:
@klihub @thaJeztah @chrishenzie

klihub (approved):

mikebrow (approved):
LGTM

PR #212, opened 2025-08-21
pkg/api: add OptionalRepeatedString type

marquiz:


marquiz:

Pre-work for some of my upcoming work on managing the `linux.intelRdt` object of the runtime-spec. I decided to submit this one as a separate PR with the hopes of easier and more focused review.

I'm not sure if we need all the pointer's-pointer (or nil's-nil) stuff or would it be sufficient to return `[]string` (instead of `*[]string`) from `Get()`

pkg/api/optional_test.go
```diff
@@ -0,0 +1,89 @@
...
+			result := RepeatedString(tc.input)
+			if tc.expectNil {
+				assert.Nil(t, result)
+			} else {
+				assert.NotNil(t, result)
+			}
+
+			assert.Equal(t, tc.expectedValue, result.Get())
+		})
+	}
```
mikebrow, line 88: probably want a test that ensures the cloned result.

marquiz: There's an assert for this there (if I understand your comment)

klihub (approved):
LGTM

pkg/api/optional_test.go
mikebrow: If memory serves I wasn't sure in the reading if we needed a deep clone vs using the default golang slices clone and/or wasn't sure if assert Equal is implementing deep equals...  I see the later is true.  Not sure about the former.

> The slices.Clone function in Go creates a shallow copy of a slice, meaning it copies the elements but retains references to the same underlying data. This allows you to modify the cloned slice without affecting the original slice, as long as the elements themselves are not mutable types.

To ensure it's actually a deep clone one could modify result value(s) and ensure the original has not been modified.. this to remove incorrect future optimizations.

my concern may not be applicable in this situation.. :-)

marquiz:
> Is the goal of this wrapper type to distinguish between a `nil` (unset) list and an empty list?

Exactly. There are similar "optional" wrappers for other types. This follows suite

> If that distinction isn't critical for the `intelRdt` use case, could we simplify the API by using a standard `repeated string` field and treating an empty list as unset?

For now, it's not critical. But I want to be able to make the difference whether the list was specified or not. Make sure that we're not caught off guard if the empty list get's a special meaning in the runtime-spec in the future.

chrishenzie (approved):

klihub:
> LGTM nit some doc explanation on the involved funcs explaining that the wrapper is a shallow clone..
> 
> > The slices.Clone function in Go creates a shallow copy of a slice, meaning it copies the elements but retains references to the same underlying data. This allows you to modify the cloned slice without affecting the original slice, as long as the elements themselves are not mutable types.

@mikebrow Hmm..., okay. I read the code again, and if you mean `slices.Clone()` in the `RepeatedString(v interface{})` 'constructor', then I think we always end up shallow-cloning a `[]string` in the end, so there is no data left shared there. And the other one in `Get() *[]string` does the same with a more straightforward control flow.

mikebrow:
> > LGTM nit some doc explanation on the involved funcs explaining that the wrapper is a shallow clone..
> > > The slices.Clone function in Go creates a shallow copy of a slice, meaning it copies the elements but retains references to the same underlying data. This allows you to modify the cloned slice without affecting the original slice, as long as the elements themselves are not mutable types.
> 
> @mikebrow Hmm..., okay. I read the code again, and if you mean `slices.Clone()` in the `RepeatedString(v interface{})` 'constructor', then I think we always end up shallow-cloning a `[]string` in the end, so there is no data left shared there. And the other one in `Get() *[]string` does the same with a more straightforward control flow.

nod and strings are immutable in golang.. so even if the caller wanted to change one of those strings it would not be allowed.. feels odd.. but ok, guess I'm just missing C :-)

PR #215, opened 2025-08-26
Support direct editing of the intelRdt config

marquiz:
This PR adds support for directly adjusting the [`intelRdt`](https://github.com/opencontainers/runtime-spec/blob/main/config-linux.md#intelrdt) object of the container config.

Previously, we have the concept of RDT class, which provides indirect manipulation of the `intelRdt.closID` field. There the higher level runtime (containerd, cri-o) registers a resolver function which the NRI adaptation uses to translate the provided "RDT class" to actual closID (which is basically the resctrl group in the Linux resctrl filesystem).

The API extensions in this PR provides direct control of the relevant fields of the `intelRdt` object. The PR also includes a sample plugin to test/demonstrate the functionality (including deployment files and CI integration to build/publish images)

**REFS:** 
- https://github.com/opencontainers/runtime-spec/pull/1230
- https://github.com/opencontainers/runtime-spec/pull/1287

**NOTES:**
- ~~Depends/stacks on #212. That could be merged separately~~
- The "pkg/adaptation: support new RDT fields" commit borrows some snippets from #118.
- Known runtime implementations are runc (upcoming v1.4) and crun (v1.24+)
  - `closID`: supported by runc and crun
  - `schemata`: supported by runc  (upcoming v1.4) and crun (v1.22+)
    - runc PR: https://github.com/opencontainers/runc/pull/4830
    - crun PR: https://github.com/containers/crun/pull/1748
   - `enableMonitoring`  supported by runc  (upcoming v1.4) but not by crun (v1.24+)
     - runc PR: https://github.com/opencontainers/runc/pull/4832)
     - crun PR: https://github.com/containers/crun/pull/1866

**OPEN QUESTION/TODO:**

In this PR there is no mechanism for removing the `intelRdt` object from the config. That is probably a desirable feature.

```release-note
Support direct adjustment of the intelRdt container configuration
```


go.mod
```diff
@@ -9,7 +9,7 @@ require (
 	github.com/moby/sys/mountinfo v0.6.2
 	github.com/onsi/ginkgo/v2 v2.19.1
 	github.com/onsi/gomega v1.34.0
-	github.com/opencontainers/runtime-spec v1.1.0
+	github.com/opencontainers/runtime-spec v1.2.2-0.20250818071321-383cadbf08c0
```
klihub: @marquiz Do we know what is the expected timeframe for cutting the first such runtime-spec release which would put 383cadbf08c0 commit behind a tag ?

I am asking because we have been reluctant lately to pull in untagged versions of the runtime spec, because that would cause extra hassle/work if we wanted to or were forced to cut a new release and bump containerd and cri-o to pull in that new version. For instance #157 and #166 are current drafts exactly for this reason: the runtime-spec they require is not available yet in a tagged release.

pkg/api/update.go
```diff
@@ -16,7 +16,7 @@
 
 package api
 
-//nolint
+// nolint
```
klihub: nit: Is this intentional ? Shouldn't this be `//nolint` ?

pkg/api/adjustment.go
```diff
@@ -378,3 +396,9 @@ func (a *ContainerAdjustment) initLinuxResourcesUnified() {
 		a.Linux.Resources.Unified = make(map[string]string)
 	}
 }
+func (a *ContainerAdjustment) initLinuxRdt() {
```
klihub, line 447: Just a note: #118 has a bunch of queued similar changes, updating the result-handling code so that there is no need to always a priory create a fully constructed but empty adjustment. It would be nice to get that finally in, so I'll rebase it that on the latest main/HEAD then update with any missing init*() bits necessary.

pkg/api/update.go
marquiz: Yes. I didn't even notice this change :roll_eyes:, was automatic by gofmt. I now changed this to more appropriate format.

go.mod
marquiz: Yeah, sure. I don't think we need to hurry with this.

pkg/api/adjustment.go
marquiz: :+1: I'll review when that's updated. And then rebase when that's merged

marquiz:
I just recalled that there's actually one aspect/feature missing from the new extensions: ability to remove the `linux.intelRdt` from the container config (set it to nil/null). Suggestions for how to implement this are warmly welcome.

I couldn't come up with any nice solutions, I played with two alternatives but didn't commit either of them:
1. Have a separate field (like `drop`) in the NRI API
2. Have a well documented, otherwise invalid value for the `closID` field in the NRI API (e.g. `-/`) that causes the config to be dropped

giuseppe (approved):
LGTM

klihub:
> I just recalled that there's actually one aspect/feature missing from the new extensions: ability to remove the `linux.intelRdt` from the container config (set it to nil/null). Suggestions for how to implement this are warmly welcome.
> 
> I couldn't come up with any nice solutions, I played with two alternatives but didn't commit either of them:
> 
> 1. Have a separate field (like `drop`) in the NRI API
> 2. Have a well documented, otherwise invalid value for the `closID` field in the NRI API (e.g. `-/`) that causes the config to be dropped

@marquiz I'd vote with a mild bias for 2 above (because it's how we handle other similar cases where removal is supported). But with the addition that on the wire-protocol it's convention we use, but in the user-facing API we provide explicit functions for clearing any LinuxRdt,  with something along the lines of this:

```golang
func (a *ContainerAdjustment) RemoveLinuxRDT() {
    a.initLinux()
    a.Linux.Rdt = &LinuxRdt{
        ClosId = MarkForRemoval("") // or MarkForRemoval("/") if we want to keep
    }
}
```

marquiz:
@klihub thanks for the feedback. The MarkForRemoval thingie was a good hint, I didn't have that in my early prototype. I'll cook up something along the lines you suggested and then we can mull over that.

go.mod
klihub: @marquiz @mikebrow @chrishenzie In principle there is one way how we could get this partially in already now, to avoid having to keep rebasing this with a guaranteed conflict whenever we add something to LinuxContainer or LinuxContainerAdjustment.

It would be to split this up to 2 stacked PRs:
- PR#1: The bulk of this PR, without the OCI Spec bump + a single change to Generator.Adjust(): return an error if adjust.GetLinux().GetRdt() != nil
- draft PR#2: All the commits in PR#1 + the OCI Spec bump (eventually to a tagged version) + a Generator.Adjust() updated to do RDT adjustment if requested

Then we could merge #PR1 (and rebase PR#2 when that happens), and keep the OCI Spec-dependent bits pending until the necessary bits get tagged. We could apply the same to #157 and #166...

Maybe not worth the effort though, and we can just mark this a draft until we get a tagged OCI Spec.

marquiz:
> I'd vote with a mild bias for 2 above (because it's how we handle other similar cases where removal is supported). But with the addition that on the wire-protocol this is the convention we use, but in the user-facing API we provide explicit functions for clearing any LinuxRdt,

I ended up with adding a separate field. I implemented both variants and the "prefix closid with `-/`" variant was an ugly fragile mess tbh. So much cleaner to just have a separate field for this purpose. For envs and other guys it's a bit different when we're updating a map or slice and you selectively remove some items. But here we're managing the whole object. (if interested, I have the still lying around at https://github.com/marquiz/nri/commits/devel/rdt-remove-2/)

klihub:
> @klihub thanks for the feedback. The MarkForRemoval thingie was a good hint, I didn't have that in my early prototype. I'll cook up something along the lines you suggested and then we can mull over that.

@marquiz Sorry about the long sidetrack in our offline discussion about Remove*(). I was totally wrong and in the wrong mental model associated with most of the existing bits that support remove. We have a single RDT field (not a slice) so you were absolutely right that we can't handle removal with a MarkForRemoval inline notation, because then it would not be possible to indicate a removal prior to an addition in order to avoiding conflict with an earlier plugin. I don't know what the hell I was thinking... or I think I wasn't actually thinking at all.

klihub (approved):

marquiz:
Rebased, marked ready-for-review. Stacked on top of #212 
@klihub @mikebrow @chrishenzie

klihub:
As a purely theoretical question related to your notes in the commit log of  'api: add rdt'.

In the (I am adraid highly unlikely) situation where runc/crun/OCI compatible runtimes would gain the ability to update RDT for a running container, what would be the best way for us to model/handle that. Would we add a corresponding `&LinuxRdt{}` to `LinuxContainerUpdate{}` ?

marquiz:
> In the (I am adraid highly unlikely) situation where runc/crun/OCI compatible runtimes would gain the ability to update RDT for a running container, what would be the best way for us to model/handle that. Would we add a corresponding `&LinuxRdt{}` to `LinuxContainerUpdate{}`

Yes, I think that would likely be the place (with the usual reservations of predicting the future, how this would be implemented in runc).

marquiz:
@klihub @mikebrow @chrishenzie it would be nice to get #118 in before this. I'll then rebase

marquiz:
> @marquiz We could also do this the other way around. I think we're more eager to get this in. And I wouldn't mind at all that merge order forcing me to do another rebase of #118,

OK, thanks @klihub. I now rebased this PR.

I added you as Co-author in the `pkg/adaptation: support new RDT fields` patch where I borrowed some snippets from #118.

pkg/api/api.proto
```diff
@@ -606,6 +608,14 @@ message ContainerEviction {
   string reason = 2;
 }
 
+message LinuxRdt {
+  OptionalString clos_id = 1;
+  OptionalRepeatedString schemata = 2;
+  OptionalBool enable_monitoring = 3;
+  // NRI specific field to mark the RDT config for removal.
+  bool remove = 4;
```
chrishenzie, line 616: Why is this required? Is setting an empty message not equivalent to removing the config?

pkg/adaptation/result.go
```diff
@@ -499,6 +503,42 @@ func (r *result) adjustSysctl(sysctl map[string]string, plugin string) error {
...
+		return nil
+	}
+
+	r.initAdjustRdt()
+
+	id := r.request.create.Container.Id
+
+	if rdt.GetRemove() {
+		r.owners.ClearRdt(id, plugin)
+		r.reply.adjust.Linux.Rdt = &LinuxRdt{}
```
chrishenzie: Does `Remove` need to be set on this as well in order for it to be removed from the OCI spec?

pkg/api/strip.go
```diff
@@ -191,9 +195,6 @@ func (r *LinuxResources) Strip() *LinuxResources {
 	if r.BlockioClass != nil {
 		empty = false
 	}
-	if r.RdtClass != nil {
```
chrishenzie, line 194: Since this is still part of `LinuxResources` and supported via `AdjustRdtClass`, shouldn't we still include this in order to prevent `LinuxResources` from being stripped if it only contains `RdtClass`?

pkg/api/owners.go
```diff
@@ -773,6 +819,14 @@ func (f *FieldOwners) RdtClassOwner() (string, bool) {
 	return f.simpleOwner(Field_RdtClass.Key())
 }
```
chrishenzie, line 821: I think `RdtClosIDOwner()` is missing. Do we want to add this for consistency with owners.go?

pkg/runtime-tools/generate/generate.go
```diff
@@ -388,6 +390,42 @@ func (g *Generator) AdjustRdtClass(rdtClass *string) error {
 	return nil
 }
 
+// AdjustRdt adjusts the intelRdt object in the OCI Spec.
+func (g *Generator) AdjustRdt(r *nri.LinuxRdt) {
+	if r == nil {
+		return
+	}
+
+	if r.Remove {
```
chrishenzie, line 399: Related to above comment, the logic looks right here, but I'm wondering if this depends on `Remove` being set in the result.

I'm also wondering if it's feasible to not have a `Remove` flag and treat an empty `LinuxRdt` like a removal. One example (although slightly different cause it's a string) I could find was for block IO:
https://github.com/containerd/nri/blob/dc0990d81d5ef3de226faf6c99d048128bff4efc/pkg/runtime-tools/generate/generate.go#L357-L360

pkg/api/api.proto
marquiz: No, empty != nil. As per the runtime spec, if the struct is defined, the runtime will configure rdt for the container (with defaults)

pkg/api/owners.go
klihub: I'd say we do want it as it would be very useful for finer-grained validation if/once we manage to kick an eventually accepted version of that (with reliable plugin identification/authentication) forward. But I think it's already there: https://github.com/containerd/nri/blob/082b99b5b60a6273b0845b83853cfeb7e507a05a/pkg/api/owners.go#L339

pkg/adaptation/result.go
marquiz: Thanks for sharp-eyed review @chrishenzie. Indeed, it needs to be set so that if the original oci spec (before any nri plugins kicked in) contains RDT configuration, the final apply/adjust will then remove it. Fixed.

pkg/api/strip.go
marquiz: Hmm, yes, thanks for spotting this. Must've some collateral from an early development version where I ditched the RdtClass field... Fixed.

pkg/api/owners.go
marquiz: Yes, there is `RdtClosIDOwner()` right below.

pkg/runtime-tools/generate/generate.go
marquiz: I don't think it works because empty config means "I don't want to adjust anything". Also, e.g. ClosId  nil is different than empty string.

pkg/api/owners.go
chrishenzie: Oops, thanks for pointing that out

chrishenzie (approved):

klihub:
@mikebrow Mike, how about this one ? I labelled this v.next originally, before it got reviewed. But this is pretty self-contained and should be safe to merge, so I'd be okay with pushing this in already for v0.11.0.

mikebrow (approved):
LTGM
Well done!

PR #217, opened 2025-08-26
chore: clean and unify nolint directives

marquiz:
Drop unnecessary nolint directives and unify the style of remaining ones to match the rest of the codebase (and the golangci-lint documentation).

### DIFF
--- a/pkg/adaptation/api.go
+++ b/pkg/adaptation/api.go
@@ -25,7 +25,6 @@ import (
 //
 
 // Aliased request/response/event types for api/api.proto.
-// nolint
 type (
 	RegisterPluginRequest    = api.RegisterPluginRequest
 	RegisterPluginResponse   = api.Empty
@@ -102,7 +101,8 @@ type (
 )
 
 // Aliased consts for api/api.proto.
-// nolint
+//
+//nolint:revive // ignore const naming from auto-generated code
 const (
 	Event_UNKNOWN                       = api.Event_UNKNOWN
 	Event_RUN_POD_SANDBOX               = api.Event_RUN_POD_SANDBOX
@@ -134,7 +134,6 @@ const (
 )
 
 // Aliased types for api/optional.go.
-// nolint
 type (
 	OptionalString   = api.OptionalString
 	OptionalInt      = api.OptionalInt
@@ -147,7 +146,6 @@ type (
 )
 
 // Aliased functions for api/optional.go.
-// nolint
 var (
 	String   = api.String
 	Int      = api.Int
@@ -160,7 +158,6 @@ var (
 )
 
 // Aliased functions for api/types.go.
-// nolint
 var (
 	FromOCIMounts          = api.FromOCIMounts
 	FromOCIHooks           = api.FromOCIHooks
--- a/pkg/api/event.go
+++ b/pkg/api/event.go
@@ -26,7 +26,6 @@ const (
 	ValidEvents = EventMask((1 << (Event_LAST - 1)) - 1)
 )
 
-// nolint
 type (
 	// Define *Request/*Response type aliases for *Event/Empty pairs.
 
--- a/pkg/net/multiplex/mux.go
+++ b/pkg/net/multiplex/mux.go
@@ -363,7 +363,6 @@ func (m *mux) setError(err error) {
 	})
 }
 
-// nolint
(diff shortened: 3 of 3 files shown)


marquiz:
@klihub @chrishenzie @saschagrunert @mikebrow

mikebrow (commented):
interesting nolint by itself on it's own line should be a no op, right? 

should probably just delete them

marquiz:
> interesting nolint by itself on it's own line should be a no op, right?

It's not. It's like nolint:all (AFAIU).

pkg/adaptation/api.go
```diff
@@ -25,7 +25,8 @@ import (
 //
 
 // Aliased request/response/event types for api/api.proto.
-// nolint
+//
+//nolint:all
```
saschagrunert: can we actgually mention the linters which complain here? Like
```suggestion
//nolint:linter1,linter2 // [explanation why we don't lint]
```

klihub: @marquiz @saschagrunert Should we than have here `//nolint:revive // [ignore warnings about generated types, names, etc.]` here ?

marquiz: @saschagrunert agree, I'm all in for this, i.e. only list the linters that actually raise some errors. The original spirit of the PR was only to change the style, not modify any behavior. If we start to do that I'd suggest to remove the unnecessary nolint directives (most of the ones touched here). Thoughts @saschagrunert @klihub ?

saschagrunert: Removing them if they're not required sounds fine to me.

pkg/api/event.go
```diff
@@ -26,7 +26,7 @@ const (
 	ValidEvents = EventMask((1 << (Event_LAST - 1)) - 1)
 )
 
-// nolint
+//nolint:all
```
klihub: This could be removed as well. At least I don't get a failure if I remove it.

pkg/adaptation/api.go
klihub: > @saschagrunert agree, I'm all in for this, i.e. only list the linters that actually raise some errors. The original spirit of the PR was only to change the style, not modify any behavior. If we start to do that I'd suggest to remove the unnecessary nolint directives (most of the ones touched here). Thoughts @saschagrunert @klihub ?

@marquiz True, I did not realize that only the one for the `{Event,ContainerState}_`-block is needed in `pkg/adaptation/api.go`. And it can be a selective one for `revive`.

pkg/adaptation/api.go
```diff
@@ -160,7 +164,8 @@ var (
 )
 
 // Aliased functions for api/types.go.
-// nolint
+//
+//nolint:all
```
klihub: This can be removed.

pkg/adaptation/api.go
```diff
@@ -134,7 +136,8 @@ const (
 )
 
 // Aliased types for api/optional.go.
-// nolint
+//
+//nolint:all
```
klihub: This can be removed.

pkg/adaptation/api.go
```diff
@@ -102,7 +103,8 @@ type (
 )
 
 // Aliased consts for api/api.proto.
-// nolint
+//
+//nolint:all
```
klihub: This one we need to keep as a selective one for `revive`... something like `//nolint:revive // [ignore warnings about generated types, names, etc.]`

pkg/adaptation/api.go
marquiz: Thanks for the feedback. I'll adjust the PR accordingly

marquiz: Done

pkg/api/event.go
marquiz: Removed

pkg/adaptation/api.go
marquiz: Removed

pkg/adaptation/api.go
marquiz: Yes, kept this one

pkg/adaptation/api.go
marquiz: Removed

klihub (approved):

PR #218, opened 2025-08-28
nit: Add .idea folder to gitignore

ngopalak-redhat:
A small nit, but helps a lot when working with goland editor.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
+.idea
 build/


klihub (approved):

fuweid (approved):

.gitignore
```diff
@@ -1 +1,2 @@
+.idea
```
thaJeztah, line 1: I'd rather not have these in the upstream projects. These are user-specific files to ignore, and it will be a sliding slope to add all ignore rules for all possible editors that users may use.

These generally belong either in your global `.gitignore`, or can be added to the `.git/info/exclude` of your checkout.

Perhaps we should add a comment similar to https://github.com/moby/moby/blob/3bee048cbb32f63e45c98210fb15596a0a5e2e62/.gitignore#L1-L3


```
# If you want to ignore files created by your editor/tools, please consider a
# [global .gitignore](https://help.github.com/articles/ignoring-files).
```

PR #219, opened 2025-08-29
Update readme about nri.sock and SELinux

ngopalak-redhat:
In environments like OpenShift, its required to configure SELinux in security context. Hence added a small readme update.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ request and provide a response, both as JSON data.
 
 Plugins in NRI are daemon-like entities. A single instance of a plugin is
 now responsible for handling the full stream of NRI events and requests. A
-unix-domain socket is used as the transport for communication. Instead of
+unix-domain socket (`/var/run/nri/nri.sock`) is used as the transport for communication. Instead of
 JSON requests and responses NRI is defined as a formal, protobuf-based
 'NRI plugin protocol' which is compiled into ttRPC bindings. This should
 result in improved communication efficiency with lower per-message overhead,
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -491,6 +491,22 @@ access to these sockets and can act as NRI or Device Plugins. See the
 and [best practices](https://kubernetes.io/docs/setup/best-practices/enforcing-pod-security-standards/)
 about Kubernetes security.
 
+To use the plugins in SELinux-enabled environments, either create a new policy
```
mikebrow, line 494:
```suggestion
One expected path for running NRI plugins is to run them as a pod/container in a daemonset on each of the nodes of a cluster. 

### SELinux enabled environments
NOTE: To run the plugins, as a pod, in `SELinux-enabled` environments the kubernetes security level assigned to the pod MUST
```

pod security policies have been deprecated .. they've become levels assigned or via controller.. it's confusing :-)

README.md
```diff
@@ -491,6 +491,22 @@ access to these sockets and can act as NRI or Device Plugins. See the
...
+or set the SELinux type to spc_t (Super Privileged Container) in the pod's security
```
mikebrow, line 495:
```suggestion
set the SELinux type label to spc_t (Super Privileged Container) in the pod's security
```

mikebrow (commented):
suggest adding a couple sentences that describes the expected error that will get generated and what the problem is... volume mounting a host level socket between the pod and the container runtime.. 

This way seems like a big hammer.. is there any other way to make it work?  selinux mount label maybe?

PR #221, opened 2025-08-29
.gitignore: revert hastily reviewed editor-specific addition.

klihub:
Revert recent editor-specific addition to .gitignore, and add a comment about favoring global gitignore for such settings.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,4 @@
+# If you want to ignore files created by your editor/tools, please consider a
+# [global .gitignore](https://help.github.com/articles/ignoring-files).
+
 build/


ngopalak-redhat:
/lgtm
I didn't know about the global ignore. Thanks.

thaJeztah (approved):
Thank you! Sorry for being nit-picky on this. There's nothing critically _wrong_ with having those ignores, but it can become a bit of a sliding slope and adding extra maintenance burden; and as there's a good alternative for these, it's probably good to prevent that.


LGTM <3

fuweid:
I tried to create repo in github with default Go gitignore.

```
# If you prefer the allow list template instead of the deny list, see community template:
# https://github.com/github/gitignore/blob/main/community/Golang/Go.AllowList.gitignore
#
# Binaries for programs and plugins
*.exe
*.exe~
*.dll
*.so
*.dylib

# Test binary, built with `go test -c`
*.test

# Code coverage profiles and other test artifacts
*.out
coverage.*
*.coverprofile
profile.cov

# Dependency directories (remove the comment below to include it)
# vendor/

# Go workspace file
go.work
go.work.sum

# env file
.env

# Editor/IDE
# .idea/
# .vscode/
```

I saw there is commented `.idea/`. So I merged that original pull request.
I was against such pull request until I saw default gitignore - https://github.com/containerd/containerd/pull/7002
Anyway, revert looks good to me.

fuweid (approved):

PR #222, opened 2025-08-29
pkg/runtime-tools/generate: remove embedded field "Generator"

marquiz:
Makes linters happier (part of golangci-lint v2 related fixes).


klihub:
@marquiz These look fine, but seeing that you have a bunch of these piling up, all of them being basically a bunch of oneliners, and with the common goal of making the repo pass golangci-lint/v2 validation, I think it would save scarce review bandwidth to have all of them (as separate commits) in a single PR. Then you can stack the switch to golangci-lint/v2 as the topmost commit there to put the proof in the pudding.

So could you please close them and reopen as a single separate PR ?

PR #225, opened 2025-08-29
ci: bump golangci-lint to v2.4

marquiz:
Fixes and silences (nolint) linter errors and bumps golangci-lint version to the latest one.

Includes/depends on #222, #223 and #224

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -88,9 +88,9 @@ jobs:
         run: |
           echo "${{ github.workspace }}/bin" >> $GITHUB_PATH
 
-      - uses: golangci/golangci-lint-action@v6
+      - uses: golangci/golangci-lint-action@v8
         with:
-          version: v1.64.8
+          version: v2.4
 
   tests:
     name: Tests
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,44 +1,45 @@
+version: "2"
+
 linters:
   enable:
     - staticcheck
     - unconvert
-    - gofmt
-    - goimports
     - ineffassign
-    - vet
+    - govet
     - unused
     - misspell
     - revive
   disable:
     - errcheck
+  exclusions:
+    rules:
+      # We have protoc-generated ttRPC/gRPC code. When adding extra functions
+      # for generated types we want to consistently violate golint's semantic
+      # function name spelling rules, instead of inconsistently doing so only
+      # from automatically generated files. These rules are for that.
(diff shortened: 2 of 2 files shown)


marquiz:
@klihub @thaJeztah @saschagrunert @mikebrow @chrishenzie

pkg/runtime-tools/generate/generate.go
```diff
@@ -504,15 +504,15 @@ func (g *Generator) AdjustMounts(mounts []*nri.Mount) error {
 
 // sortMounts sorts the mounts in the generated OCI Spec.
 func (g *Generator) sortMounts() {
-	mounts := g.Generator.Mounts()
-	g.Generator.ClearMounts()
+	mounts := g.Mounts()
+	g.ClearMounts()
```
thaJeztah, line 509: If it's OK, I'd like to disable the `QF1008` in golangci-lint config. We've had some discussions around that quickfix in other projects, and think it's too opinionated to be applied as a "rule". In some cases it's good to abstract away that a type is embedded, but there's quite some cases where being explicit can be preferable, so it may be better to look at it "per-case" and not have the linter dictate that embedded types must be referenced implicitly.

(I have not looked in depth in which category this change itself should fall though)

https://github.com/moby/sys/blob/9dc3a90b82699c54554635f76a6a49757e02c523/.golangci.yml#L13-L19

mikebrow (approved):
LGTM

pkg/runtime-tools/generate/generate.go
marquiz: Thanks @thaJeztah. Very valid point. I think in this particular case here dropping the embedded field makes sense.

I'd suggest to submit a separate PR about disabling the check.

@klihub et al wdyt?

klihub: > If it's OK, I'd like to disable the `QF1008` in golangci-lint config. We've had some discussions around that quickfix in other projects, and think it's too opinionated to be applied as a "rule". In some cases it's good to abstract away that a type is embedded, but there's quite some cases where being explicit can be preferable, so it may be better to look at it "per-case" and not have the linter dictate that embedded types must be referenced implicitly.

I am also fine with that. v2 feels IMO pretty opinionated even if you have put yourself repeatedly through the v1 wringer.

> I'd suggest to submit a separate PR about disabling the check.
>
> @klihub et al wdyt?

Yes, let's file a separate PR for that.

PR #226, opened 2025-08-29
golangci: disable QF1008 from staticcheck linter

marquiz:
Disable:

    QF1008: could remove embedded field "Generator" from selector

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -35,6 +35,9 @@ linters:
       - linters:
           - revive
         text: "package-comments:"
+      - linters:
+          - staticcheck
+        text: "QF1008:"
 
 formatters:
   enable:


klihub (approved):

.golangci.yml
```diff
@@ -35,6 +35,9 @@ linters:
       - linters:
           - revive
         text: "package-comments:"
+      - linters:
+          - staticcheck
+        text: "QF1008:"
```
thaJeztah, line 40: This can use the "settings" section instead of the string matching; see https://github.com/moby/sys/blob/9dc3a90b82699c54554635f76a6a49757e02c523/.golangci.yml#L14

marquiz: Guess what, I tried that. But it's not the same, all != default linters. You cannot state something like
- "defaults"
- "-QF1008"

So I chose the minimal approach, keep using the defaults (whatever they are) and exclude the one problematic

thaJeztah: Ah! Sigh..

thaJeztah (approved):
LGTM, thanks!

PR #229, opened 2025-09-11
Makefile: build proto files unconditionally

marquiz:
Ditch the makefile pattern rule which simply doesn't work in scenarios where both the source and the build targets are stored in the git repo (as git operations mangle the time stamps).

So much grey hair avoided when the stuff just force generates everything instead of leaving you wonder why the tools didn't work as expected.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,7 +36,6 @@ COVERAGE_PATH := $(BUILD_PATH)/coverage
 
 PROTOBUF_VERSION = 3.20.1
 PROTO_SOURCES = $(shell find pkg -name '*.proto' | grep -v /vendor/)
-PROTO_GOFILES = $(patsubst %.proto,%.pb.go,$(PROTO_SOURCES))
 PROTO_INCLUDE = -I $(PWD) -I$(PROTOC_PATH)/include
 PROTO_OPTIONS = --proto_path=. $(PROTO_INCLUDE) \
     --go_opt=paths=source_relative --go_out=. \
@@ -80,11 +79,15 @@ FORCE:
 # build targets
 #
 
-build-proto: check-protoc install-ttrpc-plugin install-wasm-plugin install-protoc-dependencies $(PROTO_GOFILES)
+build-proto: check-protoc install-ttrpc-plugin install-wasm-plugin install-protoc-dependencies
+	for src in $(PROTO_SOURCES); do \
+		$(PROTO_COMPILE) $$src; \
+	done
+	sed -i '1s;^;//go:build !wasip1\n\n;' pkg/api/api_ttrpc.pb.go
 
 .PHONY: build-proto-dockerized
 build-proto-dockerized:
-	$(Q)docker build --build-arg ARTIFACTS="$(dir $(PROTO_GOFILES))" --target final \
+	$(Q)docker build --build-arg ARTIFACTS="$(dir $(PROTO_SOURCES))" --target final \
 		--output type=local,dest=$(RESOLVED_PWD) \
 		-f hack/Dockerfile.buildproto .
 	$(Q)tar xf artifacts.tgz && rm -f artifacts.tgz
@@ -175,15 +178,6 @@ validate-repo-no-changes:
 		echo "Please make sure to commit all changes, including generated files."; \
 		exit 1; \
 	}
-
-#
-# proto generation targets
-#
-
-%.pb.go: %.proto
-	$(Q)echo "Generating $@..."; \
-	$(PROTO_COMPILE) $<
-	sed -i '1s;^;//go:build !wasip1\n\n;' pkg/api/api_ttrpc.pb.go
 
 #
 # targets for installing dependencies
--- a/Makefile
+++ b/Makefile
@@ -193,7 +193,7 @@ check-protoc:
 	fi
 
 install-protoc install-protobuf:
-	$(Q)PROTOBUF_VERSION=$(PROTOBUF_VERSION) ./scripts/install-protobuf
+	$(Q)PROTOBUF_VERSION=$(PROTOBUF_VERSION) INSTALL_DIR=$(PROTOC_PATH) ./scripts/install-protobuf
 
 clean-protoc:
 	$(Q)rm -rf $(PROTOC_PATH)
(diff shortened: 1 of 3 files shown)


marquiz:
ping @klihub

klihub:
@marquiz Would you be okay with a milder version of this, for instance something like this: https://github.com/klihub/nri/tree/devel/forced-proto-rebuild ? That allows forcing rebuild of protobufs and it always forces it in the dockerized rebuild, so in your workflow it will always rebuild.

marquiz:
I think that would work. TBH, I just don't get the the insistence on these pattern rules that fundamentally don't work in this setup (when both the sources and targets are stored in git). I kind of get that from a philosophical "this is the way that make was designed to be used" pov but if it doesn't work why bother. The amount of wasted build time should certainly not be an issue, I'd guess the rule doesn't get run very often even globally.

klihub:
> I think that would work. TBH, I just don't get the the insistence on these pattern rules that fundamentally don't work in this setup (when both the sources and targets are stored in git).

Both the sources and targets are stored in git, but they are supposed to be consistently committed: if you change the protobuf definition, you should recompile and commit the compiled versions. Now I understand that right after a fresh clone the timestamps might be off, but that only should cause IMO (and in my experience) unnecessary recompilation and not the other way around (since the protobuf definition and the compiled protobuf files should always be consistently committed in the repo). I have seen this happening occasionally in CI. But I don't rememeber ever running into a situation in my local workflow, where I would have changed the protobuf definition, yet it would not have been recompiled.

> I kind of get that from a philosophical "this is the way that make was designed to be used" pov but if it doesn't work why bother. The amount of wasted build time should certainly not be an issue, I'd guess the rule doesn't get run very often even globally.

@marquiz That's not my concern. I'm more wary of folks having random different and newer versions of protoc& al. installed natively instead of the (fairly old) version we use (in sync with containerd). This then guarantees with 100% probability that a recompile will not result bit-by-bit in the same output, so it will result in false changes. If then folks rea than not extra careful and commit all their changes and file PRs from that, we need to point out that those part of the PR are not real changes so they should be reverted out of the PR. We have seen this already in the past even without a forced recompile by `make build`.

klihub:
> > Both the sources and targets are stored in git, but they are supposed to be consistently committed: if you change the protobuf definition, you should recompile and commit the compiled versions.
> 
> It doesn't matter how you commit then as git will tamper the timestamps. This is particularly annoying and breaks you when you have feature branches that you need to rebase (and e.g. happen to have conflicts).
> 
> > I'm more wary of folks having random different and newer versions of protoc& al. installed natively instead of the (fairly old) version we use (in sync with containerd).
> 
> I think this is not the correct way to try to prevent that. If this is a concern we should change all:/build: targets to depend on build-proto-containerized to get reproducible builds.

If you have a conflict in a compiled protobuf file and you try to resolve it manually, you are really doing something wrong. The right way to resolve such a conflict is to (resolve any conflicts you might have in the proto source), remove all generated pb.go files with conflicts, then run make and let it regenerate the missing compiled proto files (and all others that come from that same compilation).

klihub:
> I think this is not the correct way to try to prevent that. If this is a concern we should change all:/build: targets to depend on build-proto-containerized to get reproducible builds.

@marquiz Well, this is also true. And if I am the only one who thinks that unconditionally always regenerating and recompiling everything in the local development workflow is not a problem / the best way to go, then I realize I should simply give up and accept it. And if we go with that then I agree that a dockerized proto-build is friendlier.

klihub:
> Rebased. #232 and #233 are now merged. We should now be safe in always (re-)building the proto files with correct version of the tooling.
> 
> @klihub @mikebrow @chrishenzie 

Can you add to this PR a commit which removes dockerized proto building ? It is not necessary any more after the recently merged PRs.

klihub (approved):

Makefile
```diff
@@ -80,14 +79,11 @@ FORCE:
...
 #
 
-build-proto: check-protoc install-ttrpc-plugin install-wasm-plugin install-protoc-dependencies $(PROTO_GOFILES)
-
-.PHONY: build-proto-dockerized
-build-proto-dockerized:
-	$(Q)docker build --build-arg ARTIFACTS="$(dir $(PROTO_GOFILES))" --target final \
-		--output type=local,dest=$(RESOLVED_PWD) \
-		-f hack/Dockerfile.buildproto .
-	$(Q)tar xf artifacts.tgz && rm -f artifacts.tgz
```
thaJeztah, line 90: Shouldn't we keep a target to allow building on non-Linux machines? Don't think the native target works on macOS currently: https://github.com/containerd/nri/blob/25d9391690a7158d851364ef011e1f56fd607a70/scripts/install-protobuf#L33-L51

marquiz: Mm, true. Maybe we should keep it. I just need to fix it (I think). WDYT @klihub

thaJeztah: Sorry, my comment was a bit short (it was late :joy:)

Yeah, these are tricky; I think the changes to install the tools _locally_ were at least are good; there's various projects that require you to either "have the tools installed" already, or "figure out how to install", which may result in having to install different (system-wide) versions of the tools, which is a big pain if you happen to be jumping between various such projects for contributions :sweat_smile:

Then again, depending on the situation, having a containerized option available is good as well, as sometimes the (version of) tools to install can differ between branches for the same project, so having an option to generate code without tools lingering around on your system ... is really great to have.

hack/Dockerfile.buildproto
```diff
@@ -23,12 +23,10 @@ WORKDIR /go/src
...
 
 RUN --mount=type=cache,target=/go/pkg/mod/ \
-    --mount=src=.,target=. \
-    make install-protoc-dependencies install-ttrpc-plugin install-wasm-plugin install-protoc
-
-RUN --mount=type=cache,target=/go/pkg/mod/ \
+    --mount=type=cache,target=/go/tools/,sharing=private \
     --mount=src=.,target=.,rw=true \
-    make build-proto && \
+    find /go/tools && \
```
klihub: nit: is this for debugging ?

marquiz: :see_no_evil: Fixed

klihub (approved):

PR #230, opened 2025-09-15
api: expose container user/group ID to plugins.

klihub:
Expose container user, group and supplemental group info to NRI plugins.

```release-note
Expose container user, group, and supplemental group IDs to plugins
```


tuminoid (approved):
Lovely, I was not aware it was as simple as exposing the User from the spec.

This is very helpful to us going forward, thank you!

chrishenzie (approved):

pkg/api/api.pb.go
```diff
@@ -2741,6 +2749,70 @@ func (x *CDIDevice) GetName() string {
 	return ""
 }
 
+// User and group IDs for the container.
+type User struct {
+	state         protoimpl.MessageState
+	sizeCache     protoimpl.SizeCache
+	unknownFields protoimpl.UnknownFields
```
mikebrow, line 2756: umask?

mikebrow:
> Lovely, I was not aware it was as simple as exposing the User from the spec.
> 
> This is very helpful to us going forward, thank you!

user placed into the container (oci) spec.. is carefully crafted :-)

mikebrow (approved):
LGTM

NIT wdyt about also adding umask, username?

is this the first "useful" linux vs windows fields area did we talk about that?

see examples here:  https://github.com/opencontainers/runtime-spec/blob/main/config.md#user
not saying I want LinuxUser vs WindowsUser.. just sayin it's worth a discussion.

klihub:
> LGTM
> 
> NIT wdyt about also adding umask, username?
> 

@mikebrow Username I was looking at, but since it is Windows only and we don't really have any Windows support at the moment, it was left out. If we start adding Windows support at some point, we can/should add it then.

Umask we can definitely add, now or later, if we find it useful for an external observer or just for the sake of completeness.

> is this the first "useful" linux vs windows fields area did we talk about that?
> see examples here: https://github.com/opencontainers/runtime-spec/blob/main/config.md#user not saying I want LinuxUser vs WindowsUser.. just sayin it's worth a discussion.

We have deliberately omitted anything windows-specific so far.

PR #231, opened 2025-09-23
adaptation: test with populated initial resources.

klihub:
The current tests for collecting container resource adjustments starts out with fully empty container memory and CPU resources. This allows unintended changes to sneak in unnoticed to the semantics of collected adjustments (see #210). Let's make this less likely by starting with populated initial resources.


mikebrow (approved):
LGTM

PR #232, opened 2025-09-26
Install protoc locally in the source tree

marquiz:
Install (and use) protoc and plugins under build/tools in the source tree in an attempt to ensure that the correct version of the tooling is always used.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,15 +12,6 @@
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
 
-PROTO_SOURCES = $(shell find . -name '*.proto' | grep -v /vendor/)
-PROTO_GOFILES = $(patsubst %.proto,%.pb.go,$(PROTO_SOURCES))
-PROTO_INCLUDE = -I$(PWD):/usr/local/include:/usr/include
-PROTO_OPTIONS = --proto_path=. $(PROTO_INCLUDE) \
-    --go_opt=paths=source_relative --go_out=. \
-    --go-ttrpc_opt=paths=source_relative --go-ttrpc_out=. \
-    --go-plugin_opt=paths=source_relative,disable_pb_gen=true --go-plugin_out=.
-PROTO_COMPILE = PATH=$(PATH):$(shell go env GOPATH)/bin; protoc $(PROTO_OPTIONS)
-
 GO_CMD     := go
 GO_BUILD   := $(GO_CMD) build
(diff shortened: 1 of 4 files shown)


marquiz:
@klihub

klihub:
@marquiz I think this would be good, especially if it helps us avoiding to force dockerized protobuilds on ourselves and everybody else in the local workflow.

One related question though. Since our `scripts/install-protobuf` is a verbatim copy of the one in the containerd main repo and we (try to semi-regularly) sync it with that, do you think it would be possible to contribute the same changes there, too ?

Makefile
```diff
@@ -39,8 +30,18 @@ GINKGO        := ginkgo
...
 COVERAGE_PATH := $(BUILD_PATH)/coverage
 
+PROTO_SOURCES = $(shell find pkg -name '*.proto' | grep -v /vendor/)
+PROTO_GOFILES = $(patsubst %.proto,%.pb.go,$(PROTO_SOURCES))
+PROTO_INCLUDE = -I $(PWD) -I$(TOOLS_PATH)/include
+PROTO_OPTIONS = --proto_path=. $(PROTO_INCLUDE) \
+    --go_opt=paths=source_relative --go_out=. \
+    --go-ttrpc_opt=paths=source_relative --go-ttrpc_out=. \
+    --go-plugin_opt=paths=source_relative,disable_pb_gen=true --go-plugin_out=.
+PROTO_COMPILE = PATH=$(TOOLS_PATH)/bin protoc $(PROTO_OPTIONS)
```
klihub, line 43:
```suggestion
PPROTO_COMPILE = PATH="$(TOOLS_PATH)/bin:$$PATH" protoc $(PROTO_OPTIONS)
```
Isn't this necessary ?

marquiz: No. At least for me it works fine without. And makes sure that no plugins are accidentally taken from the system

marquiz:
> One related question though. Since our `scripts/install-protobuf` is a verbatim copy of the one in the containerd main repo and we (try to semi-regularly) sync it with that, do you think it would be possible to contribute the same changes there, too ?

Yeah, sure, I can try :) As soon as I get this working (i.e. make the ci pass muster).

One gap still remains that we don't check the protoc (or plugins) versions installed in the source tree.

Makefile
klihub: Yes, was stupid user error.

klihub (approved):

PR #233, opened 2025-09-26
Ensure that we use correct version of protoc and its deps

marquiz:
Split into multiple self-contained commits. 

1. Use protoc-specific local install directory
    Makes it possible to cleanly and reliably wipe protoc and it's dependencies without affecting other tools that might get installed under the build/tools dir in the future.
1. Makefile: add check-protoc target
1. Simplify and align the installation of protoc dependencies
    Maintain the versions in go.mod.
1. Makefile: check that we use correct version of protoc and its deps
1. Update auto-generated code

**NOTE:** Could be submitted as multiple smaller PRs, too, that feels better


marquiz:
> do we really need to change the tool versions we install and use, do we

We don't change any tools versions, except for protoc-gen-go which we cannot avoid as now the version comes from go.mod where we already have google.golang.org/protobuf v1.34.1. You can see from the last commit that the effect this is ~nil

klihub:
> > do we really need to change the tool versions we install and use, do we
> 
> We don't change any tools versions, except for protoc-gen-go which we cannot avoid as now the version comes from go.mod where we already have google.golang.org/protobuf v1.34.1. You can see from the last commit that the effect this is ~nil

@marquiz Okay, so sorry for my stupid questions which I'm asking since I have never tried to manage toolchain dependencies using the `tool` directive in `go.mod`. I did a quick read up in the go mod documentation and I'm not sure if I understand this now correctly. But is it so, that if we go with this then we completely lose control over which version of these `go tool`-managed toolchains we can use. IOW, if there is a new release of any of these which breaks everything for us, then there is no way for us to simply force `go` / `go tool` to keep using the previous version which would still work for us ?

mikebrow (approved):
LGTM nit.. squash :-)

klihub (approved):

PR #234, opened 2025-09-30
adaptation: ensure sync'ed plugins are fully registered in tests.

klihub:
A plugin having received a sync request is not enough to ensure it has been fully registered on the runtime side (IOW appended to the list of active plugins). When we are waiting for test plugins to start up, use an extra plugin synchronization block/unblock to ensure they are fully registered.

### DIFF
--- a/pkg/adaptation/suite_test.go
+++ b/pkg/adaptation/suite_test.go
@@ -125,6 +125,7 @@ func (s *Suite) WaitForPluginsToSync(plugins ...*mockPlugin) {
 	for _, plugin := range plugins {
 		Expect(plugin.Wait(PluginSynchronized, timeout)).To(Succeed())
 	}
+	s.runtime.runtime.BlockPluginSync().Unblock() // ensure plugins are fully registered
 }
 
 // Cleanup the test suite.


klihub:
> Just curious, how did we discover this issue?

@chrishenzie In the [v1beta1 API RFC/draft PR](#235) I added a [plugin shutdown test](https://github.com/containerd/nri/blob/fdb63299484b2e139840422779237330ee7fee3d/pkg/adaptation/adaptation_suite_test.go#L3191) which was failing in a weird way. That got me chasing this.

The reason for the failure was that we were waiting for plugin startup (and assumed registration) by synchronizing on the plugin getting a Synchronize request. Normally this is enough, especially if the test is doing mocked pod or container creation because those are blocking/unblocking plugin registration, so they implicitly do this synchronization.

But the test there instead immediately tried to shut down the plugin by name. Which was racy without this synchronization, so the plugin was practically never fully registered yet, so we did not find it on the adaptation side, and failed to shut it down and register the test shutdown reason. So the test failed.

klihub:
> Does the plugin loop need a new time.After per plugin? Using the same timeout channel for all the plugins seems sus.

@mikebrow Not sure if we really need it here. That timeout is to put an upper bound on how long we wait in case of failure. Semantically it is a failure if any plugin does not get synchronized within that time. So we don't want to give each separately that timeout.

mikebrow (approved):
LGTM

klihub:
> two seconds for all plugins to synch.. seemed short.

Yes, I understand and probably true in general. If we had a large number of real plugins to be started like this, and they all had to do some real processing during sync, then 2 seconds overall could be too short in some cases. But in this case, we only have a few of them at most at any time, and they are all almost no-op test plugins which merely record what they were sync'd with. So I think 2 seconds should be pretty generous here.

PR #235, opened 2025-10-01
[RFC] Add v1beta1 API revision, protocol translation for v1alpha1 plugins.

klihub:
The primary motivation for this PR is to clean up some annoying bits in the NRI wire protocol.

The most prominent such bit is the lack of proper, dedicated wire RPC calls and messages for some pod and container lifecycle events. These are instead multiplexed over a single RPC call and message (StateChange, StateChangeEvent), lacking any event-specific returned data and making any future extension unnecessarily difficult. Other, smaller annoyances are the usage of a single Empty response type in multiple other RPC calls, with similar negative side effects.

Addressing these alone is a more intrusive protocol change than the usual addition of new NRI inputs or controls and the related protocol extension with new fields. Therefore this PR defines a new v1beta1 revision of the protocol, initially identical to v1alpha1, which is then updated address the aforementioned annoyances.

Some of the original protocol problems are also reflected in the stub and the plugin/stub API. This PR tries to clean some of those up as well. It defines a new v1beta1 stub, turning the original one into a v1alpha1 revision.

In an attempt to provide a painless transition path for plugins (and to experiment with it in practice) this PR also implements an internal transparent protocol conversion from v1alpha1 to v1beta1, for plugins which register themselves talking the earlier version of the protocol.

This PR is still work-in-progress, but I'd like to gather some initial thoughts and general feedback from anybody who has the time and interest to take a look at this, already in its current state.

Here is the list of things (I think) this PR addresses in v1beta1:

- eliminate on wire StateChange in favor of dedicated event-specific requests
- eliminate remaining usage of Empty (plugin registration, plugin shutdown, wasm plugin log bridging)
- implement plugin shutdown, use it where it makes sense instead of just closing the connection
- delay closing plugin connections for fatal errors, giving error propagation back to plugins a chance

Here is a list of things I either know are still missing or I'd like to additionally address. This is probably still incomplete. I'll try to update it as I can.

- [ ] adaptation test suite tests for protocol conversion
- [ ] documentation updates
- [ ] additional stub/plugin API cleanups

@mikebrow @chrishenzie @marquiz If you have some BW, PTAL.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -113,4 +113,4 @@ jobs:
           echo "${{ github.workspace }}/bin" >> $GITHUB_PATH
 
       - run: |
-          make install-ginkgo test codecov
+          make test codecov
--- a/Makefile
+++ b/Makefile
@@ -25,7 +25,6 @@ GO_BUILD_FLAGS ?=
 GO_MODULES := $(shell $(GO_CMD) list ./...)
 
 GOLANG_CILINT := golangci-lint
-GINKGO        := ginkgo
 
 RESOLVED_PWD  := $(shell realpath $(shell pwd))
 BUILD_PATH    := $(RESOLVED_PWD)/build
@@ -71,7 +70,7 @@ clean: clean-plugins
 
 allclean: clean clean-cache
 
-test: test-gopkgs
+test: test-gopkgs test-plugins
 
 generate: generate-golang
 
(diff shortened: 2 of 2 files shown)


mikebrow:
Interesting changes.. let's do a call on this next week covering the motivations. I'm LGTM for a versioning practice run to manage proto version extensions/enhancements. Not sure about the need to split state change notifications into event type unique rpcs? Would like to hear some ideas on scenarios where that would be helpful. The other changes all sound like good tech debt cleanup.

PR #239, opened 2025-10-10
feat: Make logger a configurable struct member for stub

gauravkghildiyal:
Previously, the stub package used a global logger instance, making it difficult for applications using the stub as a library to customize logging behavior, by using `nrilog.Set()`: https://github.com/containerd/nri/blob/9b8befa666a390a66d48f3304d126b0fcb18e58a/pkg/log/log.go#L37-L40

### DIFF
--- a/pkg/stub/stub.go
+++ b/pkg/stub/stub.go
@@ -178,6 +178,9 @@ type Stub interface {
 	// This is the default timeout if the plugin has not been started or
 	// the timeout received in the Configure request otherwise.
 	RequestTimeout() time.Duration
+
+	// Logger returns the logger used by the stub.
+	Logger() nrilog.Logger
 }
 
 const (
@@ -188,9 +191,6 @@ const (
 )
 
 var (
-	// Logger for messages generated internally by the stub itself.
-	log = nrilog.Get()
-
 	// Used instead of a nil Context in logging.
 	noCtx = context.TODO()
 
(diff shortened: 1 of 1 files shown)


pkg/stub/stub.go
```diff
@@ -329,6 +327,7 @@ func New(p interface{}, opts ...Option) (Stub, error) {
 
 		registrationTimeout: DefaultRegistrationTimeout,
 		requestTimeout:      DefaultRequestTimeout,
+		logger:              nrilog.Get(),
```
klihub, line 341: @gauravkghildiyal I think it would also make sense to add a `WithLogger(nrilog.Logger) Option` option which would allow setting the used logger with a `stub.Option`.

klihub:
> > because sooner or later I ended up in a situation where I wanted/needed to log from a non-type-scoped function but had no logger
> 
> This is tempting, but it is better to resist the temptation. It makes the assumption that there's only going to be a single log output stream for the entire package in a program at any given time. This prevents running unit tests in parallel with per-test log output. I don't know about this stub that is being discussed here, but is it really safe to assume that no downstream consumer of the package has a need to run two instances at the same time in their program?

That's true and is a valid point. IOW... no, it wouldn't be safe to assume that there would be only ever a single instance in a single program. Plus, right now we only ever internally generate log messages from a type-instance context, never from a package global context. And if we ever need to start doing that, we can always add a functions for setting that logger, too.

pohly:
> And if we ever need to start doing that, we can always add a functions for setting that logger, too.

Or design those functions so that the caller can pass in a logger. It's more explicit that way.

This works best for functions which accept a `ctx context.Context`, but then you need to decide whether you want to use `logr.FromContext` (https://pkg.go.dev/github.com/go-logr/logr#FromContext) or `slog-context.FromContext` (https://pkg.go.dev/github.com/veqryn/slog-context#FromCtx) and the corresponding logger APIs.

klihub:
@chrishenzie PTAL if you have some free cycles.

samuelkarp:
https://github.com/containerd/log also uses a context-based approach, so you get `log.G(ctx).Info`. If NRI weren't shared with CRI-O that'd probably be the right package to use here...

samuelkarp (approved):

PR #242, opened 2025-11-04
update OCI runtime spec to 1.3.0

aojea:
This brings the new OCI runtime spec version 1.3.0 that brings some interesting new features like linux network devices

Ref: https://github.com/opencontainers/runtime-spec/releases/tag/v1.3.0


aojea:
oh, thanks I was also trying to untangle this https://github.com/opencontainers/runtime-tools/pull/796, let me hand over to you since you are much more familiar

Issue #165, opened 2025-04-24
[Proposal] Support single RPC operations per Pod


jipperinbham:
Currently, the NRI plugin relies on ttrpc for communication for containerd <-> plugin(s) which imposes a single RPC at a time via multiple `sync.Mutex` locks in `containerd` and `pkd/adaptation`. A [naive attempt was made](https://github.com/containerd/containerd/pull/11671) to remove the `sync.Mutex` in containerd thinking it was only to protect access to the `map` structure but it was highlighted it also was there for guaranteeing 1 RPC at a time with NRI.

What we're observing after onboarding more and more workloads to our new NRI based dataplane is the NRI Lock is causing higher than desired contention with Pod operations. 

NRI Lock P99
<img width="1284" alt="Image" src="https://github.com/user-attachments/assets/424e8559-f232-4fae-ae8e-51f2a3c245a5" />

From the above, it's clear the current limitation is going to be an issue for the team trying to manage Pod start SLIs. I can imagine it's not a straightforward change but we'd love to start the conversation.

tallclair:
I propose generalizing this idea a little, via Multiple Granularity Locking (MGL). During plugin registration, let the plugin declare a scope: Global, Pod, or Container (and maybe namespace?).

A plugin that uses Pod hooks must request at least the Pod scope. A plugin that modifies other pods must register the global scope. A plugin that only handles container operations could just use the container scope.

PR #244, opened 2025-11-11
adaptation: support concurrent RPCs via per Pod lock

jipperinbham:
For users who do not need to perform cross Pod updates, a new per Pod lock implementation can be selected via `WithPodLocking()` Option func. These changes would need to go along with a similar containerd change as https://github.com/containerd/containerd/pull/11671 to both expose the ability to select the per pod locker as well as remove the global lock on each nri RPC invocation in containerd.

Addresses #165


klihub:
> @klihub Any chance you can take a look at this and provide some feedback? 

I'll take a closer look.

Issue #240, opened 2025-10-15
Looking for collaborators / ideas on how to leverage an NRI plugin for the purpose of managing identity artifacts in containers/pods


atpugtihsrah:
The idea is about using an NRI plugin to manage setup of identity artifcats for container/pods. Oriented around Spiffe/Spire, the idea is to create and mount identity artifacts/certificate during container creation time. Instead of the application in the container creating and fetching its identify artificats, this NRI identity plugin would manage setting it up for the application/workload. An alternative way is to extend Envoy to manage identities for the application. An additional idea which I don't fully understand yet is to let applications access files outside the container by mounting a root hosted file.

This is from a rough discussion I had with @mikebrow 

Open Questions:

- Spiffe SVID certificates/documents are short lived. What should happen to the container / pod after the certificate expires?  
  - Restart or fail the pod/container.
  - Pause the container.
  - Do some tricks to update the mounted certificate/artifact dynamically. 

Some links from our discussions:

- https://github.com/spiffe/spire-plugin-sdk
- https://medium.com/kagenti-the-agentic-platform/identity-in-agentic-platforms-enabling-secure-least-privilege-access-996527f1c983
- https://github.com/kagenti/kagenti/tree/main/kagenti/examples/identity

Slack discussion thread: https://cloud-native.slack.com/archives/CGEQHPYF4/p1760517445810219

mikebrow:
Great points @klihub .. will need this to work with VM isolated containers, should be fine.. would be nice to use readonly in the container by default.    I really like setting up the bind mount first in create.. then in parallel updating the contents to meet that and make sure it's finished before start.  Might want the option of a readiness probe. Might also want to be able to tie this to pod lifecycle and make it available to selected containers in the pod.. with various sub-paths.

As for pausing... we do that today during a checkpoint is how criu works..  Nod to the point about k8s not understanding those additional states pod/container.  Back of my head it was more of a pause option while updating the contents with a new cert in  that the application running in the container could have a trigger on.

atpugtihsrah:
> will need this to work with VM isolated containers

Thanks, another good point!

grosskur:
FYI, another approach I've seen is to use a CSI driver to deliver SPIFFE SVIDs to pods:

https://github.com/cert-manager/csi-driver-spiffe

Just out of curiosity, are there specific properties that make the NRI approach preferable here? (I'm not super familiar with VM-isolated containers and whether they would work with CSI volumes.)

PR #245, opened 2025-11-11
Identity Plugin RFC

atpugtihsrah:
Identity Plugin RFC WIP

Relates to https://github.com/containerd/nri/issues/240

Implementation PR: https://github.com/containerd/nri/pull/288

### DIFF
--- a/docs/rfcs/0001-identity-plugin.md
+++ b/docs/rfcs/0001-identity-plugin.md
@@ -6,15 +6,17 @@
 
 # Summary
 
-The identity plugin, running as a daemonset, fetches identity artifacts for each annotated container and mounts the artifact volume into the container.
+The identity plugin, running as a daemonset, fetches identity artifacts (X.509 SVID certificate bundle) for each annotated container and mounts the artifact volume into the container.
 
 # Motivation
 
-The current way how we mount identity artifacts into a container is to use a initContainer / sidecar. There is a sidecar for each pod. This sidecar fetches the artifacts from identity providers, injects the volume containing identity artifacts into the pod and mounts the volume into the container. Using a sidecar in each pod for fetching identity artifacts is too resource intensive. Using a initContainer sidecar comes with an additional overhead of increased container start latency. The motivation for using a NRI Identity Plugin is that only one process is enough to fetch the identity artifacts for all of the annotated containers in all the pods on a single host. Therefore the identity PLUGIN as a daemonset can fetch the identity artifacts for all the containers in all the pods on all thee hosts in the cluster that have the appropriate annotations. Running as a single process per host instead of single process per pod we save on host resources.
+The current way how we mount identity artifacts (X.509 SVID certificate bundle) into a container is to use a initContainer / sidecar. There is a sidecar for each pod. This sidecar fetches the artifacts from identity providers, injects the volume containing identity artifacts into the pod and mounts the volume into the container. Using a sidecar in each pod for fetching identity artifacts is too resource intensive. Using a initContainer sidecar comes with an additional overhead of increased container start latency. The motivation for using a NRI Identity Plugin is that only one process is enough to fetch the identity artifacts for all of the annotated containers in all the pods on a single host. Therefore the identity Plugin as a daemonset can fetch the identity artifacts for all the containers in all the pods on all thee hosts in the cluster that have the appropriate annotations. Running as a single process per host instead of single process per pod we save on host resources.
 
 # Proposed Implementation
 
 ## Setup Steps
+
+WIP
 
 1. Create a static spiffe-id for the plugin.
 
@@ -45,15 +47,19 @@ agent {
 
 ## How the plugin will work
 
+WIP
+
 1. Plugin registers itself to the agent and get its spiffe id.
 
-2. Containerd calls the NRI plugin on CreateContainer event. The plugin receives the pod and container metadata. Using this metadata, the plugin executes step 3 below. The plugin has to ensure that the pid is a stable identifier, that the pid is not a recycled identifier. This is a platform dependent step. (e.g. by using pidfds on Linux). https://spiffe.io/docs/latest/deploying/spire_agent/#delegated-identity-api
+2. Containerd calls the NRI plugin on CreateContainer event - PENDING fix the correct event, add details on when and which volume is mounted. The plugin receives the pod and container metadata. Using this metadata, the plugin executes step 3 below. The plugin has to ensure that the pid is a stable identifier, that the pid is not a recycled identifier. This is a platform dependent step. (e.g. by using pidfds on Linux). https://spiffe.io/docs/latest/deploying/spire_agent/#delegated-identity-api
 
 3. The plugin reads the podspec and parses the annotations. Using spire-api-sdk, the plugin fetches identity artifacts and mounts the volume containing the identity artifacts into the container.  The spire-api-sdk streams artifact updates, the plugin will on receiving updates will overwrite the artifact on the host volume. Since the host volume is mounted into the container, the artifacts in the container are automatically updated. The plugin does not need a separate mount annotation in the podspec yaml. The daemonset specifies the volume that the plugin will use to fetch and download the identity artifacts. The plugin will then mount the host volume to the container volume using the cert_dir from the podspec.
 
 
 
 ## DaemonSet yaml
+
+WIP
 
 ```
 apiVersion: apps/v1
@@ -99,6 +105,11 @@ spec:
 
 
 ## Podspec example
+
+WIP PENDING - are annotations immutable? can we support mutable annotations?
+WIP PENDING - supporting pod certificate KEP 4317
+
+WIP
 
 ```
 apiVersion: v1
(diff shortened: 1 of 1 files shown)


docs/rfcs/0001-identity-plugin.md
```diff
@@ -0,0 +1,198 @@
...
+
+2. Containerd calls the NRI plugin on CreateContainer event. The plugin receives the pod and container metadata. Using this metadata, the plugin executes step 3 below. The plugin has to ensure that the pid is a stable identifier, that the pid is not a recycled identifier. This is a platform dependent step. (e.g. by using pidfds on Linux). https://spiffe.io/docs/latest/deploying/spire_agent/#delegated-identity-api
+
+3. The plugin reads the podspec and parses the annotations. Using spire-api-sdk, the plugin fetches identity artifacts and mounts the volume containing the identity artifacts into the container.  The spire-api-sdk streams artifact updates, the plugin will on receiving updates will overwrite the artifact on the host volume. Since the host volume is mounted into the container, the artifacts in the container are automatically updated. The plugin does not need a separate mount annotation in the podspec yaml. The daemonset specifies the volume that the plugin will use to fetch and download the identity artifacts. The plugin will then mount the host volume to the container volume using the cert_dir from the podspec.
```
aojea, line 76: I think this assumes the annotations can not be mutated after the pod has been created, right?

aojea: It will be also nice to use the native pod certificates feature https://github.com/kubernetes/enhancements/tree/master/keps/sig-auth/4317-pod-certificates instead of depending on annotations

atpugtihsrah: > I think this assumes the annotations can not be mutated after the pod has been created, right?

Thanks for the review comments. Correct, I am operating under the assumption that annotations will not be mutated after a Pod is created. Do you have a specific use case in mind where those annotations might be modified? 

Looking into the KEP.

atpugtihsrah: Thanks!! Makes sense to use PodCertificate Projected Volume Sources.
cc @mikebrow

aojea: > Correct, I am operating under the assumption that annotations will not be mutated after a Pod is created.

then I suggest you to make it explicit in the doc, people tend to do a lot of things and someone may get the idea of that being supported

docs/rfcs/0001-identity-plugin.md
```diff
@@ -0,0 +1,198 @@
...
+}
+
+```
+
+
+## How the plugin will work
+
+1. Plugin registers itself to the agent and get its spiffe id.
+
+2. Containerd calls the NRI plugin on CreateContainer event. The plugin receives the pod and container metadata. Using this metadata, the plugin executes step 3 below. The plugin has to ensure that the pid is a stable identifier, that the pid is not a recycled identifier. This is a platform dependent step. (e.g. by using pidfds on Linux). https://spiffe.io/docs/latest/deploying/spire_agent/#delegated-identity-api
```
klihub: Don't we have a phase error problem here if we already need to know the container's PID at this point ? You can't really obtain a container's PID before the container has been started. But that is only going to happen later, triggered by a `StartContainer` CRI call, for which you'll then get corresponding `{Start,PostStart}Container` NRI calls.

docs/rfcs/0001-identity-plugin.md
mikebrow: nod annotations should be a test platform only.. and would nec. be expected to be replaced with actual fields.. if feature based NRI plugins are to be formalized.  We could also check for pod updated annotations as we are adding  pod updates to NRI..

docs/rfcs/0001-identity-plugin.md
mikebrow: I believe we need to get the task pid which is created by the task create in start container before we call nri start container call.. also before the task.Start() is run.. and also before the nri poststart.. should be doable, no?  https://github.com/containerd/containerd/blob/main/internal/cri/server/container_start.go#L217-L251

klihub: I think in `StartContainer` we do pass that pid already to plugins.

atpugtihsrah: Ok I need to switch from CreateContainer to StartContainer. CreateContainer (or PostCreateContainer for that matter) has no task, therefore no task pid. 

> I think in StartContainer we do pass that pid already to plugins.

Almost. We have the pid in StartContainer but we don't update the container with the pid at the time of calling nri. Line 251 https://github.com/containerd/containerd/blob/main/internal/cri/server/container_start.go#L251 is where we already call nri but only later on line 263 https://github.com/containerd/containerd/blob/main/internal/cri/server/container_start.go#L263 is where we update the container status with the pid. I will look into how to make this change so that we have the pid when calling nri.

atpugtihsrah: Thanks for the feedback :)

atpugtihsrah: Ok NewTask() already has the Pid https://github.com/containerd/containerd/blob/main/client/container.go#L304. Digging in deeper.

klihub: > Ok I need to switch from CreateContainer to StartContainer. CreateContainer (or PostCreateContainer for that matter) has no task, therefore no task pid.

That's fine for auxiliary container-related actions which do not need to modify the container('s metadata/parameters) itself. But if there is a need, any container modifications you'll have to do during container creation. Once a container is created it is immutable apart from (compute) resources (controlled via cgroupfs). And I think also lately runtime-level annotations, but even if that's possible now via CRI, it's currently not via NRI after container creation.

> 
> > I think in StartContainer we do pass that pid already to plugins.
> 
> Almost. We have the pid in StartContainer but we don't update the container with the pid at the time of calling nri. Line 251 https://github.com/containerd/containerd/blob/main/internal/cri/server/container_start.go#L251 is where we already call nri but only later on line 263 https://github.com/containerd/containerd/blob/main/internal/cri/server/container_start.go#L263 is where we update the container status with the pid. I will look into how to make this change so that we have the pid when calling nri.

Hmm, but this is not what I see when I test this in practice, just using the nri logger plugin and a test pod.

```bash
$ cat ~/xfer/pods/pid-test.yaml 
apiVersion: v1
kind: Pod
metadata:
  name: pid-test
spec:
  containers:
  - name: c0
    image: busybox:latest
    command:
      - sh
      - -c
      - sleep 3600
    resources:
      requests:
        cpu: 750m
        memory: 100M
      limits:
        cpu: 750m
        memory: 100M
    imagePullPolicy: IfNotPresent
$ ~/xfer/nri/plugins/logger -idx 10 |& grep -E '="((RunPodSandbox)|(((Create)|(Start)|(PostStart))Container)):' | grep -E '(( pod)|( container)|( id)|( name)|( pid)):'

# Then creating the pod/container
$ kubectl apply -f ~/xfer/pods/pid-test.yaml

# gives these PIDs in the logger
time="2025-11-21T06:57:21Z" level=info msg="RunPodSandbox: pod:"
time="2025-11-21T06:57:21Z" level=info msg="RunPodSandbox:    id: 81e54b147dfbeeaf471e2205676d0f3cd89e330dc40390e836570103b0982f60"
time="2025-11-21T06:57:21Z" level=info msg="RunPodSandbox:    name: pid-test"
time="2025-11-21T06:57:21Z" level=info msg="RunPodSandbox:    pid: 703970"
time="2025-11-21T06:57:21Z" level=info msg="CreateContainer: pod:"
time="2025-11-21T06:57:21Z" level=info msg="CreateContainer:    id: 81e54b147dfbeeaf471e2205676d0f3cd89e330dc40390e836570103b0982f60"
time="2025-11-21T06:57:21Z" level=info msg="CreateContainer:    name: pid-test"
time="2025-11-21T06:57:21Z" level=info msg="CreateContainer:    pid: 703970"
time="2025-11-21T06:57:21Z" level=info msg="CreateContainer: container:"
time="2025-11-21T06:57:21Z" level=info msg="CreateContainer:    id: 2892bce0f96135bddd43c309b9b0efa3e72113aa8e4ad3c66559ebe537cd0f9b"
time="2025-11-21T06:57:21Z" level=info msg="CreateContainer:    name: c0"
time="2025-11-21T06:57:21Z" level=info msg="StartContainer: pod:"
time="2025-11-21T06:57:21Z" level=info msg="StartContainer:    id: 81e54b147dfbeeaf471e2205676d0f3cd89e330dc40390e836570103b0982f60"
time="2025-11-21T06:57:21Z" level=info msg="StartContainer:    name: pid-test"
time="2025-11-21T06:57:21Z" level=info msg="StartContainer:    pid: 703970"
time="2025-11-21T06:57:21Z" level=info msg="StartContainer: container:"
time="2025-11-21T06:57:21Z" level=info msg="StartContainer:    id: 2892bce0f96135bddd43c309b9b0efa3e72113aa8e4ad3c66559ebe537cd0f9b"
time="2025-11-21T06:57:21Z" level=info msg="StartContainer:    name: c0"
time="2025-11-21T06:57:21Z" level=info msg="StartContainer:    pid: 703995"
time="2025-11-21T06:57:21Z" level=info msg="PostStartContainer: pod:"
time="2025-11-21T06:57:21Z" level=info msg="PostStartContainer:    id: 81e54b147dfbeeaf471e2205676d0f3cd89e330dc40390e836570103b0982f60"
time="2025-11-21T06:57:21Z" level=info msg="PostStartContainer:    name: pid-test"
time="2025-11-21T06:57:21Z" level=info msg="PostStartContainer:    pid: 703970"
time="2025-11-21T06:57:21Z" level=info msg="PostStartContainer: container:"
time="2025-11-21T06:57:21Z" level=info msg="PostStartContainer:    id: 2892bce0f96135bddd43c309b9b0efa3e72113aa8e4ad3c66559ebe537cd0f9b"

# which are these processes
$ ps axuw | grep 703970 | grep -v grep
65535     703970  0.0  0.0   1020   712 ?        Ss   06:57   0:00 /pause
$ ps axuw | grep 703995 | grep -v grep
root      703995  0.0  0.0   4436  2212 ?        Ss   06:57   0:00 sleep 3600
```

So to me it looks like we do get the container's process's PID in StartContainer.

atpugtihsrah: Thanks Krisztian!! Also finally found the code where we set the pid https://github.com/containerd/containerd/blob/main/internal/cri/nri/nri_api_linux.go#L700. 

> That's fine for auxiliary container-related actions which do not need to modify the container('s metadata/parameters) itself. But if there is a need, any container modifications you'll have to do during container creation. Once a container is created it is immutable apart from (compute) resources (controlled via cgroupfs). And I think also lately runtime-level annotations, but even if that's possible now via CRI, it's currently not via NRI after container creation.

Do you mean we cannot mount any new volumes in container start (after container create)? Alternative could be we mount the volume in container create but only write the artifacts in container start when we have the pid.

atpugtihsrah: > time="2025-11-21T06:57:21Z" level=info msg="CreateContainer:    pid: 703970"

Oh there is also a pid in CreateContainer. Looking into this

mikebrow: > > time="2025-11-21T06:57:21Z" level=info msg="CreateContainer:    pid: 703970"
> 
> Oh there is also a pid in CreateContainer. Looking into this

the pid in the create container call is the pod's pid, which could probably be useful for namespaces shared over the pod to all containers..

samuelkarp:
Is there a particular reason this is opened against [containerd/nri](https://github.com/containerd/nri)?  From the description this (a) isn't core to the NRI framework, and (b) sounds complex enough that it's probably not really an example plugin like the others we have in the repo.  https://github.com/containers/nri-plugins could be a better home since that's a collection of NRI plugins, or this might be meaningful enough to live in its own repository?

docs/rfcs/0001-identity-plugin.md
klihub: > Thanks Krisztian!! Also finally found the code where we set the pid https://github.com/containerd/containerd/blob/main/internal/cri/nri/nri_api_linux.go#L700.
> 
> > That's fine for auxiliary container-related actions which do not need to modify the container('s metadata/parameters) itself. But if there is a need, any container modifications you'll have to do during container creation. Once a container is created it is immutable apart from (compute) resources (controlled via cgroupfs). And I think also lately runtime-level annotations, but even if that's possible now via CRI, it's currently not via NRI after container creation.
> 
> Do you mean we cannot mount any new volumes in container start (after container create)? 

Yes, you need to pre-mount a directory during container creation where you can then later write artifacts or attach new mounts as subdirectories.

> Alternative could be we mount the volume in container create but only write the artifacts in container start when we have the pid.

Yes, exactly like that.

kfox1111:
Might want to have a look at:
https://github.com/spiffe/spiffe/pull/376

PR #246, opened 2025-11-18
device-injector: always log injection summary.

klihub:
Always log device injection summary, also in verbose mode.

### DIFF
--- a/plugins/device-injector/device-injector.go
+++ b/plugins/device-injector/device-injector.go
@@ -132,9 +132,7 @@ func injectDevices(pod *api.PodSandbox, ctr *api.Container, a *api.ContainerAdju
 
 	for _, d := range devices {
 		a.AddDevice(d.toNRI())
-		if !verbose {
-			log.Infof("%s: injected device %q...", containerName(pod, ctr), d.Path)
-		}
+		log.Infof("%s: injected device %q...", containerName(pod, ctr), d.Path)
 	}
 
 	return nil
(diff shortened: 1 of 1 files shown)


plugins/device-injector/device-injector.go
```diff
@@ -122,7 +122,7 @@ func injectDevices(pod *api.PodSandbox, ctr *api.Container, a *api.ContainerAdju
 	}
 
 	if len(devices) == 0 {
-		log.Debugf("%s: no devices annotated...", containerName(pod, ctr))
+		log.Infof("%s: no devices annotated...", containerName(pod, ctr))
```
chrishenzie: Could this produce log spam? If so, I could see why we might keep this as debug. I think level=info should contain when something *does* happen as opposed to when it doesn't

klihub: @chrishenzie You'd get one extra logged message per container created. Whether that's log spam is a bit subjective. With me it usually goes so, that as long as everything works I wish the author chose to log less, but once I mistype and go goose-chasing some non-existing other bug, I eventually wish the author would have logged that as info instead of debug.

Anyway, I flipped those back to debug.

mikebrow (commented):
wdyt about:
if verbose .. log.Infof 
else log.Debugf

chrishenzie:
@mikebrow I would expect applications to always write info/warn/error logs and not condition that behind anything, but optionally (additionally) log debug logs. Maybe this would be simpler if instead of `verbose`, we just added a log-level flag.

PR #247, opened 2025-11-19
Drop dependency on opencontainers/runtime-tools

AkihiroSuda:
`pkg/runtime-tools/generate` no longer depends on `*github.com/opencontainers/runtime-tools/generate.Generator`, which has no release since 2019.

The underlying generator can be replaced with a custom implementation of the `UnderlyingGenerator` interface.


mikebrow (approved):
LGTM

PR #248, opened 2025-12-01
Add support for sysctl adjustment

samuelkarp:
Testing: unit tests including in the default validator.  I have not yet built a new containerd binary with these changes and built a plugin to set sysctls.  Let me know if you want to see that before merging.

Another note: right now the default validator configuration is `RejectSysctlAdjustment` and off by default, so either we can change that to a positive "AllowSysctlAdjustment" or we can turn it on by default in containerd when we integrate this change.

```release-note
Add support for sysctl adjustment via NRI
```


klihub:
> Testing: unit tests including in the default validator. I have not yet built a new containerd binary with these changes and built a plugin to set sysctls. Let me know if you want to see that before merging.

This looks straightforward enough to not warrant insisting on that. Then again, before a new NRI version gets eventually tagged with this included, we need to test this with the runtimes and make sure it works.

> Another note: right now the default validator configuration is `RejectSysctlAdjustment` and off by default, so either we can change that to a positive "AllowSysctlAdjustment" or we can turn it on by default in containerd when we integrate this change.

I think the current naming and semantics ('enabled by default via a reject-setting defaulting to false') is in line with how the other similar validator settings are configured. So unless there is some other reason for it, I don't think it'd need to be flipped the other way around.

klihub (approved):
LGTM.

chrishenzie (approved):

PR #251, opened 2025-12-08
pkg/api: adjustment stripping fixes.

klihub:
Add missing handling of recently introduced adjustment fields in Strip().

### DIFF
--- a/pkg/adaptation/adaptation_suite_test.go
+++ b/pkg/adaptation/adaptation_suite_test.go
@@ -529,11 +529,6 @@ var _ = Describe("Plugin container creation adjustments", func() {
 				Priority: 5,
 			})
 
-		case "clear I/O priority":
-			a.SetLinuxIOPriority(&nri.LinuxIOPriority{
-				Class: api.IOPrioClass_IOPRIO_CLASS_NONE,
-			})
-
 		case "linux net device":
 			if overwrite {
 				a.RemoveLinuxNetDevice("hostIf")
@@ -552,11 +547,6 @@ var _ = Describe("Plugin container creation adjustments", func() {
 				Flags: []api.LinuxSchedulerFlag{
 					api.LinuxSchedulerFlag_SCHED_FLAG_RESET_ON_FORK,
 				},
-			})
-
-		case "clear linux scheduler":
-			a.SetLinuxScheduler(&api.LinuxScheduler{
-				Policy: api.LinuxSchedulerPolicy_SCHED_NONE,
 			})
 
 		case "linux sysctl":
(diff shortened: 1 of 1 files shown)


pkg/api/strip.go
```diff
@@ -296,6 +320,129 @@ func (m *LinuxMemory) Strip() *LinuxMemory {
...
+	if s.Nice != 0 {
+		empty = false
+	}
+
+	if s.Priority != 0 {
+		empty = false
+	}
+
+	if len(s.Flags) == 0 {
+		s.Flags = nil
```
thaJeztah: Only had a quick glance, but if this is moved to the start, I _think_ we'd be able to use an early return instead of having to check each option for non-empty.

klihub: @thaJeztah Thanks for pointing that out. I rewrote all the newly added functions to avoid using an `empty` boolean, deal with any zero length map/slice reduction to nil first, and use an early return instead where we used to do an `empty = false`.

klihub: I'll take a look at making corresponding changes to the existing bits, but maybe in a separate PR.

pkg/api/strip_test.go
```diff
@@ -0,0 +1,329 @@
...
+
+package api_test
+
+import (
+	fmt "fmt"
+	"testing"
+
+	"github.com/containerd/nri/pkg/api"
+
+	faker "github.com/brianvoe/gofakeit/v7"
```
mikebrow, line 25: Let's see... license is MIT.. one owner.. a couple maintainer/committers.. seems to respond quickly to PRs.. seems fine.

faked random value testing.. semi-deterministic..  interesting shortcut.. 

Normally we have explicit tests, should we have both deterministic tests and also add these randomly generated values?

pkg/api/strip.go
```diff
@@ -112,6 +112,30 @@ func (l *LinuxContainerAdjustment) Strip() *LinuxContainerAdjustment {
...
+	}
+
+	if len(l.Sysctl) != 0 {
+		empty = false
+	}
+
+	if len(l.NetDevices) != 0 {
+		empty = false
+	}
+
```
mikebrow, line 144: these ^ additional strips LGTM

pkg/api/strip_test.go
```diff
@@ -0,0 +1,329 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+func TestContainerAdjustmentStrip(t *testing.T) {
```
samuelkarp: It's probably also worth having tests that validate Strip does successfully normalize zero-valued structs/maps/slices to nil.

pkg/api/strip_test.go
thaJeztah: Not sure if it's fully comparable, but in moby, we started using `pgregory.net/rapid` for property based tests; see https://github.com/moby/moby/pull/50519 (and the repository https://github.com/flyingmutant/rapid).

pkg/api/strip_test.go
klihub: @samuelkarp And I also pushed a fix for my seriously brain-dead idea of stripping *both* the collected and the expected results. It is a really bad idea. We should never Strip() the expected result, as it only masks all the various Strip()ping bugs we have/had in place where a non-empty adjustment is incorrectly reduced to nil, which then goes unnoticed if we Strip() both of them (incorrectly) as then they will compare equal.

klihub: @samuelkarp Having taken a look at your #259 I think the best course of action would be to extend/fix the tests added here to also verify valid reductions to nil, then cherry-pick from this PR all the other commits except the hand-fix to Strip() to #259, finalize and merge that.

Would it be okay with you, if I extend the tests here, then cherry-pick it and the other commits and push them to your #259 ?

samuelkarp: We can also merge this first, then I can rebase #259 on top of it.  That might be a cleaner history.

klihub: Yes, true. I think it'd make more sense that way around.

klihub: > It's probably also worth having tests that validate Strip does successfully normalize zero-valued structs/maps/slices to nil.

I added test cases for verifying reduction of empty adjustments to nil. I'm not sure if it is a comprehensive enough set.

PR #253, opened 2025-12-11
adaptation: allow compiling out WASM support altogether.

klihub:
This PR introduces the "nri_no_wasm" build tag which now allows disabling WASM support altogether at compile time. This can be helpful in resource constrained or other environments, where WASM support is not needed and the related binary size increase penalty is considered difficult to justify. 

```release-note
Add nri_no_wasm build tag to allow disabling WASM support at compile time
```

### DIFF
--- a/pkg/adaptation/adaptation.go
+++ b/pkg/adaptation/adaptation.go
@@ -32,8 +32,6 @@ import (
 	"github.com/containerd/nri/pkg/log"
 	validator "github.com/containerd/nri/plugins/default-validator/builtin"
 	"github.com/containerd/ttrpc"
-	"github.com/tetratelabs/wazero"
-	"github.com/tetratelabs/wazero/imports/wasi_snapshot_preview1"
 
 	"google.golang.org/protobuf/proto"
 )
@@ -80,6 +78,9 @@ type Adaptation struct {
 var (
 	// Used instead of nil Context in logging.
 	noCtx = context.TODO()
+
+	// ErrWasmDisabled is returned for WASM initialization if WASM support is disabled.
+	ErrWasmDisabled = errors.New("WASM support is disabled (at build time)")
 )
 
 // Option to apply to the NRI runtime.
@@ -155,23 +156,12 @@ func New(name, version string, syncFn SyncFn, updateFn UpdateFn, opts ...Option)
 		return nil, fmt.Errorf("failed to create NRI adaptation, nil UpdateFn")
 	}
 
-	wasmWithCloseOnContextDone := func(ctx context.Context) (wazero.Runtime, error) {
-		var (
-			cfg = wazero.NewRuntimeConfig().WithCloseOnContextDone(true)
-			r   = wazero.NewRuntimeWithConfig(ctx, cfg)
-		)
-		if _, err := wasi_snapshot_preview1.Instantiate(ctx, r); err != nil {
+	wasmService, err := getWasmService()
+	if err != nil {
+		log.Errorf(noCtx, "failed to initialize WASM support: %v", err)
+		if !errors.Is(err, ErrWasmDisabled) {
 			return nil, err
 		}
-		return r, nil
-	}
-
-	wasmPlugins, err := api.NewPluginPlugin(
-		context.Background(),
-		api.WazeroRuntime(wasmWithCloseOnContextDone),
-	)
-	if err != nil {
-		return nil, fmt.Errorf("unable to initialize WASM service: %w", err)
 	}
 
 	r := &Adaptation{
@@ -183,7 +173,7 @@ func New(name, version string, syncFn SyncFn, updateFn UpdateFn, opts ...Option)
 		dropinPath:  DefaultPluginConfigPath,
 		socketPath:  DefaultSocketPath,
 		syncLock:    sync.RWMutex{},
-		wasmService: wasmPlugins,
+		wasmService: wasmService,
 	}
 
 	for _, o := range opts {
--- a/pkg/adaptation/wasm-disabled.go
+++ b/pkg/adaptation/wasm-disabled.go
@@ -1,0 +1,27 @@
+//go:build nri_no_wasm
+
+/*
+   Copyright The containerd Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package adaptation
+
+import (
+	"github.com/containerd/nri/pkg/api"
+)
+
+func getWasmService() (*api.PluginPlugin, error) {
+	return nil, ErrWasmDisabled
+}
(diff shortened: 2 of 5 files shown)


pkg/adaptation/adaptation.go
```diff
@@ -183,7 +173,7 @@ func New(name, version string, syncFn SyncFn, updateFn UpdateFn, opts ...Option)
 		dropinPath:  DefaultPluginConfigPath,
 		socketPath:  DefaultSocketPath,
 		syncLock:    sync.RWMutex{},
-		wasmService: wasmPlugins,
+		wasmService: wasmService,
```
thaJeztah, line 176: I was originally considering if this could be a functional option (`WithWASMService`); that could potentially allow the caller to either pass, or not pass the option.

Compile-time could still be an option for that if a stub `WithWASMService` was created (that would either return an error, or `nil`).

pkg/adaptation/wasm-disabled.go
```diff
@@ -0,0 +1,27 @@
+//go:build nowasm
```
thaJeztah: Just thinking out loud here; build-tags cross all modules, so we should check if `nowasm` is potentially ambiguous (i.e., if it's already used elsewhere but for a different purpose, then setting the build-tag could cause side-effects).

If that's a problem indeed, we may need something more unique (`nri_no_wasm` or `nri_nowasm`)

thaJeztah:
Oh, thanks! @robmry PTAL 🤗 


The thing I was struggling with was that various of the WASM types are also deeply engrained in the API definition, but not sure if that caused the binary size to grow as well.

I started looking if we could somehow get that dependency out of the API definition, but didn't get _very_ far; just a quick "let's see if this could work", but there were a lot of places and additional definitions; I pushed what I had here, but it very likely won't even compile; 

- https://github.com/containerd/nri/pull/254


edit: wrong link

pkg/adaptation/wasm-disabled.go
klihub: Yes, I also think we should namespace our build tags. I updated it to `nri_nowasm`.

klihub:
> Oh, thanks! @robmry PTAL 🤗
> 
> The thing I was struggling with was that various of the WASM types are also deeply engrained in the API definition, but not sure if that caused the binary size to grow as well.

Yeah, I'm unable to tell. But I tried the effect of tis and compiling containerd main/HEAD now with BUILDTAGS=nri_nowasm results in on x86_64 in a 45272408 binary size vs. 48184952 when compiling with WASM support enabled. A size reduction of 2912544 bytes.

pkg/adaptation/wasm-disabled.go
thaJeztah, line 1: I _think_ using underscores instead of hyphens is more common in Go; perhaps `adaptation_wasm.go` and `adaptation_nowasm.go` could work?

thaJeztah (approved):
thanks!

I think this change looks good; AFAICS, it doesn't paint us in a corner on considering some of the other options (functional option to add the plugin and looking further at what can be done for the API definitions).

I left one nit for the naming of the files, but not really blocking, just a suggestion.

LGTM

pkg/adaptation/wasm-disabled.go
klihub: I think that calling it `adaptation_wasm.go` would not work. It would break the builds, because `wasm` is a valid GOARCH, so it would be excluded from x86_64 and arm builds.

thaJeztah: Oh! good call, forgot that `wasm` was a builtin... 😢

austinvazquez (approved):

pkg/adaptation/wasm-disabled.go
```diff
@@ -0,0 +1,27 @@
+//go:build nri_nowasm
```
AkihiroSuda:
```suggestion
//go:build no_wasm
```

For consistency with `no_btrfs`, `no_systemd`, etc.

https://github.com/containerd/containerd/blob/main/BUILDING.md#build-containerd

AkihiroSuda: Or maybe

```suggestion
//go:build no_wazero
```

AkihiroSuda (changes requested):
The build tag seems inconsistent with existing ones

https://github.com/containerd/nri/pull/253/changes#r2612981454

pkg/adaptation/wasm-disabled.go
klihub: @AkihiroSuda And if we end up flipping the logic around, to have WASM plugin support disabled by default and enabled by a tag, what would be a good build tag name ? enable_wasm_plugins ?

klihub:
> > > thanks!
> > > I think this change looks good; AFAICS, it doesn't paint us in a corner on considering some of the other options (functional option to add the plugin and looking further at what can be done for the API definitions).
> > 
> > 
> > Yes, I also think we should explore your 'inject WASM support using a functional option' further, and see if we can sort everything out to get it working, and then take a look at what that would mean in terms of maintenance wrt. expected WASM runtime API stability/etc.
> 
> nod.. I wonder if functional would allow for variants per runtime type?

If you mean plugging in other WASM implementations than wazero, then moving to a functional option to set up WASM support would make this theoretically/from a pure mechanics PoV possible. However, since our WASM support uses knqyf263/go-plugin to generate all the necessary bindings and other boilerplate to interact with WASM-compiled code from our protobuf definitions, and since go-plugin uses wazero, this would not really be possible in practice at the moment.

pkg/adaptation/wasm-disabled.go
AkihiroSuda: For compatibility reason, I don't think we can flip the logic around

klihub: Okay, I went with your original `no_wasm` suggestion as a build tag.

thaJeztah: @AkihiroSuda the reason for using a namespace / prefix was to reduce the likelihood of conflicting with other places; https://github.com/containerd/nri/pull/253#discussion_r2609853501

> Just thinking out loud here; build-tags cross all modules, so we should check if `nowasm` is potentially ambiguous (i.e., if it's already used elsewhere but for a different purpose, then setting the build-tag could cause side-effects).
> 
> If that's a problem indeed, we may need something more unique (`nri_no_wasm` or `nri_nowasm`)

AkihiroSuda (approved):
Thanks

klihub:
@mikebrow @dmcgowan Could you guys chime in if we are fine with the current build tag semantics, IOW keeping WASM support on by default for backward compatibility and turning it off using the `no_wasm` tag ? If we are fine with it, I'll merge this as is, then tag a v0.11.0 release.

marquiz (approved):
Looks good to me.

That said, if we think that disabling wasm support by default is desirable and a good thing, I think now would be a good time to do it (before we get to NRI v1.0). Put a big fat note in the release notes.

mikebrow (approved):
LGTM

klihub:
> Looks good to me.
> 
> That said, if we think that disabling wasm support by default is desirable and a good thing, I think now would be a good time to do it (before we get to NRI v1.0). Put a big fat note in the release notes.

I think disabling WASM makes the most sense for those who consume and compile containerd themselves from sources and really count bytes, for instance k3s. I'd be fine with keeping it enabled by default.

pkg/adaptation/wasm-disabled.go
AkihiroSuda: nri_no_wasm looks good

klihub: Okay, reverted build tag to be `nri_no_wasm`.

AkihiroSuda (approved):

PR #257, opened 2026-01-08
docs: add Linux sysctl's as restrictable by validation.

klihub:
Update README.md, adding the recently introduced Linux sysctls to the set of adjustments which can be restricted using the default validator.


mikebrow (approved):
LGTM

PR #258, opened 2026-01-09
docs: document which versions various NRI features were first available.

klihub:
Document which versions of NRI, containerd and CRI-O support first appeared for each piece of documented property for pods, containers, container adjustments, and container updates.

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,16 +134,23 @@ NRI plugins can subscribe to the following pod lifecycle events:
   - stopping
   - removal
 
-The following pieces of pod metadata are available to plugins in NRI:
+The following table lists the pod sandbox properties exposed to NRI plugins, together with
+the first NRI, containerd and CRI-O versions each was available in.
 
-  - ID
-  - name
-  - UID
-  - namespace
-  - labels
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -134,16 +134,24 @@ NRI plugins can subscribe to the following pod lifecycle events:
...
-  - name
-  - UID
-  - namespace
-  - labels
-  - annotations
-  - cgroup parent directory
-  - runtime handler name
+The following table lists the pod sandbox properties exposed to NRI plugins, together with
+the first NRI, containerd and CRI-O versions each was available in. An omitted specific
+version in the table implies NRI v0.2.0, containerd v1.7.0, and CRI-O v1.26.0.
```
samuelkarp: I'm okay with this just for getting the tables into the readme, but I think omitting the versions makes it pretty hard to understand as a reader.

mikebrow: yeah might as well just busy up the table .. ditto marks would also work.

klihub: Okay, I added the implied/omitted versions explicitly to each table instead.

PR #262, opened 2026-01-12
device-injector: implement sysctl adjustment annotations.

klihub:
Add support for adjusting sysctl settings.

### DIFF
--- a/plugins/device-injector/README.md
+++ b/plugins/device-injector/README.md
@@ -1,7 +1,14 @@
 ## Device Injector Plugin
 
-This sample plugin can inject Linux device nodes, CDI devices, and mounts into
-containers using pod annotations.
+This sample plugin can be used to adjust selected container properties
+prior to the creation of a container. The desired adjustments happen
+based on pod annotations. The possible adjustments include
+
+  - injection of Linux device nodes,
+  - injection of mounts,
+  - injection of CDI devices,
+  - injection of network devices,
+  - adjustments to sysctl settings
 
 ### Device Annotations
 
@@ -127,6 +134,29 @@ metadata:
 requests the injection of the host network interface `ens2.100` into container `c0`
 as the network interface `netdev0`, and the host network interface `ens2.101` into
 container `c1` as the network interface `netdev1`.
+
+### Sysctl Annotations
+
+Sysctl settings are annotated in a similar manner to devices, but using the
+`sysctl.noderesource.dev` annotation key prefix.
+
+The annotation value for sysctls is a map of sysctl keys to adjusted values.
+For instance, the following annotation
+
+```
+metadata:
+  name: sysctl-test
+  annotations:
+    sysctl.noderesource.dev/container.c0: |
+      net.core.somaxconn: "256"
+```
+
+will adjust the maximum number of queued TCP connections for a socket to 256.
+
+WARNING: Note that many of the sysctl settings are effective for the full pod
+instead of a single container. Also many of the settings are not namespaced at
+all. Trying to adjust such a setting will result in an error and a failure to
+create the container.
 
 ## Deployment
 
--- a/plugins/device-injector/device-injector.go
+++ b/plugins/device-injector/device-injector.go
@@ -55,6 +55,8 @@ const (
 	schedulerKey = "scheduling-policy.noderesource.dev"
 	// Deprecated: Prefix of the key used for scheduler attribute adjustment.
 	oldSchedulerKey = "scheduling-policy.nri.io"
+	// Prefix of the key used for sysctl adjustment.
+	sysctlKey = "sysctl.noderesource.dev"
 )
 
 var (
@@ -137,6 +139,10 @@ func (p *plugin) CreateContainer(_ context.Context, pod *api.PodSandbox, ctr *ap
 	}
 
 	if err := injectNetDevices(pod, ctr, adjust); err != nil {
+		return nil, nil, err
+	}
+
+	if err := adjustSysctl(pod, ctr, adjust); err != nil {
 		return nil, nil, err
 	}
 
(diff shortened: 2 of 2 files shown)


plugins/device-injector/README.md
```diff
@@ -128,6 +128,30 @@ requests the injection of the host network interface `ens2.100` into container `
 as the network interface `netdev0`, and the host network interface `ens2.101` into
 container `c1` as the network interface `netdev1`.
 
+### Sysctl Annotations
```
mikebrow, line 138: need to add sysctl to the top of the readme..

samuelkarp, line 138: It seems a little odd to me to put sysctls in the device injector.  This plugin has turned into a bit of a grab-bag.  Should we rename this so it's more generic of a plugin?

plugins/device-injector/README.md
```diff
@@ -128,6 +128,30 @@ requests the injection of the host network interface `ens2.100` into container `
...
+
+Sysctl settings are annotated in a similar manner to devices, but using the
+`sysctl.noderesource.dev` annotation key prefix. As with devices, the
+`sysctl.nri.io` annotation key prefix is also supported.
```
samuelkarp: I don't see a reason to support `nri.io` for new annotations.

plugins/device-injector/README.md
klihub: Yes, we should. I was thinking the same, basically to 1) rename the plugin as `container-adjuster`, 2) add command line options (and probably env. vars) to allow selectively disabling unwanted types of adjustments.

klihub: Do you prefer doing it as part of this PR, or as a separate one ?

klihub: Added.

samuelkarp: I'm fine with it being separate, but we should do it soon.

> 2) add command line options (and probably env. vars) to allow selectively disabling unwanted types of adjustments

Is our goal here mostly to demo different NRI adjustments or to provide a plugin we expect users to actually run? For a demo I don't think the command-line options are really necessary, but for a production-ready plugin we probably need more than just those changes.

plugins/device-injector/README.md
```diff
@@ -128,6 +135,29 @@ requests the injection of the host network interface `ens2.100` into container `
...
+    sysctl.noderesource.dev/container.c0: |
+      net.core.somaxconn: "256"
+```
+
+will adjust the maximum number of queued TCP connections for a socket to 256.
+
+WARNING: Note that many of the sysctl settings are effective for the full pod
+instead of a single container. Also many of the settings are not namespaced at
+all. Trying to adjust such a setting will result in an error and a failure to
+create the container.
```
samuelkarp, line 159: An error from the container runtime? I don't know if we have validation for namespaced sysctls within containerd...

plugins/device-injector/README.md
mikebrow: is also a nice test tool..

plugins/device-injector/device-injector.go
```diff
@@ -395,6 +401,48 @@ func parseScheduler(ctr string, annotations map[string]string) (*scheduler, erro
...
 }
 
+func adjustSysctl(pod *api.PodSandbox, ctr *api.Container, a *api.ContainerAdjustment) error {
+	sysctl, err := parseSysctl(ctr.Name, pod.Annotations)
+	if err != nil {
+		return err
+	}
+
+	if len(sysctl) == 0 {
+		log.Debugf("%s: no network sysctl adjustment annotated...", containerName(pod, ctr))
```
samuelkarp:
```suggestion
		log.Debugf("%s: no sysctl adjustment annotated...", containerName(pod, ctr))
```

plugins/device-injector/README.md
samuelkarp: Agreed, but test/demo are the same category to me.

mikebrow: as with ctr.. test/demo/debug tools tend to get used beyond our recommendations

samuelkarp: Understood. My question is still about intent though. If it is our _intent_ for this to be a more broadly-useful plugin, we have additional work to do. If not, then I think the command-line options are unnecessary.

plugins/device-injector/README.md
klihub: It's probably in runc/crun. With `runc` I got an error message specifically stating that I was trying to set a non-namespaced sysctl setting, and it did fail container creation.

plugins/device-injector/README.md
klihub: @mikebrow @samuelkarp We can limit our changes for now to those that we think are necessary regardless of whether we intend it as a test/demo tool or as a more generally useful plugin for folks to consume and use. I think those would include at least this PR, a few fixes to the docs for undocumented existing features I spotted, and the renaming of the plugins to something more closely reflecting reality (maybe `container-adjuster`).

Then we can decide later if we intend this as a more generally useful plugin that folks could consume, and do the remaining necessary work if decide to do so.

mikebrow: fair

plugins/device-injector/README.md
mikebrow: runc guys.. it's probably in lib container ... lib container guys.. it's the kernel

mikebrow (approved):
LGTM

plugins/device-injector/README.md
samuelkarp: > this PR, a few fixes to the docs for undocumented existing features I spotted, and the renaming of the plugins to something more closely reflecting reality

That sounds good. Let's punt on the selective disable logic until we've worked out how we want to position this plugin.

samuelkarp (approved):

PR #263, opened 2026-01-15
api: generate field ownership functions

samuelkarp:
:warning: AI-assisted code :warning:

---

Introduce a `protoc` generator (`tools/protoc-gen-owners`) to automatically generate the repetitive field ownership tracking functions. This refactors the `pkg/api/owners.go` file.

The new generator uses the `Field` enum in `api.proto` as the source of truth. The original `owners.go` file has been split logically into generated and handwritten parts:
- `owners_generated.go`: Contains the auto-generated boilerplate for most fields.
- `owners.go`: Retains the core handwritten logic, including constructors, conflict resolution, and special-cased handlers for fields with custom logic (`OciHooks`, `Rdt`).

This change reduces manual boilerplate, improves maintainability, and ensures consistency between the protobuf definitions and the Go API.

Assisted-by: gemini-cli


tools/protoc-gen-owners/main.go
```diff
@@ -0,0 +1,296 @@
...
+		fieldName := originalFieldName
+		if override, ok := singularOverrides[originalFieldName]; ok {
+			fieldName = override
+		}
+
+		isCompound := compoundFields[originalFieldName]
+		keyName := compoundKeys[originalFieldName]
+
+		if isCompound {
+
```
chrishenzie: Are these additional newlines intended?

pkg/api/owners.go
chrishenzie, line 1: I feel like we might want to add some file-level comment somewhere (maybe here or in the generator) clarifying under what circumstances should we hand-roll the owners logic (vs. have it auto-generated)

chrishenzie (commented):
Nice change, this cuts down on a lot of boilerplate. Left some small comments. It looks like this isn't passing DCO checks yet and needs a signature

tools/protoc-gen-owners/main.go
samuelkarp: No, I haven't done any polishing yet.  This is basically a first-draft of AI output (and why the PR is still in draft).

pkg/api/owners.go
samuelkarp: I'd like to get rid of the hand-written owners logic if I can, but these were exceptions to the pattern already.

PR #264, opened 2026-01-16
api: fix OCI hook ownership tracking.

klihub:
Fix hook ownership tracking and add a proper test case for it.

Ownership tracking for OCI hooks is supposed to be accumulative. The implementation tried to accumulate owners, but it was buggy. It tried to delete existing ownership by unclaiming which marks ownershipdeleted by the plugin instead of clearing it.

This went unnoticed since we lacked any kind of proper test for hook ownership accumulation. It was only noticed thanks to #263 which switches much of the ownership tracking code to generated from hand-written.

```release-note
Accumulate owners for OCI hook adjustments, disallowing commas in plugin names
```

### DIFF
--- a/pkg/adaptation/plugin.go
+++ b/pkg/adaptation/plugin.go
@@ -435,9 +435,9 @@ func (p *plugin) qualifiedName() string {
 // RegisterPlugin handles the plugin's registration request.
 func (p *plugin) RegisterPlugin(ctx context.Context, req *RegisterPluginRequest) (*RegisterPluginResponse, error) {
 	if p.isExternal() {
-		if req.PluginName == "" {
-			p.regC <- fmt.Errorf("plugin %q registered with an empty name", p.qualifiedName())
-			return &RegisterPluginResponse{}, errors.New("invalid (empty) plugin name")
+		if err := api.CheckPluginName(req.PluginName); err != nil {
+			p.regC <- fmt.Errorf("plugin registered with an invalid name: %w", err)
+			return &RegisterPluginResponse{}, fmt.Errorf("invalid plugin name: %w", err)
 		}
 		if err := api.CheckPluginIndex(req.PluginIdx); err != nil {
 			p.regC <- fmt.Errorf("plugin %q registered with an invalid index: %w", req.PluginName, err)
--- a/pkg/api/owners.go
+++ b/pkg/api/owners.go
@@ -475,14 +475,7 @@ func (f *FieldOwners) ClaimMount(destination, plugin string) error {
 }
 
 func (f *FieldOwners) ClaimHooks(plugin string) error {
-	plugins := plugin
-
-	if current, ok := f.simpleOwner(Field_OciHooks.Key()); ok {
-		f.clearSimple(Field_OciHooks.Key(), plugin)
-		plugins = current + "," + plugin
-	}
-
-	f.claimSimple(Field_OciHooks.Key(), plugins)
+	f.accumulateSimple(Field_OciHooks.Key(), plugin)
 	return nil
 }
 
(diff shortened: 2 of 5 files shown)


pkg/api/owners.go
```diff
@@ -678,6 +671,14 @@ func (f *FieldOwners) ClearRdt(plugin string) {
 	f.clearSimple(Field_RdtEnableMonitoring.Key(), plugin)
 }
 
+func (f *FieldOwners) accumulateSimple(field int32, plugin string) {
+	old, ok := f.simpleOwner(field)
+	if ok {
+		plugin = old + "," + plugin
+	}
+	f.Simple[field] = plugin
```
samuelkarp, line 679:
```suggestion
	old, _ := f.simpleOwner(field)
	f.Simple[field] = strings.Join([]string{old, plugin}, ",")
```

Though maybe that's more complicated.

pkg/api/owners_test.go
```diff
@@ -113,3 +113,24 @@ func TestCompoundClaims(t *testing.T) {
...
+	err = o.ClaimHooks("ctr0", "plugin1")
+	require.NoError(t, err, "ctr0 OCI hooks by plugin1")
+
+	// claim OCI hooks of ctr0
+	err = o.ClaimHooks("ctr0", "plugin2")
+	require.NoError(t, err, "ctr0 OCI hooks by plugin2")
+
+	owners, ok := o.HooksOwner("ctr0")
+	require.True(t, ok, "ctr0 has hooks owners")
+	require.Equal(t, "plugin0,plugin1,plugin2", owners, "ctr0 hooks owners")
```
samuelkarp, line 134: Are we validating anywhere that the plugin name does not contain a `,`?

klihub: No, we only had a non-emptiness check in place. I split the name check out to pkg/api next to the existing CheckPluginIndex() and made it much stricter wrt. what we had in place, only accepting [a-zA-Z0-9_+.-]. Much more restrictive than used to be, I hope not too restrictive in practice.

pkg/api/owners.go
klihub: I think you'd still need to test for existence (first or subsequent owner), otherwise the first owner would get prefixed with a comma if done like that. And with that then, it's really a matter of taste which one is simpler/more readable...

samuelkarp: > it's really a matter of taste which one is simpler/more readable...

agreed, I'm fine with the original implementation too.

pkg/adaptation/plugin.go
```diff
@@ -435,9 +435,9 @@ func (p *plugin) qualifiedName() string {
 // RegisterPlugin handles the plugin's registration request.
 func (p *plugin) RegisterPlugin(ctx context.Context, req *RegisterPluginRequest) (*RegisterPluginResponse, error) {
 	if p.isExternal() {
-		if req.PluginName == "" {
-			p.regC <- fmt.Errorf("plugin %q registered with an empty name", p.qualifiedName())
-			return &RegisterPluginResponse{}, errors.New("invalid (empty) plugin name")
+		if err := api.CheckPluginName(req.PluginName); err != nil {
```
samuelkarp, line 438: This is technically a breaking change now :cry: 

I think we should still take it in, but we need to make that clear in release notes.

samuelkarp (approved):

pkg/adaptation/plugin.go
klihub: Yeah, even if we relaxed the new check to only explicitly disallow commas, it still would be a breaking change.

mikebrow (approved):
LGTM

PR #265, opened 2026-01-20
api,adaptation,stub: version exchange and capability advertisement.

klihub:
This patch series
- introduces capabilities intended to advertise and require runtime NRI features
- adds an adaptation option for declaring runtime supported NRI capabilities
- adds a plugin/stub option for require a set of capabilities from the runtime

Updates: 
@samuelkarp @mikebrow I reworked this, ditching version-based checks and implementing more complete support for capability-based checks. Also added capability requirements to plugins which need some specific features that correspond to a capability.

Updates 2:
Rebased on latest HEAD with version exchange support merged.


tools/gen-git-version/generate.sh
```diff
@@ -0,0 +1,36 @@
...
+
+echo "" > $out
+emit "//go:build nri_git_version"
+emit "// Code generated by $0. DO NOT EDIT."
+emit ""
+emit "package version"
+emit ""
+emit "const ("
+emit "    FallbackVersion = \"$version\""
+emit ")"
```
samuelkarp: Rather than generating a Go file (which can be a little weird from a compilation perspective, which you're working around with a build-tag...) I think [`go:embed`](https://pkg.go.dev/embed) will work more reliably; we can have this tool generate a plain text file, use `//go:embed` to read it in a built binary, and have some conditional that determines which value to use.

pkg/api/version.go
```diff
@@ -0,0 +1,208 @@
...
+	if nriVersion == "" {
+		return unknown, fmt.Errorf("unknown runtime version %q %q", runtime, runtimeVersion)
+	}
+
+	return nriVersion, nil
+}
+
+var (
+	runtimeVersionMap = map[string]map[string]string{
+		"containerd": {
```
samuelkarp, line 56: Is this map only needed for versions released before this change? i.e., we'll have a couple more times maybe of hand-updating it but it should stay static afterward?

containerd 1.7 is EOL in September 2026, 2.2 is EOL in November, so we'd need to hand-update it until then?

pkg/api/api.proto
```diff
@@ -134,6 +136,34 @@ message LogRequest {
...
+  ADJUST_LINUX_SCHEDULING_POLICY = 15;
+  INPUT_LINUX_SCHEDULING_POLICY = 16;
+  ADJUST_LINUX_NETWORK_DEVICES = 17;
+  INPUT_LINUX_NETWORK_DEVICES = 18;
+  ADJUST_LINUX_RDT_CLOS = 19;
+  ADJUST_LINUX_SYSCTL = 20;
+  INPUT_LINUX_SYSCTL = 21;
+  EXTRA_CONTAINER_STATUS = 22;
+  INPUT_UID_AND_GIDS = 23;
+  MAX_CAPABILITY = 24;
```
samuelkarp: What is `MAX_CAPABILITY`?

pkg/api/api.proto
```diff
@@ -145,6 +175,10 @@ message ConfigureRequest {
   int64 registration_timeout = 4;
   // Configured request processing timeout in milliseconds.
   int64 request_timeout = 5;
+  // Version of NRI used by the runtime.
+  string NRI_version = 6;
```
samuelkarp, line 209:
```suggestion
  string nri_version = 6;
```

?

pkg/version/version.go
```diff
@@ -0,0 +1,117 @@
...
+	return version
+}
+
+// MajorMinorPatch returns the major.minor.patch prefix of the semantic version v.
+func MajorMinorPatch(v string) string {
+	return strings.TrimSuffix(strings.TrimSuffix(v, Build(v)), Prerelease(v))
+}
+
+// LastStableTagged returns the last stable tagged version less than or equal to v.
+func LastStableTagged(v string) (string, error) {
```
samuelkarp: This is probably better implemented using [semver.Sort](https://pkg.go.dev/golang.org/x/mod/semver#Sort) and semver.Compare.  The semver spec allows for more-complicated pre- and post-release notation (using `-` and `+`) than is handled here...and rather than reimplementing it we can use the semver library.

samuelkarp (commented):
I'm not entirely sure I understand the value of having both version comparison and capabilities. They both effectively serve the same purpose of understanding compatibility between plugin and runtime, but operate in slightly different ways. I tend to think capabilities are generally a better option in that they're more descriptive of the required functionality and need less manual mapping over time, but I can see some debugging value in exposing version too.

pkg/api/version.go
klihub: Yes, it is only for existing releases, and this would require some updates before it could be left alone/intact.

pkg/api/api.proto
klihub: It is the largest known capability + 1 or the number of capabilities... which I now realize is not necessary to enumerate, because we can always infer it with something like `len(Capability_name)`.

tools/gen-git-version/generate.sh
klihub: I'm not sure if that would eliminate the need for conditional compilation. I mean, the conditional compilation is only in place so that we can use a different method to determine the NRI version for plugins which are hosted in the main NRI repository. And the motivation for this is that (it looks to me as if) the golang runtime build info reports `"(devel)"` for the NRI repository itself, even when we are building from unmodified sources at an exact tag.

So then we have two choices for these plugins. 1) Either we ignore the replacement directive and take the (replaced) NRI version in the plugins' `go.mod` at face value, or 2) we try some other way to figure out what the NRI version of the repository might be and use that. This PR tried to do the second. Hence the go-generated git-based version detection at compile time and the extra build tag for the repo-hosted plugins to opt in for using that.

But maybe instead of this extra complication for a few sample plugins, we should instead just go with 1) above. Then whatever happens to be in a plugin's `go.mod` gets advertised as its NRI version.

pkg/api/api.proto
klihub: Yeah, not strong opinions, so I can update it like that. I chose NRI_version (NRIVersion) as we already have CDI_devices (CDIDevices) and NRI is a CDI-like TLA. WDYT ?

pkg/version/version.go
klihub: Okay, I rewrote this as `FindClosestMatch` which semver-looks for the largest version smaller or equal to the one we look for.  I had to toss in the stripping of any git-described version suffix to get any somewhat sane results for develoment versions.

tools/gen-git-version/generate.sh
samuelkarp: > I'm not sure if that would eliminate the need for conditional compilation.

`go:embed` wouldn't need to be conditional, though we'd still generate the same $version and store it in a text file.  Something like:

```go
package version

//go:embed version.txt
var fallbackVersion string

func FallbackVersion() string {
	if fallbackVersion == "" {
		return UnknownVersion
	}
	return fallbackVersion
}
```

To be clear I think the approach (2) above is reasonable, but I'd rather use `go:embed` like here than the build tag approach.

pkg/api/api.proto
samuelkarp: We can leave it as `NRI_version` then to be consistent.  But I wish we had it as `cdi_devices` rather than `CDI_devices`...

pkg/api/api.proto
samuelkarp: I think changing the value of an existing enum would be an incompatible proto change, so it's better to remove that.

tools/gen-git-version/generate.sh
klihub: Okay, but the problem with embed is that it is not optional. You cannot seed a string by go:embedding something only if it exists. Only using a bigger `fs.FS` hammer would allow semantically something equivalent. With a plain version string we also have to commit an empty `version.txt` to the repository. Otherwise compiling against NRI outside the repo would fail with a missing version.txt.

But then when compiling the repo-hosted plugins, we'd always end up with a dirty repository once we put the git-described NRI repo version in the originally empty file. So without extra logic, we'd never have a repo-hosted plugin reporting a clean NRI version without the `.m` suffix. Also, without extra logic I think making the repo dirty prior to compilation would break CI, where we verify that we build, run tests, etc. from a clean repo.

So if we go with embedding, then should we try to accommodate for a dirty repo in the affected places (version.txt generation, clean slate CI check), or go with fs.FS and version.txt becoming an embedded file which is present for in-repo compiled plugins but absent for others ? To me the latter then starts to sound more compelling.

klihub: @samuelkarp Okay, I reworked fallback to git version using a simpler go:generator, go:embed:ing, and without using related build tags. I suspect those bits are now as good/simple as they can get.

samuelkarp:
#271 was merged, this needs a rebase

pkg/stub/stub.go
```diff
@@ -283,33 +287,46 @@ func WithLogger(logger nrilog.Logger) Option {
 	}
 }
 
+// WithRequiredCapabilities sets the capabilities the plugin requires from the runtime.
```
mikebrow, line 290:
```suggestion
// WithRequiredRuntimeCapabilities sets the capabilities the plugin requires from the runtime.
```

pkg/stub/stub.go
```diff
@@ -283,33 +287,46 @@ func WithLogger(logger nrilog.Logger) Option {
...
+func WithRequiredCapabilities(required ...api.Capability) Option {
```
mikebrow, line 291:
```suggestion
func WithRequiredRuntimeCapabilities(required ...api.Capability) Option {
```

pkg/adaptation/adaptation_suite_test.go
```diff
@@ -3385,6 +3386,139 @@ var _ = Describe("NRI version exchange", func() {
...
+				api.Capability_INPUT_POSIX_RLIMITS,
+				api.Capability_INPUT_IP_ADDRESSES,
+				api.Capability_ADJUST_LINUX_CONTAINER_ARGS,
+			}
+			plugins = []*mockPlugin{
+				{
+					idx:  "00",
+					name: "test",
+					options: []stub.Option{
+						stub.WithRequiredCapabilities(supportedCapabilities...),
```
mikebrow, line 3457:
```suggestion
						stub.WithRequiredRuntimeCapabilities(supportedCapabilities...),
```

pkg/adaptation/adaptation_suite_test.go
```diff
@@ -3385,6 +3386,139 @@ var _ = Describe("NRI version exchange", func() {
...
+					},
+				},
+				{
+					idx:  "01",
+					name: "test",
+					options: []stub.Option{
+						stub.WithRequiredCapabilities(subsetCapabilities...),
```
mikebrow, line 3464:
```suggestion
						stub.WithRequiredRuntimeCapabilities(subsetCapabilities...),
```

pkg/adaptation/adaptation_suite_test.go
```diff
@@ -3385,6 +3386,139 @@ var _ = Describe("NRI version exchange", func() {
...
+				api.Capability_ADJUST_LINUX_CONTAINER_ARGS,
+			}
+			requiredCapabilities = []api.Capability{
+				api.Capability_ADJUST_LINUX_SCHEDULING_POLICY,
+			}
+			plugin = &mockPlugin{
+				idx:  "00",
+				name: "test",
+				options: []stub.Option{
+					stub.WithRequiredCapabilities(requiredCapabilities...),
```
mikebrow, line 3502:
```suggestion
					stub.WithRequiredRuntimeCapabilities(requiredCapabilities...),
```

...

plugins/device-injector/device-injector.go
```diff
@@ -730,6 +731,7 @@ func main() {
 	flag.StringVar(&pluginIdx, "idx", "", "plugin index to register to NRI")
 	flag.StringVar(&socketPath, "socket-path", "", "path of the NRI socket file")
 	flag.BoolVar(&verbose, "verbose", false, "enable (more) verbose logging")
+	flag.BoolVar(&noCaps, "no-caps", false, "don't require runtime NRI capabilities")
```
mikebrow, line 734:
```suggestion
	flag.BoolVar(&noCaps, "no-caps", false, "when true, disables verification of runtime capabilities for this plugin") 
```

mikebrow (commented):
couple nits

pkg/stub/stub.go
klihub: Updated.

pkg/stub/stub.go
klihub: Updated.

pkg/adaptation/adaptation_suite_test.go
klihub: Updated.

pkg/adaptation/adaptation_suite_test.go
klihub: Updated.

pkg/adaptation/adaptation_suite_test.go
klihub: Updated.

plugins/device-injector/device-injector.go
klihub: Updated.

PR #119, opened 2024-10-26
api: add support for reading Pod IPs

aojea:
Allow NRI plugins to read the allocated IPs to the pod on the PodSandbox hooks. This is specially useful for networking daemons that depend on the Pod IPs assigned via CNI by the container runtime (service mesh, network policies, ... ) to create seamless integrations without having to modify the existing environment.

Existing network integrations use techniques like:

- CNI chaining (invasive) as the implementation needs to modify the existing deployment and can cause disruption due to mutations on the existing configurations https://github.com/Azure/AKS/issues/4385

- Read Pod Status, that is too late in the process, since the pod lifecycle does not update the pod status until the containers are running, but there are applications like network policy implemetnations or service meshes that needs this information before the pod is completely created (ref: https://github.com/kubernetes/kubernetes/issues/85966)

- Inspect the host namespaces to gather the information, that is very intrusive and not portable, creating a tighly coupling between the implementations.

Add a nri plugin that can be used as reference implementations for this integration, the nri plugin just log the interesting networking details of the pod (ips and network namespace) on the different hooks of the pod lifecucle.

## Container runtime dependencies

This feature requires the container runtime to pass to the nri plugin the IP information.

Containerd patch
```diff
diff --git a/internal/cri/nri/nri_api_linux.go b/internal/cri/nri/nri_api_linux.go
index 0adcd9d11..598e656b2 100644
--- a/internal/cri/nri/nri_api_linux.go
+++ b/internal/cri/nri/nri_api_linux.go
@@ -633,6 +633,14 @@ func (p *criPodSandbox) GetPid() uint32 {
        return p.pid
 }

+func (p *criPodSandbox) GetIPs() []string {
+       if p.IP == "" {
+               return nil
+       }
+       ips := append([]string{p.IP}, p.AdditionalIPs...)
+       return ips
+}
+
 //
 // NRI integration wrapper for CRI Containers
 //
diff --git a/internal/nri/sandbox.go b/internal/nri/sandbox.go
index 85d6178aa..15ad9538d 100644
--- a/internal/nri/sandbox.go
+++ b/internal/nri/sandbox.go
@@ -34,6 +34,7 @@ type PodSandbox interface {
        GetLinuxPodSandbox() LinuxPodSandbox

        GetPid() uint32
+       GetIPs() []string
 }

 type LinuxPodSandbox interface {
@@ -55,6 +56,7 @@ func commonPodSandboxToNRI(pod PodSandbox) *nri.PodSandbox {
[...]

PR #267, opened 2026-01-28
Implement support for PodSandboxStatus RPC

MikeZappa87:
This PR is to add PodSandboxStatus support to NRI. 

The idea to why we want to bring this RPC in now is that we want to try and begin the process to deprecate the CNI and we need a way to return Pod IP's back to the kubelet. Without this we ether need to come up with a new RPC outside of NRI or experiment directly in the kubelet. Using NRI seems to be the best place right now. This is why the PodSandboxStatusResponse type only returns pod ip addresses, its open to more fields be added. 

I put together this branch in containerd that brings this NRI branch in and disables the CNI as a test bed. 

https://github.com/containerd/containerd/compare/main...MikeZappa87:containerd:mzappa/knext


LionelJouin:
This is a great thing to add, thanks @MikeZappa87 !

/cc @aojea @squeed @mskrocki

pkg/api/api.proto
```diff
@@ -224,6 +228,20 @@ message UpdatePodSandboxRequest {
 
 message UpdatePodSandboxResponse {}
 
+message PodSandboxStatusRequest {
+  // Pod being queried for status.
+  PodSandbox pod = 1;
+  // Whether verbose information should be returned.
+  bool verbose = 2;
```
klihub: nit: Do we use this for something at this point ? If not, we could just leave it out and add it if/once we need and start using it then.

pkg/api/api_ttrpc.pb.go-e
```diff
@@ -0,0 +1,276 @@
+// Code generated by protoc-gen-go-ttrpc. DO NOT EDIT.
```
klihub: Was this git mv'd then accidentally committed (I mean it's a `.go-e`) ?

pkg/adaptation/adaptation.go
```diff
@@ -238,6 +238,29 @@ func (r *Adaptation) UpdatePodSandbox(ctx context.Context, req *UpdatePodSandbox
...
+	for _, plugin := range r.plugins {
+		pluginRsp, err := plugin.podSandboxStatus(ctx, req)
+		if err != nil {
+			return nil, err
+		}
+		// Use the last plugin's non-empty IP response
+		if pluginRsp != nil && (pluginRsp.Ip != "" || len(pluginRsp.AdditionalIps) > 0) {
+			rsp = pluginRsp
+		}
+	}
```
klihub: What is the reason for the chosen semantics wrt. multiple plugins ? Especially for the who comes later wins overriding everything, even if the former set only one attribute and the latter only the other, in which case a(n AFAICT) non-conflicting setting is silently lost.

I think we should consider having a similar (or using the same) conflict detection/resolution mechanism as we have/do for containers. Even if we don't put it fully in place now/yet, we could try faking it here with something like...

```golang
func (r *Adaptation) PodSandboxStatus(ctx context.Context, req *PodSandboxStatusRequest) (*PodSandboxStatusResponse, error) {
	r.Lock()
	defer r.Unlock()
	defer r.removeClosedPlugins()

    var (
        rsp                = &PodSandboxStatusResponse{}
        ipOwner            = ""
        additionalIpsOwner = ""
    )
    
	for _, plugin := range r.plugins {
		pluginRsp, err := plugin.podSandboxStatus(ctx, req)
		if err != nil {
			return nil, err
		}
        if pluginRsp == nil || (pluginRsp.Ip == "" && len(pluginRsp.AdditionaIps) == 0) {
            continue
        }
        if pluginRsp.Ip != "" {
            if ipOwner != "" {
                return nil, fmt.Errorf("plugins %q and %q both tried to set PodSandboxStatus IP address field", ipOwner, plugin.name()
            }
            rsp.Ip = pluginRsp.Ip
            rsp.ipOwner = plugin.name()
        } 
        if len(pluginRsp.AdditionalIps) > 0 {
            if additionalIpOwner != "" {
                return nil, fmt.Errorf("plugins %q and %q both tried to set PodSandboxStatus additional IP addresses field", additionalIpOwner, plugin.name()) 
            }
            rsp.AdditionalIps = pluginRsp.AdditionalIps
            additionaIpsOwner = plugin.name()
        }
	}

    return rsp, nil
}
```

klihub (commented):
I have a few questions about the chosen semantics. But it might be that I just lack the necessary context here...

pkg/adaptation/adaptation.go
MikeZappa87: This is I am going to change. I did this for the PR. I had a to do note to fix. Will resolve

pkg/api/api_ttrpc.pb.go-e
MikeZappa87: This was part of the generated code, I thought it went with the change. I had to make a change in the make file to generate the go files since I am on a mac.

pkg/stub/stub.go
```diff
@@ -93,6 +93,14 @@ type PostUpdatePodInterface interface {
 	PostUpdatePodSandbox(context.Context, *api.PodSandbox) error
 }
 
+// PodSandboxStatusInterface handles PodSandboxStatus API requests.
+type PodSandboxStatusInterface interface {
+	// PodSandboxStatus relays a PodSandboxStatus request to the plugin.
+	// The plugin can return modified IP addresses for the pod sandbox.
+	// Returns: primary IP, additional IPs, error
+	PodSandboxStatus(context.Context, *api.PodSandbox) (string, []string, error)
```
MikeZappa87: @klihub we probably want to modify this signature or leave it? Not sure if want an envelope type here to reduce future changes?

klihub: > @klihub we probably want to modify this signature or leave it? Not sure if want an envelope type here to reduce future changes?

Yeah, since this feels like it might get more fields added later, either just return a *api.PodSandboxStatusResponse in the stub interface, or then a wrapping/envelope type like 

```proto
message PodSandboxStatus {
    string ip = 1;
    repeated strings additional_ips = 2;
}
```

with a corresponding `PodSandboxStatus status = 1;` field in the `PodSandboxStatusResponse` message, so then it is easy to extend this later with more fields. And you'd return a `(*api.PodSandboxStatus, error)` in the corresponding stub interface.

This is especially compelling, if that status might see some use on the plugin input side, or other context, some day. IOW if in some cases we could/would need to pass the same/similar info to plugins as input as well. Then it feels like a natural choice to have a dedicated wrapping type/message for it.

I guess there could be a third alternative too, which would be to limit the scope more closely/explicitly to network setup and then have this visible in naming too. IOW, to call these something else than a generic `PodSandboxStatus`, both in the RPC call and in the message name(s). But that's orthogonal to the 'what we should return from the stub interface' question of yours.

pkg/adaptation/adaptation.go
```diff
@@ -238,6 +238,47 @@ func (r *Adaptation) UpdatePodSandbox(ctx context.Context, req *UpdatePodSandbox
...
+			return nil, err
+		}
+		if pluginRsp == nil || (pluginRsp.Ip == "" && len(pluginRsp.AdditionalIps) == 0) {
+			continue
+		}
+		if pluginRsp.Ip != "" {
+			if ipOwner != "" {
+				return nil, fmt.Errorf("plugins %q and %q both tried to set PodSandboxStatus IP address field", ipOwner, plugin.name())
+			}
+			rsp.Ip = pluginRsp.Ip
```
aojea: we already return the pod ips https://github.com/containerd/nri/pull/119 , do you need anything else from the Status?

aojea: oh, ok, this is not for read, this is for update ... this has a high risk of causing problems with the runtimes, last time I checked I think that crio and containerd both had different hooks for the container IP setting and the NRI hook ... first of all I think that we should define the contract between the NRI hooks and the runtimes of at what point one thing is executed ... I added integration tests in containerd to ensure the current defined order and lifecycle is respected and does not break https://github.com/containerd/containerd/pull/11331 but we should do the same exercise in crio to avoid drifting the runtimes

pkg/adaptation/adaptation.go
```diff
@@ -238,6 +238,47 @@ func (r *Adaptation) UpdatePodSandbox(ctx context.Context, req *UpdatePodSandbox
 	return &UpdatePodSandboxResponse{}, nil
 }
 
+// PodSandboxStatus relays the corresponding CRI request to plugins.
+// If a plugin returns IP addresses, those will be returned to the caller.
+// An error is returned if multiple plugins attempt to set the same field.
```
aojea: this will be a nightmare to troubleshoot in prod on kubernetes clusters, since the error surfaced will most probably do not indicate something, the CNI will say it is correct, since it assignes the IP, and this NRI plugin will be probably opaque to the user ... I think we need to define first a better authoritative model for the IP assignment in the runtimes

aojea:
Thanks for doing this, I initiated this path with this in mind but I unfortunately didn't have the necessary time to continue, so super glad you are taking the initiative, added some comments from my previous research in this area

pkg/adaptation/adaptation.go
MikeZappa87: I think we have a couple things here. If this PR merges, the CNI would be optional so people would need to take that into consideration. I believe this concern is shared across NRI as well? NRI can modify fields on the Pod and containers without the user knowing. The way I was going to implement this in containerd at least was to not allow this RPC to be modified if the CNI was enabled, at least that was just a thought to prevent these conditions however open to all the ideas here.

pkg/adaptation/adaptation.go
MikeZappa87: https://github.com/containerd/containerd/blob/317286ac00e07bebd7d77925c0fef2af0d620e40/internal/cri/server/sandbox_run.go#L329

Are you talking about this? If you make any changes those are not commited back to the internal store of containerd. I am not 100% following you here? Since kubelet calls PodSandboxStatus, we need to hook into that RPC directly otherwise the change needs to be made in either RunPodSandbox or the kubelet? 

Having kubelet use something internal like reading a claim instead of PodSandboxStatus would be an idea for sure. If we update RunPodSandbox, we would need to have that update the internal store for example? 

https://github.com/kubernetes/kubernetes/blob/8c9c67c000104450cfc5a5f48053a9a84b73cf93/pkg/kubelet/kuberuntime/kuberuntime_manager.go#L1568 for reference.

pkg/adaptation/adaptation.go
MikeZappa87: What are your thoughts on the authority for ip assignments?

pfl: Just some thoughts. If the NRI plugin can specify all necessary networking details for a pre-existing interface, then the container runtime could perhaps configure everything itself. But CNIs exist to set up an optimal (to them or the problem at hand) overlay network between nodes, so involving CNIs should perhaps not be removed from the equation altogether?

aojea: >  NRI can modify fields on the Pod and containers without the user knowing

Yeah, but those are properties of the application itself, rlimits, cpu, and assigned by the runtime IIUIC .. the IP assigned is a property required to communicate with the rest of the world, and also assigned by another actor, so you are changing a convention that will generate a lot of issues in prod

@MikeZappa87  this idea of a dedicated hook

>  explicit Pre- and PostSetupNetwork

Makes it easier to think about, so runtime can choose 

```
if NRISetupNetwork {
  callNRINetworkPlugin
} if CNISetupNetwork {
  call CNI // current state
} else {
  return error, no network plugin configured
```

This way there is no risk someone assigns an IP and later it realizes is not being used

MikeZappa87: if NRISetupNetwork {
  callNRINetworkPlugin
} if CNISetupNetwork {
  call CNI // current state
} else {
  return error, no network plugin configured
  
The POC works in this way. However I wonder if I want to move this PR forward or not. I debate. It might be better to close for now.

aojea: at the end of the day is a containerd/crio decision, I can only provide non-binding opinions here

danwinship: > `An error is returned if multiple plugins attempt to set the same field.`

The runtime can return an arbitrary number of pod IPs to kubelet (with kubelet ignoring all but the first IP or the first dual-stack pair), so you could just aggregate them instead...

> this will be a nightmare to troubleshoot in prod on kubernetes clusters

Well... yes, if you install both a CNI-based pod network implementation and an NRI-based pod network implementation on the same node, then that will fail spectacularly, but "don't do that then".

(And this is nothing new: if you run any two of Calico, Cilium, and OVN-Kubernetes on the same node at the same time, it will also fail spectacularly. Sure, the runtime will only talk to whichever one of them has the lexically-first CNI config file name, but meanwhile, the node-level daemon of the other pod network will be reconfiguring the node in a way that will totally break the first pod network.)

> If this PR merges, the CNI would be optional

ISTM that the ideal implementation would be to make "CNI support" be an NRI plugin. (Maybe even the same plugin for containerd and cri-o?) And then if you don't want to use CNI, you just don't install/run the CNI NRI plugin.

danwinship:
> But back to the CNI/non-CNI issue. As most (all?) of the real-world CNIs seem to have network daemons running while getting configuration requests via CNI directory binaries, maybe there should be a revised protocol for contacting CNIs directly from container runtimes with a perhaps a fallback to current CNI messaging if needed.

There's been some discussion about that (https://github.com/containernetworking/cni/issues/821). The big problem is with delegated plugins (e.g., telling your pod network to invoke another CNI plugin for IPAM). Even if the pod network's daemon container mounts the host CNI plugins directory so that it can _find_ the delegated-to plugin, it still may not be able to invoke it successfully, since the other plugin is expecting to be invoked in a context that looks like the container runtime, not in the context of a random hostNetwork pod.

The other problem with CNI is that it is really trying to solve Kubernetes 1.0's problems, and we need an API that is trying to solve Kubernetes 1.36's problems.

So I would not worry about CNI. We should assume that CRI/NRI are going to evolve beyond it.

MikeZappa87:
I need to pick this back up, and resolve the conflicts so we can push it forward

pkg/adaptation/adaptation.go
MikeZappa87: I am going to implement it as @danwinship unless someone has objections?

MikeZappa87:
> I have a few questions about the chosen semantics. But it might be that I just lack the necessary context here...

What questions did you have?

pkg/api/api.proto
```diff
@@ -329,6 +333,18 @@ message RemoveContainerRequest {
 
 message RemoveContainerResponse {}
 
+message PodSandboxStatusRequest {
+  // Pod being queried for status.
+  PodSandbox pod = 1;
+}
+
+message PodSandboxStatusResponse {
+  // Primary IP address of the pod sandbox.
```
mikebrow, line 342: maybe put these fields into a PodSandboxNetworkStatus struct from the root of this message

also include an info map[string]string for the typical "extra" debug/status information at the root of the response

Maybe just use a repeated string for IPs.. as was done on the RunPadSandbox call where the primary is the first one..

pkg/api/api.proto
mikebrow: yeah verbose would be for the info map... I think we need it, see request for new network struct for IPs and for info to be passed back...

mikebrow (commented):
Was there consideration for putting the NRI plugin's pod sandbox response IP(s) into the RunPodSandbox request?   Getting the IPs at status time, currently is a read on the pod store metadata... moving it to status would change the pod lifecycle timing somewhat.  At any rate I think we need to stow this in the pod store? and consider using pod update if there is a dynamic change to the pod IPs.

MikeZappa87:
> Was there consideration for putting the NRI plugin's pod sandbox response IP(s) into the RunPodSandbox request? Getting the IPs at status time, currently is a read on the pod store metadata... moving it to status would change the pod lifecycle timing somewhat. At any rate I think we need to stow this in the pod store? and consider using pod update if there is a dynamic change to the pod IPs.

The considered alternative has always been for RunPodSandbox was to return the ip addresses instead. 

https://github.com/containerd/nri/blob/520641efc61c07ccb5a39b13dcfe5598ab1f9ab9/pkg/api/api.proto#L209

@aojea @danwinship @LionelJouin  thoughts??

mikebrow:
nod.. start with the NRI changes.. I like that this emulates the create/start split I'd like to put into CRI states.  I think the right time to fix IP reporting in the Kubelet is when we do the create/start split in the CRI api. We would then naturally migrate the nri network plugin api over to the create/run step.

pkg/api/api.proto
mikebrow: course if we're moving to runpodsandbox response.. then we don't need to pass in the status verbose flag :)

MikeZappa87:
> nod.. start with the NRI changes.. I like that this emulates the create/start split I'd like to put into CRI states. I think the right time to fix IP reporting in the Kubelet is when we do the create/start split in the CRI api. We would then naturally migrate the nri network plugin api over to the create/run step.

https://github.com/containerd/nri/pull/295

PR #268, opened 2026-02-03
runtime-tools: log container adjustments.

klihub:
This PR implements NRI audit logging for OCI Spec adjustments, which has been identified as one of the missing things we need to add (be)for(e) a v1.0. This patch
- adds an extended generator option for runtimes to set an external audit event logger, and
- updates the generator to use the configured logger for audit messages when an OCI Spec is adjusted

Here are updated trees for contained and CRI-O:
- [containerd](https://github.com/klihub/containerd/commits/devel/main/nri-audit-logging)
- [cri-o](https://github.com/klihub/cri-o/commits/devel/main/nri-audit-logging)

### DIFF
--- a/pkg/adaptation/adaptation.go
+++ b/pkg/adaptation/adaptation.go
@@ -231,7 +231,7 @@ func (r *Adaptation) UpdatePodSandbox(ctx context.Context, req *UpdatePodSandbox
 	for _, plugin := range r.plugins {
 		_, err := plugin.updatePodSandbox(ctx, req)
 		if err != nil {
-			return nil, err
+			return nil, pluginError(plugin, err)
 		}
 	}
 
@@ -257,7 +257,7 @@ func (r *Adaptation) RemovePodSandbox(ctx context.Context, evt *StateChangeEvent
 }
 
 // CreateContainer relays the corresponding CRI request to plugins.
-func (r *Adaptation) CreateContainer(ctx context.Context, req *CreateContainerRequest) (*CreateContainerResponse, error) {
+func (r *Adaptation) CreateContainer(ctx context.Context, req *CreateContainerRequest) (*CreateContainerResponse, *api.OwningPlugins, error) {
 	r.Lock()
 	defer r.Unlock()
 	defer r.removeClosedPlugins()
@@ -280,15 +280,20 @@ func (r *Adaptation) CreateContainer(ctx context.Context, req *CreateContainerRe
 		}
 		rpl, err := plugin.createContainer(ctx, req)
 		if err != nil {
-			return nil, err
+			return nil, nil, pluginError(plugin, err)
 		}
 		err = result.apply(rpl, plugin.name())
 		if err != nil {
-			return nil, err
+			return nil, nil, err
 		}
 	}
 
-	return r.validateContainerAdjustment(ctx, validate, result)
+	rpl, err := r.validateContainerAdjustment(ctx, validate, result)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	return rpl, result.owners, nil
 }
 
 // PostCreateContainer relays the corresponding CRI event to plugins.
@@ -319,7 +324,7 @@ func (r *Adaptation) UpdateContainer(ctx context.Context, req *UpdateContainerRe
 	for _, plugin := range r.plugins {
 		rpl, err := plugin.updateContainer(ctx, req)
 		if err != nil {
-			return nil, err
+			return nil, pluginError(plugin, err)
 		}
 		err = result.apply(rpl, plugin.name())
 		if err != nil {
@@ -346,7 +351,7 @@ func (r *Adaptation) StopContainer(ctx context.Context, req *StopContainerReques
 	for _, plugin := range r.plugins {
 		rpl, err := plugin.stopContainer(ctx, req)
 		if err != nil {
-			return nil, err
+			return nil, pluginError(plugin, err)
 		}
 		err = result.apply(rpl, plugin.name())
 		if err != nil {
@@ -376,7 +381,7 @@ func (r *Adaptation) StateChange(ctx context.Context, evt *StateChangeEvent) err
 	for _, plugin := range r.plugins {
 		err := plugin.StateChange(ctx, evt)
 		if err != nil {
-			return err
+			return pluginError(plugin, err)
 		}
 	}
 
@@ -719,4 +724,11 @@ func (b *PluginSyncBlock) Unblock() {
 		b.r.syncLock.RUnlock()
 		b.r = nil
 	}
+}
+
+func pluginError(plugin *plugin, err error) error {
+	if err == nil {
+		return nil
+	}
+	return fmt.Errorf("plugin %q: %w", plugin.name(), err)
 }
--- a/pkg/api/owners.go
+++ b/pkg/api/owners.go
@@ -49,7 +49,7 @@ func (o *OwningPlugins) ClaimHooks(id, plugin string) error {
 }
 
 func (o *OwningPlugins) HooksOwner(id string) (string, bool) {
-	return o.ownersFor(id).simpleOwner(Field_OciHooks.Key())
+	return o.ownersFor(id).SimpleOwner(Field_OciHooks.Key())
 }
 
 func (o *OwningPlugins) ClearHooks(id, plugin string) {
@@ -148,7 +148,7 @@ func (f *FieldOwners) ClaimHooks(plugin string) error {
 }
 
 func (f *FieldOwners) HooksOwner() (string, bool) {
-	return f.simpleOwner(Field_OciHooks.Key())
+	return f.SimpleOwner(Field_OciHooks.Key())
 }
 
 func (f *FieldOwners) ClearHooks(plugin string) {
@@ -175,7 +175,7 @@ func (f *FieldOwners) ClearRdt(plugin string) {
 }
 
 func (f *FieldOwners) accumulateSimple(field int32, plugin string) {
-	old, ok := f.simpleOwner(field)
+	old, ok := f.SimpleOwner(field)
 	if ok {
 		plugin = old + "," + plugin
 	}
@@ -200,7 +200,7 @@ func (f *FieldOwners) compoundOwnerMap(field int32) (map[string]string, bool) {
 	return m.Owners, true
 }
 
-func (f *FieldOwners) compoundOwner(field int32, key string) (string, bool) {
+func (f *FieldOwners) CompoundOwner(field int32, key string) (string, bool) {
 	if f == nil {
 		return "", false
 	}
@@ -214,7 +214,7 @@ func (f *FieldOwners) compoundOwner(field int32, key string) (string, bool) {
 	return plugin, ok
 }
 
-func (f *FieldOwners) simpleOwner(field int32) (string, bool) {
+func (f *FieldOwners) SimpleOwner(field int32) (string, bool) {
 	if f == nil {
 		return "", false
 	}
--- a/pkg/runtime-tools/generate/generate.go
+++ b/pkg/runtime-tools/generate/generate.go
@@ -44,6 +44,7 @@ type UnderlyingGenerator interface {
 	AddPostStopHook(postStopHook rspec.Hook)
 	AddPreStartHook(preStartHook rspec.Hook)
 	AddProcessEnv(name, value string)
+	AddProcessRlimits(typ string, hard, soft uint64)
 	AddLinuxResourcesDevice(allow bool, devType string, major, minor *int64, access string)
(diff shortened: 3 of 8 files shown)


mikebrow (commented):
Cool.. Maybe add a little more detail to the short explanation in the [README.md](https://github.com/containerd/nri/blob/main/README.md#wrapped-oci-spec-generator) maybe links to the containerd/crio use of these opts.  (where the integration test happens)

pkg/runtime-tools/generate/generate.go
```diff
@@ -222,11 +239,30 @@ func (g *Generator) AdjustEnv(env []*nri.KeyValue) {
 			}
 			if m, ok := mod[keyval[0]]; ok {
 				delete(mod, keyval[0])
-				if _, marked := m.IsMarkedForRemoval(); !marked {
+				if key, marked := m.IsMarkedForRemoval(); !marked {
+					g.log(AuditAddProcessEnv,
+						Fields{
+							"value.name":  m.Key,
+							"value.value": m.Value,
```
chrishenzie: I think we may want to omit logging env var values since this could potentially leak secrets.

pkg/api/owners.go
```diff
@@ -684,7 +684,7 @@ func (f *FieldOwners) Conflict(field int32, plugin, other string, qualifiers ...
 		plugin, other, qualify(field, qualifiers...))
 }
 
-func (f *FieldOwners) compoundOwnerMap(field int32) (map[string]string, bool) {
+func (f *FieldOwners) CompoundOwnerMap(field int32) (map[string]string, bool) {
```
chrishenzie: I don't think this method is used in the generator so it could remain unexported.

pkg/runtime-tools/generate/generate.go
```diff
@@ -148,6 +151,20 @@ func WithCDIDeviceInjector(fn func(*rspec.Spec, []string) error) GeneratorOption
...
+// WithLogger specifies a function for logging (audit) messages.
+func WithLogger(logger Logger, owners *nri.FieldOwners) GeneratorOption {
+	return func(g *Generator) {
+		g.logger = logger
+		g.owners = owners
+	}
+}
+
+// Fields can be used to pass extra information for logged messages.
+type Fields = map[string]any
```
chrishenzie, line 163: If this is a `map`, how do we ensure deterministic ordering of logged fields? Is that an exercise left for the `Logger`?

pkg/runtime-tools/generate/generate.go
```diff
@@ -222,11 +239,30 @@ func (g *Generator) AdjustEnv(env []*nri.KeyValue) {
 			}
 			if m, ok := mod[keyval[0]]; ok {
 				delete(mod, keyval[0])
-				if _, marked := m.IsMarkedForRemoval(); !marked {
+				if key, marked := m.IsMarkedForRemoval(); !marked {
+					g.log(AuditAddProcessEnv,
+						Fields{
+							"value.name":  m.Key,
```
chrishenzie, line 245: It depends on the `Logger` implementation, but this `value.` prefix will produce longer lines. Is the idea here for the `Logger` to parse this prefix in order to potentially handle these fields a certain way?

pkg/adaptation/adaptation.go
```diff
@@ -257,7 +257,7 @@ func (r *Adaptation) RemovePodSandbox(ctx context.Context, evt *StateChangeEvent
 }
 
 // CreateContainer relays the corresponding CRI request to plugins.
-func (r *Adaptation) CreateContainer(ctx context.Context, req *CreateContainerRequest) (*CreateContainerResponse, error) {
+func (r *Adaptation) CreateContainer(ctx context.Context, req *CreateContainerRequest) (*CreateContainerResponse, *api.OwningPlugins, error) {
```
chrishenzie, line 260: Can you provide more context why we need to return the `OwningPlugins`?

pkg/runtime-tools/generate/generate.go
klihub: Yes. This is (expected to be) ultimately passed on to slog, logrus, or some other structured logger and then they deal with it as they see fit. This not different in any way from how you would expect (or not expect) a `log.G(ctx).WithFields(Fields{...})` in containerd core code to result in deterministic logging order.

pkg/runtime-tools/generate/generate.go
klihub: No, the idea is to avoid conflicting with some existing key in an outer logging context. In this particular instance, to avoid a known conflict in CRI-O, which uses the `name` key to put the container name in the logging context, so that every logged message has (among others) also the name of the associated container. There would certainly be other ways to try avoiding it, and it would also be possible to not log adjustments this way but as a single value formatted with `%v`. The approach taken here is just one possibility.

klihub: But `nri.plugin` could really just be `plugin` everywhere. I really don't think that could conflict with anything already in the context.

pkg/runtime-tools/generate/generate.go
klihub: Replaced it with a `<value omitted>`.

pkg/adaptation/adaptation.go
klihub: The approach taken by this PR is to log adjustments to the OCI Spec as/when they are made.

Therefore we need to pass ownership information along with the actual adjustment results, so we can eventually shove it in our `runtime-tools/generator` wrapper which then will use this new information to pull out the responsible/source plugin when it logs OCI Spec changes as they are being made.

With everything else staying the same (so still the same approach by large), one alternative to this detail would be to add a new `OwningPlugins owners = 4;` to the `CreateContainerResponse` message in `api.proto`, document that it is for the runtime with anything filled in on the stub/plugin side simply getting ignored, and then instead of returning it separately here, just set that new response.Owner field to it and let it travel in the response alongside with the rest of the data.

pkg/runtime-tools/generate/generate.go
samuelkarp: The `Logger` can also add prefixes or adjust as necessary, especially if we have well-known fields here. I don't think we need to think too hard about collision.  The `Logger` could unconditionally prefix `nri.` on every `Field` key.

pkg/runtime-tools/generate/generate.go
```diff
@@ -222,11 +239,30 @@ func (g *Generator) AdjustEnv(env []*nri.KeyValue) {
...
+							"value.value": "<value omitted>",
```
samuelkarp, line 246:
```suggestion
							"value.value": "<removed>",
```
might be more clear.

Though `<removed>` (or any string we pick here) could be an actual value. Might be worth making sure the value name has a `-` prefix?

pkg/runtime-tools/generate/generate.go
samuelkarp: It's worth considering here whether we want a map or to have named fields. A lot of the lines in this diff use the same names (`value.name`, `value.value`, `nri.plugin`).

klihub: The repetitive logging of value.name, value.value in most of the cases is due to most of the adjusted fields having scalar values.

By 'named fields' do you mean a struct that has as fields the 'union' of all possible fields (IOW now keys) that we log for various adjustments, and then set those fields in a struct instead as opposed to key/value pairs in a map ?

Wouldn't that be a bit unnatural as the provided logger function is just acting as a postman towards a structured logger ? And wouldn't it also force the logger function to have to interpret the struct, so it can pick only the fields set and pass them on to a structured logger ?

pkg/runtime-tools/generate/generate.go
klihub: The "value omitted" content is an attempt to indicate that we do not log the value of environment variables. @chrishenzie was concerned about possibly logging secrets in the environment, so I changed it to this.

But my impression is that your comment is about a different context, when we remove environment values. But those we log with a different message and only with the variable name included. But did I misinterpret your comment ?

pkg/api/owners.go
klihub: I only exported it because I think it should have been exported already initially for (the convenience of) custom validating plugins. But it's true that it is not required by this PR, so I reverted it to be unexported.

pkg/runtime-tools/generate/generate.go
```diff
@@ -836,3 +1165,78 @@ func (g *Generator) initConfigLinuxIntelRdt() {
...
+	if g.logger != nil {
+		g.logger(event, fields)
+	}
+}
+
+func (g *Generator) simpleOwner(field nri.Field) string {
+	owner := "unknown"
+
+	if g.owners != nil {
+		owner, _ = g.owners.SimpleOwner(field.Key())
```
chrishenzie: If this doesn't find a plugin, this will return `""`, which should override `"unknown"` above

pkg/runtime-tools/generate/generate.go
```diff
@@ -836,3 +1165,78 @@ func (g *Generator) initConfigLinuxIntelRdt() {
...
+	AuditSetProcessArgs             = "set process arguments"
+	AuditRemoveAnnotation           = "remove annotation"
+	AuditAddAnnotation              = "add annotation"
+	AuditAddOCIHook                 = "add OCI hook"
+	AuditSetLinuxCPUPeriod          = "set linux CPU period"
+	AuditSetLinuxCPUQuota           = "set linux CPU quota"
+	AuditSetLinuxCPUShares          = "set linux CPU shares"
+	AuditSetLinuxCPUSetCPUs         = "set linux cpuset CPUs"
+	AuditSetLinuxCPUSetMems         = "set linux cpuset mems"
+	AuditSetLinuxCPURealtimeRuntime = "set linux cpuset mems"
```
chrishenzie: Should this be `"set linux CPU realtime runtime"`?

pkg/runtime-tools/generate/generate.go
```diff
@@ -836,3 +1165,78 @@ func (g *Generator) initConfigLinuxIntelRdt() {
...
+	AuditSetLinuxCPURealtimePeriod  = "set linux cpuset mems"
```
chrishenzie: Should this be `"set linux CPU realtime period"`?

pkg/runtime-tools/generate/generate_suite_test.go
```diff
@@ -42,12 +96,14 @@ var _ = Describe("Adjustment", func() {
 				adjust *api.ContainerAdjustment
 			)
 
+			al := newLogger()
 			rg := &rgen.Generator{Config: spec}
-			xg := xgen.SpecGenerator(rg)
+			xg := xgen.SpecGenerator(rg, xgen.WithLogger(al.Log, nil))
```
chrishenzie, line 103: It looks like `owners` is `nil` for all of these logger instantiations. Do we want any cases that pass in owners to verify the `nri.plugin` is correct?

chrishenzie (dismissed):

pkg/runtime-tools/generate/generate.go
klihub: Okay, I assume you'd prefer `"unknown"` owner not to be overridden with an empty one, so I updated it to work like that. And made the same change for identical code in `compoundOwner`.

pkg/runtime-tools/generate/generate.go
klihub: Thanks for spotting this. Fixed.

pkg/runtime-tools/generate/generate.go
klihub: Thanks for spotting. Fixed.

pkg/runtime-tools/generate/generate_suite_test.go
klihub: Yes, we should. I added owning/nri.plugin verification to tests.

klihub:
@chrishenzie Added test cases. PTAL.

pkg/runtime-tools/generate/generate.go
```diff
@@ -836,3 +1165,84 @@ func (g *Generator) initConfigLinuxIntelRdt() {
...
+			owner = o
+		}
+	}
+
+	return owner
+}
+
+// Audit 'events' we use in logged audit messages.
+const ( //nolint:revive
+	AuditRemoveProcessEnv           = "remove environment variable"
```
samuelkarp, line 1203: These are all human-readable strings, but I think it'd actually be more valuable to have something that's closer to the actual field adjustment.  For example, `process.env` which matches the JSON path in the OCI bundle.  For being the most precise, I think we might want something as expressive as [JSON Patch](https://jsonpatch.com/) ([RFC 6902](https://datatracker.ietf.org/doc/html/rfc6902/)); this could describe the exact modification that a plugin is making.

PR #269, opened 2026-02-05
Add writable-cgroups experimental plugin

chrishenzie:
Adds a new `writable-cgroups` plugin, designed to enable safe delegation of cgroup management to containers. This plugin allows containers to mount `/sys/fs/cgroup` as read-write, enabling workloads (like AI/ML frameworks) to manage their own sub-cgroups.

This plugin serves as a reference implementation and test-bed for validating the `nsdelegate` security model proposed in KEP-5474 as an alternative to introducing new Kubernetes API fields.

@Divya063 @samuelkarp


chrishenzie:
I'm still unsure on the name, because all the other NRI plugins have "verb-focused" names. Maybe `cgroup-adjuster` is more appropriate?

plugins/writable-cgroups/writable-cgroups.go
```diff
@@ -0,0 +1,245 @@
...
+	"github.com/moby/sys/mountinfo"
+	"github.com/sirupsen/logrus"
+
+	"github.com/containerd/nri/pkg/api"
+	"github.com/containerd/nri/pkg/stub"
+)
+
+const (
+	// WritableCgroupsAnnotation is the annotation key that enables writable cgroups.
+	WritableCgroupsAnnotation = "cgroups.noderesource.dev/writable"
```
klihub: nit: Anything against using `writable-cgroups.noderesource.dev` as the annotation key ?

klihub:
> I'm still unsure on the name, because all the other NRI plugins have "verb-focused" names. Maybe `cgroup-adjuster` is more appropriate?

`cgroup-adjuster` sounds very generic to me compared to what the plugin does, implying more adjustments to containers' cgroups rather than their cgroup mounts. Maybe `cgroup-mount-adjuster` ?

klihub:
@chrishenzie I only skimmed through it quickly yet, but it LGTM. Should we also add a `contrib/kustomize/writable-cgroups` (or whatever we end up calling this) like we have for the other plugins ?

plugins/writable-cgroups/README.md
```diff
@@ -0,0 +1,94 @@
...
+(`/sys/fs/cgroup`) inside containers.
+
+## Purpose & Context
+
+This plugin serves as a test-bed for validating the behavior and security model
+of "delegated cgroup management" in Kubernetes environments. It was developed in
+response to [KEP-5474](https://github.com/kubernetes/enhancements/issues/5474),
+which originally proposed adding explicit API support for writable cgroups.
+
+However, the evolving consensus is to move away from API additions and instead
```
samuelkarp: My understanding is the current consensus is still that a user-facing API change is desired, but that writable cgroups should only be available when `nsdelegate` is present.

plugins/writable-cgroups/README.md
```diff
@@ -0,0 +1,94 @@
...
+```yaml
+annotations:
+  cgroups.noderesource.dev/writable.container.<container_name>: "true"
+```
+
+## Status
+
+This plugin is **experimental** and intended for testing and validation purposes
+only. It is not recommended for production use until the behavior is
+standardized in upstream container runtimes.
```
samuelkarp: I think once the behavior is standardized there's no need for this plugin either?

plugins/writable-cgroups/writable-cgroups.go
```diff
@@ -0,0 +1,245 @@
...
+	defer f.Close()
+
+	filter := func(info *mountinfo.Info) (skip, stop bool) {
+		if info.FSType != "cgroup2" {
+			return true, false
+		}
+		if info.Mountpoint == "/sys/fs/cgroup" || info.Mountpoint == "/sys/fs/cgroup/unified" {
+			return false, true
+		}
+		return true, false
```
samuelkarp: I find these hard to read:

```
if (condition) {
	// error case
}
if (condition) {
	// success case
}
// error case
```

I'd swap the logic in the second conditional (apply [De Morgan's law](https://en.wikipedia.org/wiki/De_Morgan%27s_laws)) to make the indented return also an error so the final line is the success case (i.e., the mount we're looking for).

```go
if info.Mountpoint != "/sys/fs/cgroup" && info.Mountpoint != "/sys/fs/cgroup/unified"
```

plugins/writable-cgroups/writable-cgroups.go
```diff
@@ -0,0 +1,245 @@
...
+	}
+
+	if len(mounts) == 0 {
+		return false, nil
+	}
+
+	return slices.Contains(strings.Split(mounts[0].VFSOptions, ","), "nsdelegate"), nil
+}
+
+// isWritableCgroupsEnabled checks if the pod has the writable cgroups annotation.
```
samuelkarp:
```suggestion
// isWritableCgroupsEnabled checks if the container has the writable cgroups annotation.
```

samuelkarp (approved):

plugins/writable-cgroups/writable-cgroups.go
chrishenzie: Ack, reordered lines to make success case last

plugins/writable-cgroups/README.md
chrishenzie: Correct. Tried to make that more explicit in the README

plugins/writable-cgroups/README.md
chrishenzie: That's right. Updated README to clarify current proposed solution

plugins/writable-cgroups/writable-cgroups.go
chrishenzie: Works for me. The only limitation is there is a 63 character limit for Pod annotations, so if a container name is long, it might push us over the edge when appended with the `.container.<container_name>` suffix.

```py
>>> len("writable-cgroups.noderesource.dev.container.")
44
```

So long as a container name is less than 19 characters it should be fine. We could potentially remove  `.container` and save an extra 10 characters. I'm okay with keeping as-is though.

plugins/writable-cgroups/README.md
```diff
@@ -0,0 +1,96 @@
...
+```yaml
+annotations:
+  writable-cgroups.noderesource.dev: "true"
+```
+
+**Container-Level (Applies to a specific container):**
+
+```yaml
+annotations:
+  writable-cgroups.noderesource.dev.container.<container_name>: "true"
```
klihub: IIRC, in other plugins the per-container (pod) annotation syntax is slightly different. I think it would be better to use the same syntax here for consistency, which then would be `writable-cgroups.noderesource.dev/container.<container_name>: "true"`. We even have a common implementation for it as `pkg/plugin.GetEffectiveAnnotation()` if you are fine with all the 3 allowed syntaxes: `$key/container.$ctr: $value`, `$key/pod: $value`, and `$key: $value`.

plugins/writable-cgroups/writable-cgroups.go
```diff
@@ -0,0 +1,245 @@
...
+	if len(mounts) == 0 {
+		return false, nil
+	}
+
+	return slices.Contains(strings.Split(mounts[0].VFSOptions, ","), "nsdelegate"), nil
+}
+
+// isWritableCgroupsEnabled checks if the container has writable cgroups enabled.
+func isWritableCgroupsEnabled(pod *api.PodSandbox, container *api.Container, l *logrus.Entry) bool {
+	perContainerAnnotation := fmt.Sprintf("%s.container.%s", WritableCgroupsAnnotation, container.Name)
```
klihub: The syntax here is slightly different from other plugins which use the `$key/container.$ctr: $value` syntax for per-container annotations. It would be better to use that for consistency here.

Maybe you could also consider using `pkg/plugin.GetEffectiveAnnotation()` which implements that syntax (as well as `$key/pod: $value` and `$key: $value` for pod-level annotations).

chrishenzie: Sounds good, refactored to use this library

plugins/writable-cgroups/README.md
chrishenzie: Agreed. Updated to follow consistent format

PR #271, opened 2026-02-19
adaptation,stub: exchange NRI versions in use.

klihub:
This PR adds NRI version to the pieces of information exchanged between a plugin and the runtime during plugin registration.

@mikebrow @samuelkarp @chrishenzie I split this out from #265 for easier review. These are self-contained and do not depend on the rest in #265. Some bits there (version based inferred capabilities for old releases) depend on this, but I'm not sure if there is consensus about adding capability advertisement in the first place. If there is we'll just need to merge this first then rebase.

```release-note
Exchange NRI versions between plugins and the runtime during registration
```


pkg/api/version.go
```diff
@@ -0,0 +1,210 @@
...
+import (
+	"fmt"
+	"maps"
+	"slices"
+
+	"github.com/containerd/nri/pkg/version"
+)
+
+const (
+	// UnknownVersion is an alias of version.UnknownVersion.s
```
samuelkarp: typo (`s`)

pkg/api/version.go
```diff
@@ -0,0 +1,210 @@
...
+	UnknownVersion = version.UnknownVersion
+	// DevelVersion is an alias for version.DevelVersion.
+	DevelVersion = version.DevelVersion
+)
+
+// InferVersionFromRuntime tries to infer the version of NRI used in the runtime.
+func InferVersionFromRuntime(runtime, runtimeVersion string) (string, error) {
+	var (
+		unknown    = version.UnknownVersion
```
samuelkarp: This local `unknown` seems redundant since we already have `UnknownVersion`?

pkg/version/version.go
```diff
@@ -0,0 +1,137 @@
...
+
+	return version
+}
+
+// MajorMinorPatch returns the major.minor.patch prefix of the semantic version v.
+func MajorMinorPatch(v string) string {
+	return strings.TrimSuffix(strings.TrimSuffix(v, Build(v)), Prerelease(v))
+}
+
+// FindClosestMatch returns the largest version smaller or equal to a given one.
```
samuelkarp, line 70: It would be helpful to have some unit tests for FindClosestMatch and StripGitSuffix.  Not a lot, just a few basic ones.

pkg/version/version.go
```diff
@@ -0,0 +1,137 @@
...
+
+import (
+	"runtime/debug"
+	"strconv"
+	"strings"
+
+	"golang.org/x/mod/semver"
+)
+
+var (
```
samuelkarp: Is there a particular reason for these redefinitions?

pkg/version/version.go
```diff
@@ -0,0 +1,137 @@
...
+// We expect a valid git suffix to be of the form "-N-gSHA1[.m], where
+// N is an decimal integer and SHA1 is a hexadecimal integer.
+func StripGitSuffix(version string) string {
+	mmp := MajorMinorPatch(version)
+	pre := Prerelease(version)
+	if mmp+pre != version {
+		return version
+	}
+
+	if pre == "" || pre[0] != '-' {
```
samuelkarp: I find `len(pre) == 0` easier to read since the next condition indexes into it.

pkg/version/version.go
```diff
@@ -0,0 +1,137 @@
...
+		return version
+	}
+
+	commits, gsha1, ok := strings.Cut(pre[1:], "-")
+	if !ok || gsha1 == "" || gsha1[0] != 'g' {
```
samuelkarp: same here with `len(ghsa1) == 0`

pkg/version/version.go
```diff
@@ -0,0 +1,137 @@
...
+
+	if version == DevelVersion {
+		return fallbackVersion()
+	}
+
+	return version
+}
+
+// MajorMinorPatch returns the major.minor.patch prefix of the semantic version v.
+func MajorMinorPatch(v string) string {
```
samuelkarp: Looks like this can be private (`majorMinorPatch`).

pkg/version/version.go
```diff
@@ -0,0 +1,137 @@
...
+		}
+		latest = ver
+	}
+	return latest
+}
+
+// StripGitSuffix strips any git described suffix from a version string.
+// We expect a valid git suffix to be of the form "-N-gSHA1[.m], where
+// N is an decimal integer and SHA1 is a hexadecimal integer.
+func StripGitSuffix(version string) string {
```
samuelkarp: Looks like this can be private (`stripGitSuffix`).

pkg/api/version.go
klihub: Fixed.

pkg/api/version.go
klihub: Replaced with UnknownVersion.

pkg/version/version.go
klihub: Added unit tests for these.

pkg/version/version.go
klihub: Updated.

pkg/version/version.go
klihub: Updated.

pkg/version/version.go
klihub: Unexported.

pkg/version/version.go
klihub: Unexported.

pkg/version/version.go
klihub: The (bit vague) idea was to shadow `x/mod/semver`, providing (a symbol for) everything we need from it through `pkg/version` so they do not leak visibily through to our other packages that import `pkg/version`. This is split out from #265 with capability advertisement, where we need a bit more from semver (but not this much). I can remove them if it's considered a bad idea.

samuelkarp: None of these are called from outside the `version` package though, as far as I can tell.

I'd probably recommend just removing them for now. If we end up needing more usage of `x/mod/semver` elsewhere later, we can address it then.  Right now this is a bit of a [YAGNI smell](https://testing.googleblog.com/2017/08/code-health-eliminate-yagni-smells.html).

pkg/version/version_test.go
```diff
@@ -0,0 +1,159 @@
...
+		},
+		{
+			name:     "prerelease version without git suffix",
+			version:  "v1.2.3-alpha.1",
+			expected: "v1.2.3-alpha.1",
+		},
+		{
+			name:     "working tree version with git suffix",
+			version:  "v0.11.0-37-g41b9c58",
+			expected: "v0.11.0",
```
samuelkarp, line 59: Maybe a couple more that don't quite match:

* `v1.2.3-45`
* `v1.2.3-45-g`
* `v1.2.3-45-gz`

pkg/version/version_test.go
```diff
@@ -0,0 +1,159 @@
...
+			name:       "closest match for a pre-release version",
+			version:    "v1.2.4-alpha.1",
+			candidates: candidates,
+			expected:   "v1.2.3",
+		},
+		{
+			name:       "closest match for a working tree version",
+			version:    "v0.11.0-37-g41b9c58",
+			candidates: candidates,
+			expected:   "v0.10.9",
```
samuelkarp, line 165: maybe test `v0.10.9-37-g41b9c58` too so we know that the git working tree is correctly excluded too

klihub: Added additional test case.

pkg/version/version_test.go
klihub: Added additional test cases.

pkg/version/version.go
klihub: Okay. Removed all of them.

samuelkarp (approved):

mikebrow (approved):
LGTM

Issue #272, opened 2026-02-25
Provide basic runtime metrics for NRI plugins


chrishenzie:
To improve the observability of the NRI framework and its plugins, we need to provide basic runtime metrics. This is a requirement for v1.0. Currently this data is not collected which makes troubleshooting difficult.

This is a tracking issue to discuss the design and implementation of basic runtime metrics for NRI plugins.

Specifically, the NRI framework needs to collect and provide data such as:

- Plugin counts (registered, active, etc.)
- Success and failure counts for plugin invocations/adjustments
- General plugin health

In the v1.0 requirements doc it's mentioned that the NRI framework itself should not expose these metrics directly. Instead, it should be passed to the container runtime, which can then expose it using its existing mechanisms (Prometheus, OpenTelemetry, etc.) for consumption.

PR #274, opened 2026-02-26
wire protocol: use dedicated RPC calls for all pod and container life-cycle events.

klihub:
This PR updates the wire protocol/ttrpc API to have proper RPC calls and messages for all pod and container life-cycle events.

The added calls and messages replace the 'catch-most' `StateChange` call which is now used to relay {Run,Stop,Remove}PodSandbox pod and {PostCreate,Start,PostStart, PostUpdate,Remove} container events using a single 'event type + field union' message on the wire. The visible stub/plugin and runtime adaptation interfaces are unaffected. The PR leaves `StateChange` in the protocol to allow transparent backward compatibility, and updates the runtime adaptation to fall back to using StateChange if the plugin side NRI version does not implement the new RPC calls yet.

This has been attempted a few times earlier in connection with other changes, but since we've had multiple other real functional changes in flight, we always shied away from the resulting rebase conflicts and seeing this through. This time this is a self-contained PR with no other functional changes.

Although functionally straightforward, it's still painfully intrusive as it touches a lot of boilerplate. Currently we have 3 remaining PRs open which touch the protocol and will conflict: #166 (memory policy adjustment), #183 (plugin authentication), and #271 (version exchange). Of these #166 and #217 are probably small footprint enough to be fairly easily rebases (either way around).

Here are my updated runtime test trees with these changes included:
- [containerd]( https://github.com/klihub/containerd/tree/devel/main/nri-dedicated-lifecycle-events)
- [cri-o]( https://github.com/klihub/cri-o/tree/devel/main/nri-dedicated-lifecycle-events)

```release-note
Use dedicated RPC calls for all pod and container life-cycle events via the NRI wire protocol
```

PR #277, opened 2026-03-05
add basic metrics for NRI framework

chrishenzie:
Provides necessary hooks for container runtimes to collect basic runtime metrics regarding NRI plugin activity. This improves observability and helps satisfy v1.0 requirements.

Introduces a `Metrics` interface that allows us to collect plugin counts, invocations, adjustments, and latency information.

To avoid burying state change events under a generic operation name, `StateChange` invocations append the specific event type using human-readable strings (e.g., "StateChange/CreateContainer").

These hooks are called during plugin invocations and container lifecycle events. Runtimes can provide their own implementation of the `Metrics` interface (e.g. via a `WithMetrics` option) to route these data points into their existing telemetry setups.

Ref: https://github.com/containerd/nri/issues/272
Signed-off-by: Chris Henzie <chrishenzie@gmail.com>
Assisted-by: gemini-cli

```release-note
Add basic metrics collection for the NRI framework
```

### DIFF
--- a/pkg/adaptation/adaptation.go
+++ b/pkg/adaptation/adaptation.go
@@ -75,6 +75,7 @@ type Adaptation struct {
 	builtin     []*builtin.BuiltinPlugin
 	syncLock    sync.RWMutex
 	wasmService *api.PluginPlugin
+	metrics     Metrics
 }
 
 var (
@@ -137,6 +138,16 @@ func WithBuiltinPlugins(plugins ...*builtin.BuiltinPlugin) Option {
 	}
 }
 
+// WithMetrics allows consumers to register an implementation of the Metrics interface.
+func WithMetrics(m Metrics) Option {
+	return func(r *Adaptation) error {
+		if m != nil {
+			r.metrics = m
+		}
+		return nil
+	}
+}
+
 // WithDefaultValidator sets up builtin validator plugin if it is configured.
 func WithDefaultValidator(cfg *validator.DefaultValidatorConfig) Option {
 	return func(r *Adaptation) error {
@@ -177,6 +188,7 @@ func New(name, version string, syncFn SyncFn, updateFn UpdateFn, opts ...Option)
 		socketPath:  DefaultSocketPath,
 		syncLock:    sync.RWMutex{},
 		wasmService: wasmService,
+		metrics:     &noopMetrics{},
 	}
 
 	for _, o := range opts {
@@ -540,6 +552,7 @@ func (r *Adaptation) removeClosedPlugins() {
 
 	r.plugins = active
 	r.validators = validators
+	r.metrics.UpdatePluginCount(len(r.plugins))
 }
 
 func (r *Adaptation) startListener() error {
@@ -607,7 +620,9 @@ func (r *Adaptation) acceptPluginConnections(l net.Listener) error {
 					r.validators = append(r.validators, p)
 				}
 				r.sortPlugins()
+				count := len(r.plugins)
 				r.Unlock()
+				r.metrics.UpdatePluginCount(count)
 				log.Infof(ctx, "plugin %q connected and synchronized", p.name())
 			}
 			r.finishedPluginSync()
--- a/pkg/adaptation/metrics.go
+++ b/pkg/adaptation/metrics.go
@@ -1,0 +1,47 @@
+/*
+   Copyright The containerd Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package adaptation
+
+import "time"
+
+// Metrics defines the interface that a consumer can implement to collect
+// and emit metrics regarding NRI plugin activity.
+// Implementations of this interface must be thread-safe, as these methods
+// are invoked concurrently across multiple goroutines handling plugin requests.
+type Metrics interface {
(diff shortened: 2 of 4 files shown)


pkg/adaptation/plugin.go
```diff
@@ -695,7 +713,9 @@ func (p *plugin) StateChange(ctx context.Context, evt *StateChangeEvent) (err er
 	ctx, cancel := context.WithTimeout(ctx, getPluginRequestTimeout())
 	defer cancel()
 
-	if err = p.impl.StateChange(ctx, evt); err != nil {
+	err = p.impl.StateChange(ctx, evt)
+	recordPluginInvocation(p.name(), "StateChange", err)
```
chrishenzie: Note to self: This probably needs fixing. Do we want to track per event type?

chrishenzie: Fixed in latest revision, appending the specific state to the label

pkg/adaptation/metrics_test.go
```diff
@@ -0,0 +1,127 @@
...
+		updates:    updates,
+		evicts:     evicts,
+	})
+}
+
+func (m *mockMetrics) UpdatePluginCount(count int) {
+	m.pluginCount = count
+}
+
+func TestMockMetrics(t *testing.T) {
```
samuelkarp: This is just unit-testing the mock itself...

pkg/adaptation/plugin.go
```diff
@@ -695,7 +713,14 @@ func (p *plugin) StateChange(ctx context.Context, evt *StateChangeEvent) (err er
 	ctx, cancel := context.WithTimeout(ctx, getPluginRequestTimeout())
 	defer cancel()
 
-	if err = p.impl.StateChange(ctx, evt); err != nil {
+	err = p.impl.StateChange(ctx, evt)
```
samuelkarp, line 733: I expect this will cause some conflicts with https://github.com/containerd/nri/pull/274

klihub (approved):
LGTM. Thank you for getting this done!

pkg/adaptation/adaptation.go
```diff
@@ -604,6 +617,7 @@ func (r *Adaptation) acceptPluginConnections(l net.Listener) error {
 					r.validators = append(r.validators, p)
 				}
 				r.sortPlugins()
+				r.metrics.UpdatePluginCount(len(r.plugins))
 				r.Unlock()
```
samuelkarp: Invoking an externally injected, consumer-provided interface (`r.metrics.UpdatePluginCount`) while holding a mutex introduces the opportunity for a deadlock (or an unbounded time for the lock to be held). This could happen if the consumer's metric sink blocks, performs I/O, or attempts to acquire other locks.

It'd be better to capture the required state while under the lock, release the lock, and then emit the metric.

```suggestion
				count := len(r.plugins)
				r.Unlock()
				r.metrics.UpdatePluginCount(count)
```

pkg/adaptation/metrics_test.go
```diff
@@ -0,0 +1,324 @@
...
+	}
+
+	return m, p
+}
+
+func TestPluginSynchronizeMetrics(t *testing.T) {
+	m, p := setupTestPlugin()
+
+	_, err := p.synchronize(context.Background(), nil, nil)
+	assert.Nil(t, err)
```
samuelkarp: This code validates error states using a raw nil check assertion. When asserting on errors in tests, prefer the assertion library's built-in error matchers (e.g., `assert.NoError`) so that failure output is automatically formatted and the underlying error message is printed if the invariant breaks.

```suggestion
	_, err := p.synchronize(context.Background(), nil, nil)
	assert.NoError(t, err)
```

*(Note: Consider applying this to the other `assert.Nil(t, err)` validations in this test suite for `createContainer`, `updateContainer`, `stopContainer`, `updatePodSandbox`, `StateChange`, and `ValidateContainerAdjustment`.)*

pkg/adaptation/metrics.go
```diff
@@ -0,0 +1,45 @@
...
+   limitations under the License.
+*/
+
+package adaptation
+
+import "time"
+
+// Metrics defines the interface that a consumer can implement to collect
+// and emit metrics regarding NRI plugin activity.
+type Metrics interface {
```
samuelkarp, line 25: The `Metrics` interface definition lacks a strict concurrency contract. Because its methods are invoked directly from container lifecycle handlers (e.g., `createContainer`, `stopContainer`) running in multiple goroutines, the interface contract should explicitly mandate thread-safe implementations to prevent data races and potential panics in downstream consumer code.

```suggestion
// Metrics defines the interface that a consumer can implement to collect
// and emit metrics regarding NRI plugin activity.
// Implementations of this interface must be thread-safe, as these methods
// are invoked concurrently across multiple goroutines handling plugin requests.
type Metrics interface {
```

samuelkarp (commented):
LGTM after addressing these comments (LLM-generated, but I agree with them).

pkg/adaptation/adaptation.go
chrishenzie: Done

pkg/adaptation/metrics_test.go
chrishenzie: Done

pkg/adaptation/metrics.go
chrishenzie: Done

klihub:
@samuelkarp Can you take another look ? Chris has addressed all review comments. We could get this in if we're happy with it. I could then rebase #274 on top of it, so we could start kicking that one forward next.

pkg/adaptation/plugin.go
klihub: @samuelkarp I rebased this on top of  #274, in preparation of merging that first.

klihub:
> @samuelkarp Can you take another look ? Chris has addressed all review comments. We could get this in if we're happy with it. I could then rebase #274 on top of it, so we could start kicking that one forward next.

I went ahead an readily rebased #274 on top of this.

mikebrow (approved):
LGTM 

todo: needs an integration test

samuelkarp (commented):
I know I'm late here but this LGTM too.

PR #279, opened 2026-03-09
adaptation: close plugin if initial sync fails.

klihub:
Don't leave a plugin connected but effectively unregistered if initial plugin synchronization fails. Log the fact as an error instead of info and close the plugin.

```release-note
Close plugins if initial synchronization fails to prevent unregistered connections
```


fuweid (approved):

mikebrow (approved):
LGTM

fuweid (dismissed):

fuweid (approved):
LGTM

fuweid (approved):

Issue #282, opened 2026-03-31
Kubernetes / NRI communication path


aojea:
This was found by @gauravkghildiyal , see [doc](https://docs.google.com/document/d/1bssPQW4L_RVyVtGhq1rWidxFSVmvN2iwEHG0t5KN7S4/edit?tab=t.0#heading=h.ss87mhtpkle3) with more details.

 in DRANET, a DRA driver for kubernetes networking, we implement an architecture with 2 legs for the driver: 

1. Kubelet via DRA
2. Container runtime via NRI

This architecture allow to handle complex workloads that require to execute at different stages of a pod creation. Also, this guarantees that the kubernetes system will not schedule nothing to the node until the DRA driver in the node is running, the driver can store the state in memory during the kubelet DRA NodePrepareResources() callback, and use the NRI callback to execute the actions on the Pod during the container runtime stage. 

However, there is still a race in this process, if for any case the runtime is restarted, the DRA driver will not be able to notice it and the Pod can be running without the NRI call back called ... we need to ensure that a specific driver is always called for a specific pod 

NRI added the concept of required plugins https://github.com/containerd/nri/pull/278 , this indicates that drivers can implement this behavior globally modifying the configuration or via annotations.

Since global is an admin option that requires changing containerd config, is hard for independent drivers to modify it, so we need to use the annotations path, however, the DRA hook in kubelet does not have a way to plumb dumb annotations (Device plugin does have)

I commented this during kubecon with NRI maintainers @klihub @kad they mentioned it should be possible to use the existing NodePrepareResources() hook to be able to use CDI to allow this functionality, but may require some work in some places to enable it.

@klihub @gauravkghildiyal @samuelkarp @chrishenzie

klihub:
> Since global is an admin option that requires changing containerd config, is hard for independent drivers to modify it, so we need to use the annotations path, however, the DRA hook in kubelet does not have a way to plumb dumb annotations (Device plugin does have)
>
> I commented this during kubecon with NRI maintainers @klihub @kad they mentioned it should be possible to use the existing NodePrepareResources() hook to be able to use CDI to allow this functionality, but may require some work in some places to enable it.

@aojea @kad I took a look at this, to see 1) what exactly is there in CDI for annotations, 2) if it could be used for this, and 3) what is the extent of changes required to make this work.

It was pretty easy to shoestring together for a working proto. Fundamentally two things were needed:
- allow CDI to annotate a container when a CDI device is injected
[...]

Issue #114, opened 2024-10-05
NRI plugins penalized with death if they take 2 seconds


mbaynton:
## Description

NRI plugins running in `containerd` by default have 2 seconds per event to provide a response. This is fine. But, if it misses a single response in that timeframe, it is closed / cut off from future events. For plugins built on github.com/containerd/nri, that results in the process exiting.

There is a specific set of errors that induce this close-the-connection behavior:
https://github.com/containerd/nri/blob/7b3bcee97eeaa0bdeef8dd771e5daeb73d4e0123/pkg/adaptation/plugin.go#L520-L533

The other ones in that list look very reasonable. But, I'd like to suggest that a plugin responding to one event in more than (by default) 2 seconds doesn't indicate that the plugin has entirely failed and it can probably still be used for future events, so a better behavior would be to simply time out that one event but continue.

lengrongfu:
I was thinking if we can handle this logic by adding an auto restart plugin, we can add `WithOnClose` this method to handler close.

```
opts = append(opts, stub.WithOnClose(func() {
		log.Info("Stopping plugins.")
		closeCtx <- struct{}{}
	}))
```

Andreagit97:
> filed https://github.com/containerd/ttrpc/issues/203

The feature proposed here seems really useful. Is there any discussion in progress regarding it?

We use an NRI plugin in our product to extract information about the container startup, and while the extraction flow is currently below the default 2 seconds, we will always be concerned about exceeding this limit. 
One alternative is to use the NRI to inject the OCI hook and use the OCI hook to do the job so that we are safe even if we exceed the 2-second timeout.
This works, but actually, it is just a workaround to avoid the timeout since we don't really need an OCI hook for our task.

Is there anything planned in this direction? Maybe a per-plugin timeout or a way to detect if the plugin is still alive

PR #286, opened 2026-04-09
define pod sandbox lifecycle contract

aojea:
Define the contract for the PodSandbox hooks for the  NRI plugins.

The Sandbox hooks are based on the CRI-API RPCs , since the OCI runtime only specify the container lifecycle.

/assign @samuelkarp @haircommander

### DIFF
--- a/docs/pod-sandbox-lifecycle.md
+++ b/docs/pod-sandbox-lifecycle.md
@@ -1,0 +1,162 @@
+# NRI Pod Sandbox Lifecycle Hooks
+
+## Relationship to CRI API
+
+This specification defines how NRI plugins interact with pod sandbox lifecycle events. The underlying pod sandbox operations are defined by the [Kubernetes CRI API](https://github.com/kubernetes/cri-api):
+
+- **RunPodSandbox (CRI)**: Creates and starts a pod-level sandbox. Runtimes must ensure the sandbox is in the ready state on success.
+- **StopPodSandbox (CRI)**: Stops any running process that is part of the sandbox and directs the runtime to reclaim certain pod resources (e.g. Network Namespace, CNI teardown, and image mounts). May be called multiple times, and is idempotent.
+- **RemovePodSandbox (CRI)**: Removes the sandbox. If there are any running containers, they must be forcibly terminated and removed.
+
+This NRI specification details when and under what conditions NRI plugins receive notifications for these events, ensuring plugins can reliably depend on consistent sandbox state across different runtime implementations.
+
+## Overview
+
+The pod sandbox lifecycle consists of three distinct phases, each with a corresponding NRI event that plugins can subscribe to:
+
+1. **RunPodSandbox**: Fired during the the runtime CRI RunPodSandbox execution, after the PodSandbox is created but before setting the pod to running and then replying success to CRI RunPodSandbox request.
+2. **StopPodSandbox**: Fired when the runtime initiates CRI StopPodSandbox
+3. **RemovePodSandbox**: Fired when the runtime performs CRI RemovePodSandbox
+
+For each event, this specification defines:
+
+- **Sandbox State Contract**: What sandbox infrastructure conditions runtimes MUST satisfy when firing the NRI event
+- **Plugin Responsibilities and Capabilities**: What plugins can safely do in response to the event
+
+## RunPodSandbox
+
+**CRI Operation**: RunPodSandbox - Creates and starts a pod-level sandbox.
+
+**NRI Event Timing**: The RunPodSandbox NRI event is fired after the runtime has successfully executed most of the CRI RunPodSandbox operation; NRI plugin execution is the final step before the sandbox reaches a "Ready" state. The Kubelet does not start workload containers until after the sandbox becomes "Ready".
+
+### Sandbox State Contract
+
+When the runtime fires the RunPodSandbox NRI event, it guarantees:
+
+- The Pod-level cgroup hierarchy has been established
+- The Sandbox namespaces (IPC, Network, UTS) are created and active
+- Network setup has been fully configured (network interfaces are up and assigned addressing)
+- The pod IP address (if applicable) is assigned and available
+- The "pause" container (if the runtime uses one) is running
+- All prerequisite operations for workload container startup are complete, the pod is in the "unknown state" and will become "Ready" once the NRI event is processed. This guarantees the NRI plugin has a window to allocate resources for the pod before any workload containers are started.
+
+### Plugin Responsibilities and Capabilities
+
+Upon receiving the RunPodSandbox event, plugins can safely:
+
+- Access the network namespace and inspect network configuration
+- Perform network-level operations or monitoring
+- Inject sandbox-level hardware configurations (e.g., RDMA, RoCEv2)
+- Establish plugin-specific tracking or monitoring for the pod
+- Store initial state or baseline metrics for later reference
+
+Plugins should treat this as an initialization phase. The sandbox infrastructure will remain accessible throughout the pod's lifetime until StopPodSandbox is called.
+
+## StopPodSandbox
+
+**CRI Operation**: StopPodSandbox - Stops any running process that is part of the sandbox and reclaims certain pod resources (e.g. Network Namespace, CNI teardown, and image mounts).
+
+**NRI Event Timing**: The StopPodSandbox NRI event is fired when the runtime initiates the CRI StopPodSandbox operation.
+
+### Sandbox State Contract
+
+When the runtime fires the StopPodSandbox NRI event, it guarantees:
+
+- Workload containers within the sandbox are stopped or are stopping
+- **CRITICAL**: The sandbox infrastructure still exists and remains fully accessible during this hook
+- The pod resources allocated by the runtime; such as network namespace, CNI networks, and image mounts; are not unmounted or deleted until this hook completes
+- The pod's cgroups remain accessible
+- All pod-level resources remain stable until this hook returns
+
+### Plugin Responsibilities and Capabilities
+
+StopPodSandbox is the designated cleanup and observation phase for plugins. Upon receiving this event, plugins can:
+
+- Access the pod's network namespace to read final telemetry or metrics
+- Collect final state for observability or troubleshooting
+- Detach hardware interfaces or reconfigure resources
+- Clean up custom firewall configurations, routing rules, or other network-level state
+- Perform graceful cleanup or resource release before sandbox teardown
+
+**Important**: Plugin processing must complete within the configured request timeout. Do not assume sandbox access persists after this hook returns or times out.
+
+## RemovePodSandbox
+
+**CRI Operation**: RemovePodSandbox - Removes the sandbox and forcibly terminates any remaining containers.
+
+**NRI Event Timing**: The RemovePodSandbox NRI event is fired when the runtime initiates the CRI RemovePodSandbox operation, just prior to removing the pod from the pod list.
+
+### Sandbox State Contract
+
+When the runtime fires the RemovePodSandbox NRI event:
+
+- All workload containers have been removed
+- The StopPodSandbox operation has completed
+- Network setup teardown may be underway or complete
+- The pod's namespaces (Network, IPC, UTS) may have already been deleted
+- Pod-level cgroups may be destroyed
+- Sandbox infrastructure access is **not guaranteed**
+
+### Plugin Responsibilities and Capabilities
+
+RemovePodSandbox is strictly for plugin-internal cleanup. Plugins MUST NOT attempt to access pod infrastructure (namespaces, cgroups, network configuration) during this hook, as their existence is not guaranteed.
+
+Plugins receiving this event should only:
+
+- Clean up plugin-internal memory caches or object tracking associated with the podSandboxID
+- Remove host-level tracking files, database entries, or other locally stored pod references
+- Release any plugin resources held for this specific pod
+- Perform final accounting or bookkeeping
+
+**Important**: This hook is informational only. Plugins should not assume any pod infrastructure exists. Only clean up information the plugin created or stored internally.
+
+## Event Ordering and Guarantees
+
+Runtimes MUST guarantee the following ordering:
+
+1. **RunPodSandbox** NRI event fires after successful CRI RunPodSandbox execution, but before the pod is set to the "Ready" state.
+2. **StopPodSandbox** NRI event fires during CRI StopPodSandbox execution, just prior to removing the runtime pod resources allocated by the runtime; such as network namespace, CNI networks, and image mounts
+3. **RemovePodSandbox** NRI event fires during CRI RemovePodSandbox execution
+4. These events MUST fire in strict order: RunPodSandbox → StopPodSandbox → RemovePodSandbox
+5. No workload containers will be started until after RunPodSandbox hook completes
+6. All workload containers will be stopped before StopPodSandbox hook is called
+7. No network resource reclamation should occur during StopPodSandbox hook execution
+
+See the [CRI API specification](https://github.com/kubernetes/cri-api) for details on each CRI operation.
+
+## Plugin Implementation Guidance
+
+### Subscribing to Events
+
+Plugins subscribe to these events during the Configure phase by returning the appropriate event flags in the ConfigureResponse:
+
+- `Event_RUN_POD_SANDBOX` (1 << 0) for RunPodSandbox
+- `Event_STOP_POD_SANDBOX` (1 << 1) for StopPodSandbox
+- `Event_REMOVE_POD_SANDBOX` (1 << 2) for RemovePodSandbox
+
+These events are delivered to plugins using the RunPodSandbox, StopPodSandbox and RemovePodSandbox event handlers.
+
+### Timeout Handling
+
+All plugin processing must complete within the configured request timeout. Plugins should plan accordingly:
+
+- **RunPodSandbox**: Failure may result in pod creation failure
+- **StopPodSandbox**: Non-blocking for subsequent operations; the plugin should not depend on completion of subsequent teardown
+- **RemovePodSandbox**: Non-blocking; removal will proceed regardless of plugin timeout
+
+### Error Handling
+
+On the teardown path, plugin errors MUST NOT prevent the operation from proceeding. Runtimes MUST ensure that a failing plugin cannot block pod or container teardown:
+
+- **RunPodSandbox errors**: A plugin error may prevent the pod from being created, depending on runtime policy. Plugins bear responsibility for errors they return at this phase.
+- **StopPodSandbox errors**: A plugin error MUST NOT prevent the sandbox from being stopped. The runtime MUST proceed with teardown regardless of plugin failures.
+- **RemovePodSandbox errors**: A plugin error MUST NOT prevent the sandbox from being removed. The runtime MUST proceed with removal regardless of plugin failures.
+
+
+### Multi-Plugin Coordination
+
+When multiple plugins are active:
+
+- All RunPodSandbox hooks complete before first workload container starts
+- Hooks execute in plugin index order; later plugins should not assume earlier plugins' modifications will persist
+- RemovePodSandbox hooks are independent; plugins should not rely on side effects from other plugins
--- a/docs/pod-sandbox-lifecycle.md
+++ b/docs/pod-sandbox-lifecycle.md
@@ -14,7 +14,7 @@ This NRI specification details when and under what conditions NRI plugins receiv
 
 The pod sandbox lifecycle consists of three distinct phases, each with a corresponding NRI event that plugins can subscribe to:
 
-1. **RunPodSandbox**: Fired during the the runtime CRI RunPodSandbox execution, after the PodSandbox is created but before setting the pod to running and then replying success to CRI RunPodSandbox request.
+1. **RunPodSandbox**: Fired during the runtime CRI RunPodSandbox execution, after the PodSandbox is created but before setting the pod to running and then replying success to CRI RunPodSandbox request.
 2. **StopPodSandbox**: Fired when the runtime initiates CRI StopPodSandbox
 3. **RemovePodSandbox**: Fired when the runtime performs CRI RemovePodSandbox
 
@@ -35,6 +35,7 @@ When the runtime fires the RunPodSandbox NRI event, it guarantees:
 
 - The Pod-level cgroup hierarchy has been established
 - The Sandbox namespaces (IPC, Network, UTS) are created and active
+- The Sandbox will not be reused
 - Network setup has been fully configured (network interfaces are up and assigned addressing)
 - The pod IP address (if applicable) is assigned and available
 - The "pause" container (if the runtime uses one) is running
@@ -138,7 +139,7 @@ These events are delivered to plugins using the RunPodSandbox, StopPodSandbox an
 
 ### Timeout Handling
 
-All plugin processing must complete within the configured request timeout. Plugins should plan accordingly:
+All plugin processing must complete within the configured request timeout. A plugin timeout is treated as an error by the runtime:
 
 - **RunPodSandbox**: Failure may result in pod creation failure
 - **StopPodSandbox**: Non-blocking for subsequent operations; the plugin should not depend on completion of subsequent teardown
--- a/README.md
+++ b/README.md
@@ -146,9 +146,12 @@ subscription.
 
 NRI plugins can subscribe to the following pod lifecycle events:
 
-  - creation
-  - stopping
-  - removal
+  - creation (RunPodSandbox)
+  - stopping (StopPodSandbox)
+  - removal (RemovePodSandbox)
+
+For detailed specifications of pod sandbox event timing, state requirements, and plugin
+expectations, see [Pod Sandbox Lifecycle Hooks](docs/pod-sandbox-lifecycle.md).
 
 The following table lists the pod sandbox properties exposed to NRI plugins, together with
 the first NRI, containerd and CRI-O versions each was available in.


SergeyKanzhelev:
same way we do conformance for runtimes, some of these contracts may be added to conformance: https://github.com/kubernetes-sigs/cri-tools/issues/2046. I think critest is the right place for it. But it opens an interesting discussion on whether we will also want to do conformance with other things like CNI. NRI is different enough that I would see it as a reasonable conformance requirement

aojea:
> other things like CNI.

CNI is not part of CRI, is an implementation detail of the runtimes and has several flaws that cause a lot of problem with current workloads ... during last kubecon during the OCI meeting we also discussed to replace it by a modular solution based on NRI ... I have a draft that once I have time I plan to finish and share, but I suggest to not include CNI as part of conformance of anything

docs/pod-sandbox-lifecycle.md
```diff
@@ -0,0 +1,161 @@
...
+
+### Subscribing to Events
+
+Plugins subscribe to these events during the Configure phase by returning the appropriate event flags in the ConfigureResponse:
+
+- `Event_RUN_POD_SANDBOX` (1 << 0) for RunPodSandbox
+- `Event_STOP_POD_SANDBOX` (1 << 1) for StopPodSandbox
+- `Event_REMOVE_POD_SANDBOX` (1 << 2) for RemovePodSandbox
+
+These events are notified via the StateChange RPC call.
```
klihub: This is a wire protocol detail which is going to change once #274 is merged, at which point we are going to have dedicated messages and RPC calls on the wire for every single pod and container life cycle event, and eventually remove `StateChange` after a deprecation period.

I think it would be better to refer here to the stub-offered API which (so far has) always had stronger stability and backward compatibility promise than the wire protocol. So could we perhaps rephrase this as "These events are delivered to plugins using the RunPodSandbox, StopPodSandbox and RemovePodSandbox event handlers." ?

docs/pod-sandbox-lifecycle.md
```diff
@@ -0,0 +1,161 @@
...
+- **RunPodSandbox**: Failure may result in pod creation failure
+- **StopPodSandbox**: Non-blocking for subsequent operations; the plugin should not depend on completion of subsequent teardown
+- **RemovePodSandbox**: Non-blocking; removal will proceed regardless of plugin timeout
+
+### Error Handling
+
+Plugins should handle errors gracefully and avoid leaving the pod or system in an inconsistent state. Error recovery strategies:
+
+- **RunPodSandbox errors**: Problematic; may block pod creation depending on failure severity and runtime policy
+- **StopPodSandbox errors**: May not prevent scenario termination depending on runtime policy
```
klihub: I think on the teardown path, actually both for pods and containers, we should not allow a plugin to try and prevent the operation with an error. If we agree, then we should clearly state here that, for StopPodSandbox and RemovePodSandbox, a plugin failing with an error will not prevent the operation from proceeding.

The current implementation has incorrect/inconstent behavior in this regard when multiple plugins are involved in the sense that for some of these (or corresponding container) teardown lifecycle events, a failure in a plugin will incorrectly prevent the event from being delivered to subsequent plugins, although it will not prevent the CRI-level operation from proceeding. There is a fix coming in for this, but it is waiting for #274 from get merged first (which is waiting for #277 to get merged first).

klihub:
@aojea Thank you, this looks great ! I only have a few comments.

docs/pod-sandbox-lifecycle.md
aojea: done

docs/pod-sandbox-lifecycle.md
aojea: @klihub can you please suggest the better wording for this?

I do not feel I'm able to translate that correctly to words :)

aojea: ok, let me take a stab at this,  I think now I got what you mean

aojea: see new commit rephrasing it

aojea:
> @aojea Thank you, this looks great ! I only have a few comments.

@klihub addressed comments on a new commit to simplify the review

klihub:
> > @aojea Thank you, this looks great ! I only have a few comments.
> 
> @klihub addressed comments on a new commit to simplify the review

@aojea Thanks ! This now LGTM. We can squash the commits before/during merging.

aojea:
squashed

klihub (approved):
LGTM.

dkennetzoracle:
@aojea thanks for putting this together, great doc.

docs/pod-sandbox-lifecycle.md
```diff
@@ -0,0 +1,162 @@
...
+For each event, this specification defines:
+
+- **Sandbox State Contract**: What sandbox infrastructure conditions runtimes MUST satisfy when firing the NRI event
+- **Plugin Responsibilities and Capabilities**: What plugins can safely do in response to the event
+
+## RunPodSandbox
+
+**CRI Operation**: RunPodSandbox - Creates and starts a pod-level sandbox.
+
+**NRI Event Timing**: The RunPodSandbox NRI event is fired after the runtime has successfully executed the CRI RunPodSandbox operation and the sandbox has reached a "Ready" state, but before any workload containers are started.
```
mikebrow:
```suggestion
**NRI Event Timing**: The RunPodSandbox NRI event is fired when the runtime has tentatively finished executing the CRI RunPodSandbox operation but just before setting the pod to the "Ready" state, which occurs immediately after NRI event processing, and thus before any workload containers are/can be started as the Pod is still in the "unknown" state.
```

docs/pod-sandbox-lifecycle.md
```diff
@@ -0,0 +1,162 @@
+# NRI Pod Sandbox Lifecycle Hooks
+
+## Relationship to CRI API
+
+This specification defines how NRI plugins interact with pod sandbox lifecycle events. The underlying pod sandbox operations are defined by the [Kubernetes CRI API](https://github.com/kubernetes/cri-api):
+
+- **RunPodSandbox (CRI)**: Creates and starts a pod-level sandbox. Runtimes must ensure the sandbox is in the ready state on success.
+- **StopPodSandbox (CRI)**: Stops any running process that is part of the sandbox and reclaims network resources.
```
mikebrow:
```suggestion
- **StopPodSandbox (CRI)**: Stops any running process that is part of the sandbox and directs the runtime to reclaim certain pod resources (e.g. Network Namespace, CNI teardown, and image mounts). May be called multiple times, and is idempotent. 
```

docs/pod-sandbox-lifecycle.md
```diff
@@ -0,0 +1,162 @@
...
+- **RemovePodSandbox (CRI)**: Removes the sandbox. If there are any running containers, they must be forcibly terminated and removed.
+
+This NRI specification details when and under what conditions NRI plugins receive notifications for these events, ensuring plugins can reliably depend on consistent sandbox state across different runtime implementations.
+
+## Overview
+
+The pod sandbox lifecycle consists of three distinct phases, each with a corresponding NRI event that plugins can subscribe to:
+
+1. **RunPodSandbox**: Fired after the runtime successfully executes CRI RunPodSandbox
```
mikebrow:
```suggestion
1. **RunPodSandbox**: Fired after the runtime successfully creates the pod, but before setting the pod to running and then replying success to CRI RunPodSandbox request.
```

docs/pod-sandbox-lifecycle.md
```diff
@@ -0,0 +1,162 @@
...
+### Sandbox State Contract
+
+When the runtime fires the RunPodSandbox NRI event, it guarantees:
+
+- The Pod-level cgroup hierarchy has been established
+- The Sandbox namespaces (IPC, Network, UTS) are created and active
+- Network setup has been fully configured (network interfaces are up and assigned addressing)
+- The pod IP address (if applicable) is assigned and available
+- The "pause" container (if the runtime uses one) is running
+- All prerequisite operations for workload container startup are complete
```
mikebrow:
```suggestion
- All prerequisite operations for workload container startup are complete, the pod is in the "unknown state" and will become "Ready" once the NRI event is processed. *This guarantees the NRI plugin has a window to allocate resources for the pod before any workload containers are started.
```

docs/pod-sandbox-lifecycle.md
```diff
@@ -0,0 +1,162 @@
...
+
+**NRI Event Timing**: The StopPodSandbox NRI event is fired when the runtime initiates the CRI StopPodSandbox operation.
+
+### Sandbox State Contract
+
+When the runtime fires the StopPodSandbox NRI event, it guarantees:
+
+- Workload containers within the sandbox are stopped or are stopping
+- **CRITICAL**: The sandbox infrastructure still exists and remains fully accessible during this hook
+- The network namespace is not unmounted or deleted until this hook completes
```
mikebrow:
```suggestion
- The pod resources allocated by the runtime; such as network namespace, CNI networks, and image mounts; are not unmounted or deleted until this hook completes
```

docs/pod-sandbox-lifecycle.md
```diff
@@ -0,0 +1,162 @@
...
+- Clean up custom firewall configurations, routing rules, or other network-level state
+- Perform graceful cleanup or resource release before sandbox teardown
+
+**Important**: Plugin processing must complete within the configured request timeout. Do not assume sandbox access persists after this hook returns or times out.
+
+## RemovePodSandbox
+
+**CRI Operation**: RemovePodSandbox - Removes the sandbox and forcibly terminates any remaining containers.
+
+**NRI Event Timing**: The RemovePodSandbox NRI event is fired when the runtime initiates the CRI RemovePodSandbox operation, during final garbage collection.
```
mikebrow:
```suggestion
**NRI Event Timing**: The RemovePodSandbox NRI event is fired by the runtime just prior to removing the pod from the pod list.
```

docs/pod-sandbox-lifecycle.md
```diff
@@ -0,0 +1,162 @@
...
+- Release any plugin resources held for this specific pod
+- Perform final accounting or bookkeeping
+
+**Important**: This hook is informational only. Plugins should not assume any pod infrastructure exists. Only clean up information the plugin created or stored internally.
+
+## Event Ordering and Guarantees
+
+Runtimes MUST guarantee the following ordering:
+
+1. **RunPodSandbox** NRI event fires after successful CRI RunPodSandbox execution
```
mikebrow:
```suggestion
1. **RunPodSandbox** NRI event fires after successful CRI RunPodSandbox execution, but before the pod is set to the "Ready" state
```

docs/pod-sandbox-lifecycle.md
```diff
@@ -0,0 +1,162 @@
...
+2. **StopPodSandbox** NRI event fires during CRI StopPodSandbox execution
```
mikebrow:
```suggestion
2. **StopPodSandbox** NRI event fires during CRI StopPodSandbox execution, just prior to removing the runtime pod resources allocated by the runtime; such as network namespace, CNI networks, and image mounts
```

mikebrow (commented):
looking good.. see comments

docs/pod-sandbox-lifecycle.md
samuelkarp:
```suggestion
1. **RunPodSandbox**: Fired during the the runtime CRI RunPodSandbox execution
```

docs/pod-sandbox-lifecycle.md
samuelkarp:
```suggestion
**NRI Event Timing**: The RunPodSandbox NRI event is fired after the runtime has successfully executed most of the CRI RunPodSandbox operation; NRI plugin execution is the final step before the sandbox reaches a "Ready" state. The Kubelet does not start workload containers until after the sandbox becomes "Ready".
```

docs/pod-sandbox-lifecycle.md
samuelkarp: The runtime doesn't initiate RemovePodSandbox, the Kubelet does.

aojea:
new commit trying to reconcile @mikebrow and @samuelkarp comments

docs/pod-sandbox-lifecycle.md
```diff
@@ -0,0 +1,162 @@
...
+- Perform network-level operations or monitoring
+- Inject sandbox-level hardware configurations (e.g., RDMA, RoCEv2)
+- Establish plugin-specific tracking or monitoring for the pod
+- Store initial state or baseline metrics for later reference
+
+Plugins should treat this as an initialization phase. The sandbox infrastructure will remain accessible throughout the pod's lifetime until StopPodSandbox is called.
+
+## StopPodSandbox
+
+**CRI Operation**: StopPodSandbox - Stops any running process that is part of the sandbox and reclaims network resources.
```
mikebrow:
```suggestion
**CRI Operation**: StopPodSandbox - Stops any running process that is part of the sandbox and reclaims certain pod resources (e.g. Network Namespace, CNI teardown, and image mounts).
```
repeats text from above... alternatively could end with ...", as described above" instead of listing the resource types

mikebrow:
pls squash :-)

mikebrow (approved):
LGTM

docs/pod-sandbox-lifecycle.md
```diff
@@ -0,0 +1,162 @@
...
+- **StopPodSandbox (CRI)**: Stops any running process that is part of the sandbox and directs the runtime to reclaim certain pod resources (e.g. Network Namespace, CNI teardown, and image mounts). May be called multiple times, and is idempotent.
+- **RemovePodSandbox (CRI)**: Removes the sandbox. If there are any running containers, they must be forcibly terminated and removed.
+
+This NRI specification details when and under what conditions NRI plugins receive notifications for these events, ensuring plugins can reliably depend on consistent sandbox state across different runtime implementations.
+
+## Overview
+
+The pod sandbox lifecycle consists of three distinct phases, each with a corresponding NRI event that plugins can subscribe to:
+
+1. **RunPodSandbox**: Fired during the the runtime CRI RunPodSandbox execution, after the PodSandbox is created but before setting the pod to running and then replying success to CRI RunPodSandbox request.
```
SergeyKanzhelev:
```suggestion
1. **RunPodSandbox**: Fired during the runtime CRI RunPodSandbox execution, after the PodSandbox is created but before setting the pod to running and then replying success to CRI RunPodSandbox request.
```

docs/pod-sandbox-lifecycle.md
```diff
@@ -0,0 +1,162 @@
...
+For each event, this specification defines:
+
+- **Sandbox State Contract**: What sandbox infrastructure conditions runtimes MUST satisfy when firing the NRI event
+- **Plugin Responsibilities and Capabilities**: What plugins can safely do in response to the event
+
+## RunPodSandbox
+
+**CRI Operation**: RunPodSandbox - Creates and starts a pod-level sandbox.
+
+**NRI Event Timing**: The RunPodSandbox NRI event is fired after the runtime has successfully executed most of the CRI RunPodSandbox operation; NRI plugin execution is the final step before the sandbox reaches a "Ready" state. The Kubelet does not start workload containers until after the sandbox becomes "Ready".
```
SergeyKanzhelev, line 30: Will container runtime keep retrying or fail on first failure/timeout?

docs/pod-sandbox-lifecycle.md
```diff
@@ -0,0 +1,162 @@
...
+
+### Sandbox State Contract
```
SergeyKanzhelev, line 32: If NRI plugin timed out, will it receive the Stop event?

docs/pod-sandbox-lifecycle.md
```diff
@@ -0,0 +1,162 @@
...
+
+When the runtime fires the RunPodSandbox NRI event, it guarantees:
+
+- The Pod-level cgroup hierarchy has been established
+- The Sandbox namespaces (IPC, Network, UTS) are created and active
+- Network setup has been fully configured (network interfaces are up and assigned addressing)
+- The pod IP address (if applicable) is assigned and available
+- The "pause" container (if the runtime uses one) is running
+- All prerequisite operations for workload container startup are complete, the pod is in the "unknown state" and will become "Ready" once the NRI event is processed. This guarantees the NRI plugin has a window to allocate resources for the pod before any workload containers are started.
```
SergeyKanzhelev, line 42: Can NRI pluging start it's own processes in this sandbox?

docs/pod-sandbox-lifecycle.md
```diff
@@ -0,0 +1,162 @@
...
+
+## RunPodSandbox
+
+**CRI Operation**: RunPodSandbox - Creates and starts a pod-level sandbox.
+
+**NRI Event Timing**: The RunPodSandbox NRI event is fired after the runtime has successfully executed most of the CRI RunPodSandbox operation; NRI plugin execution is the final step before the sandbox reaches a "Ready" state. The Kubelet does not start workload containers until after the sandbox becomes "Ready".
+
+### Sandbox State Contract
+
+When the runtime fires the RunPodSandbox NRI event, it guarantees:
```
SergeyKanzhelev, line 34: let's say something about volumes and DRA devices in this list

SergeyKanzhelev: Should DRA downward API be accessible?

docs/pod-sandbox-lifecycle.md
```diff
@@ -0,0 +1,162 @@
...
+
+### Plugin Responsibilities and Capabilities
+
+Upon receiving the RunPodSandbox event, plugins can safely:
+
+- Access the network namespace and inspect network configuration
+- Perform network-level operations or monitoring
+- Inject sandbox-level hardware configurations (e.g., RDMA, RoCEv2)
+- Establish plugin-specific tracking or monitoring for the pod
+- Store initial state or baseline metrics for later reference
```
SergeyKanzhelev, line 52: let's clarify the ordering of NRI plugins and what NRI plugin will "see"  if there are multiple plugins.

docs/pod-sandbox-lifecycle.md
```diff
@@ -0,0 +1,162 @@
...
+
+StopPodSandbox is the designated cleanup and observation phase for plugins. Upon receiving this event, plugins can:
+
+- Access the pod's network namespace to read final telemetry or metrics
+- Collect final state for observability or troubleshooting
+- Detach hardware interfaces or reconfigure resources
+- Clean up custom firewall configurations, routing rules, or other network-level state
+- Perform graceful cleanup or resource release before sandbox teardown
+
+**Important**: Plugin processing must complete within the configured request timeout. Do not assume sandbox access persists after this hook returns or times out.
```
SergeyKanzhelev, line 82: Again, I think we need to say that there will be a retry if timeout happened.

docs/pod-sandbox-lifecycle.md
```diff
@@ -0,0 +1,162 @@
...
+
+When the runtime fires the StopPodSandbox NRI event, it guarantees:
+
+- Workload containers within the sandbox are stopped or are stopping
+- **CRITICAL**: The sandbox infrastructure still exists and remains fully accessible during this hook
+- The pod resources allocated by the runtime; such as network namespace, CNI networks, and image mounts; are not unmounted or deleted until this hook completes
+- The pod's cgroups remain accessible
+- All pod-level resources remain stable until this hook returns
+
+### Plugin Responsibilities and Capabilities
```
SergeyKanzhelev, line 72: responsibility: this event processing must be re-entrant.

docs/pod-sandbox-lifecycle.md
```diff
@@ -0,0 +1,162 @@
...
+**CRI Operation**: StopPodSandbox - Stops any running process that is part of the sandbox and reclaims certain pod resources (e.g. Network Namespace, CNI teardown, and image mounts).
+
+**NRI Event Timing**: The StopPodSandbox NRI event is fired when the runtime initiates the CRI StopPodSandbox operation.
+
+### Sandbox State Contract
+
+When the runtime fires the StopPodSandbox NRI event, it guarantees:
+
+- Workload containers within the sandbox are stopped or are stopping
+- **CRITICAL**: The sandbox infrastructure still exists and remains fully accessible during this hook
```
SergeyKanzhelev, line 67: can runtime guarantee that the state is "good enough" to start another container in this sandbox?

docs/pod-sandbox-lifecycle.md
SergeyKanzhelev: From that perspective, can we say that it can be assumed as long as plugin keeps returning error?

docs/pod-sandbox-lifecycle.md
```diff
@@ -0,0 +1,162 @@
...
+
+## StopPodSandbox
+
+**CRI Operation**: StopPodSandbox - Stops any running process that is part of the sandbox and reclaims certain pod resources (e.g. Network Namespace, CNI teardown, and image mounts).
+
+**NRI Event Timing**: The StopPodSandbox NRI event is fired when the runtime initiates the CRI StopPodSandbox operation.
+
+### Sandbox State Contract
+
+When the runtime fires the StopPodSandbox NRI event, it guarantees:
```
SergeyKanzhelev, line 64: let's add guarantee: this sandbox will never be reused. There is never a scenario when any new workload (containers) will be started there by container runtime after the first attempt to Stop sandbox.

docs/pod-sandbox-lifecycle.md
```diff
@@ -0,0 +1,162 @@
...
+
+## RemovePodSandbox
+
+**CRI Operation**: RemovePodSandbox - Removes the sandbox and forcibly terminates any remaining containers.
+
+**NRI Event Timing**: The RemovePodSandbox NRI event is fired when the runtime initiates the CRI RemovePodSandbox operation, just prior to removing the pod from the pod list.
+
+### Sandbox State Contract
+
+When the runtime fires the RemovePodSandbox NRI event:
```
SergeyKanzhelev, line 92: another guarantee is that it never be called until the Stop succeeded

docs/pod-sandbox-lifecycle.md
SergeyKanzhelev, line 64: this can be called even if the Run was never called. Mostly for garbage collecting the sandbox which creation was interrupted. E.g. Run called, almost at the end containerd crashed, on restart containerd doesn't know if it needs clean up state so it must call Stop on each NRI plugin. Ortherwise there will be resources leak

docs/pod-sandbox-lifecycle.md
```diff
@@ -0,0 +1,162 @@
...
+
+- All workload containers have been removed
+- The StopPodSandbox operation has completed
+- Network setup teardown may be underway or complete
+- The pod's namespaces (Network, IPC, UTS) may have already been deleted
+- Pod-level cgroups may be destroyed
+- Sandbox infrastructure access is **not guaranteed**
+
+### Plugin Responsibilities and Capabilities
```
SergeyKanzhelev, line 101: Plugin must be reentrant here as well

docs/pod-sandbox-lifecycle.md
```diff
@@ -0,0 +1,162 @@
...
+## Event Ordering and Guarantees
+
+Runtimes MUST guarantee the following ordering:
+
+1. **RunPodSandbox** NRI event fires after successful CRI RunPodSandbox execution, but before the pod is set to the "Ready" state.
+2. **StopPodSandbox** NRI event fires during CRI StopPodSandbox execution, just prior to removing the runtime pod resources allocated by the runtime; such as network namespace, CNI networks, and image mounts
+3. **RemovePodSandbox** NRI event fires during CRI RemovePodSandbox execution
+4. These events MUST fire in strict order: RunPodSandbox → StopPodSandbox → RemovePodSandbox
+5. No workload containers will be started until after RunPodSandbox hook completes
+6. All workload containers will be stopped before StopPodSandbox hook is called
```
SergeyKanzhelev, line 123: this contradicts the contract above. It was saying stopped or being stopped:

> - Workload containers within the sandbox are stopped or are stopping

docs/pod-sandbox-lifecycle.md
```diff
@@ -0,0 +1,162 @@
...
+
+These events are delivered to plugins using the RunPodSandbox, StopPodSandbox and RemovePodSandbox event handlers.
+
+### Timeout Handling
+
+All plugin processing must complete within the configured request timeout. Plugins should plan accordingly:
+
+- **RunPodSandbox**: Failure may result in pod creation failure
+- **StopPodSandbox**: Non-blocking for subsequent operations; the plugin should not depend on completion of subsequent teardown
+- **RemovePodSandbox**: Non-blocking; removal will proceed regardless of plugin timeout
```
SergeyKanzhelev, line 146: why are we not guaranteeing successful execution here? It may lead to resource leak. With DRA we guarantee successful unprepare call, why shouldn't we do the same here?

docs/pod-sandbox-lifecycle.md
```diff
@@ -0,0 +1,162 @@
...
+
+- `Event_RUN_POD_SANDBOX` (1 << 0) for RunPodSandbox
+- `Event_STOP_POD_SANDBOX` (1 << 1) for StopPodSandbox
+- `Event_REMOVE_POD_SANDBOX` (1 << 2) for RemovePodSandbox
+
+These events are delivered to plugins using the RunPodSandbox, StopPodSandbox and RemovePodSandbox event handlers.
+
+### Timeout Handling
+
+All plugin processing must complete within the configured request timeout. Plugins should plan accordingly:
```
SergeyKanzhelev: this statement is hard to parse as a spec. I suggest to specify that the timeout is treated as error by runtime

docs/pod-sandbox-lifecycle.md
```diff
@@ -0,0 +1,162 @@
...
+- **RunPodSandbox**: Failure may result in pod creation failure
+- **StopPodSandbox**: Non-blocking for subsequent operations; the plugin should not depend on completion of subsequent teardown
+- **RemovePodSandbox**: Non-blocking; removal will proceed regardless of plugin timeout
+
+### Error Handling
+
+On the teardown path, plugin errors MUST NOT prevent the operation from proceeding. Runtimes MUST ensure that a failing plugin cannot block pod or container teardown:
+
+- **RunPodSandbox errors**: A plugin error may prevent the pod from being created, depending on runtime policy. Plugins bear responsibility for errors they return at this phase.
+- **StopPodSandbox errors**: A plugin error MUST NOT prevent the sandbox from being stopped. The runtime MUST proceed with teardown regardless of plugin failures.
```
SergeyKanzhelev, line 153: why? Is there any alternative way for plugin to handle clean up if runtime will not guarantee call and retry?

docs/pod-sandbox-lifecycle.md
```diff
@@ -0,0 +1,162 @@
...
+- **RemovePodSandbox errors**: A plugin error MUST NOT prevent the sandbox from being removed. The runtime MUST proceed with removal regardless of plugin failures.
+
+
+### Multi-Plugin Coordination
+
+When multiple plugins are active:
```
SergeyKanzhelev, line 159: if plugin didn't exist on sandbox creation and was notified about existing sandboxes on loading. What is the guarantee on consistency between that notification and plugin Stop/Remove calls?

docs/pod-sandbox-lifecycle.md
klihub: If an NRI plugin times out the runtime kicks it out by forcibly disconnecting the plugin.

docs/pod-sandbox-lifecycle.md
klihub: On errors other than plugin timeout, the runtime will fail the pod creation request. On plugin timeout the plugin is kicked out by the runtime.

docs/pod-sandbox-lifecycle.md
klihub: Not by using NRI for this. NRI itself does not provide means for this.

docs/pod-sandbox-lifecycle.md
klihub: Currently NRI plugins cannot mutate a pod via NRI. Therefore in the case of multiple plugins, all NRI plugins see identical pod sandbox data.

docs/pod-sandbox-lifecycle.md
mikebrow: no.. this is currently post stopping the pod which includes killing the containers and the pod container, during which when the pod container is killed and during exit processing of it's task the pod state is set to NotReady, which subsequently blocks new containers on the pod.  This call is pre-tear down of certain pod resources, noting there was a discrepancy in the crio placement of the call, WIP to nomalize.  

We could add another call earlier in the stop processing if we want to add the goal of creating notifications that stop pod has been requested. And possibly a third call to indicate the pod containers are stopped but the pod container is still running and thus another container can be added.

docs/pod-sandbox-lifecycle.md
mikebrow: nod.. at which point I believe the state of the pod would be "unknown"

docs/pod-sandbox-lifecycle.md
mikebrow: nod was wondering why the "or" didn't get a chance to verify that

docs/pod-sandbox-lifecycle.md
mikebrow: we have discussed MUST run plugins vs optional plugins. Is a good point that in this RM processing we need to also consider which of the plugins MUST complete, or possibly cause leaks.

docs/pod-sandbox-lifecycle.md
mikebrow: One alternative:  on sync of the list of pods a plugin may be able to normalize it's internal list of pods with resources vs the actual sync list.

docs/pod-sandbox-lifecycle.md
aojea: done

docs/pod-sandbox-lifecycle.md
aojea: I think this is related https://github.com/containerd/nri/issues/114

docs/pod-sandbox-lifecycle.md
aojea: NRI plugin has access to the OCI spec so it knows the necessary data to start its own processes on its own ... although that does not sounds like a good separation of concerns ... network plugins using NRI use to move network interfaces from the host to the network namespace per example

docs/pod-sandbox-lifecycle.md
aojea: this is out of scope of this proposal , that just try to write down current state to avoid problems like the one in https://github.com/kubernetes-sigs/dranet/issues/137 , where CRI-O and Containerd differences break compatibility , having consensus on the hooks lifecycle will allow to provide portability across platforms


The DRA / Kubernetes integration is being discussed in https://github.com/containerd/nri/issues/282

docs/pod-sandbox-lifecycle.md
aojea: As commented before, the goal is to document current state, not to make a proposal on how it should behave , this specific topic is discussed more in detail in https://github.com/containerd/nri/issues/114 . We also need to be careful on changing behaviors since those will break compatibility with existing plugins

docs/pod-sandbox-lifecycle.md
aojea: can you expand more on what you expect to see?
we can warn to plugin authors that if they do not make the logic reentrant then they will have problems ... but we can not enforce nothing here, you just define an interface, if the people implementing the interface makes it wrong 🤷

docs/pod-sandbox-lifecycle.md
aojea: is this something that existi today and we can guarantee or something we aim to achieve?

docs/pod-sandbox-lifecycle.md
aojea: good point

docs/pod-sandbox-lifecycle.md
aojea: yeah, I think it is

docs/pod-sandbox-lifecycle.md
aojea: we already have `- The StopPodSandbox operation has completed` , is not that covering this comment?

docs/pod-sandbox-lifecycle.md
aojea: This is coming from the API definition in CRI https://github.com/kubernetes/kubernetes/blob/bb0bcb8a85d54fad8f8f3aa036aa751c73138edb/staging/src/k8s.io/cri-api/pkg/apis/services.go#L73-L75

> 	// StopPodSandbox stops the sandbox. If there are any running containers in the
	// sandbox, they should be force terminated.
	StopPodSandbox(ctx context.Context, podSandboxID string) error


Although, the question here is if the runtimes are already doing this "force termination" step before calling the NRI hook or is undefined ... or if we want to define it and make it deterministic

docs/pod-sandbox-lifecycle.md
aojea: done

docs/pod-sandbox-lifecycle.md
aojea: I do not fully understand the question, the notifications are events generated by the runtime IIUIC, so the only guarantee is that you get the events when they happen ... plugins should reconcile the state using the `Synchronize` hook to receive all existing Pods running at the startup

aojea:
great questions , some of them open since require much more runtime knowledge that I have ... added a new commit to facilitate review, once we get consensus I will squash them

docs/pod-sandbox-lifecycle.md
MikeZappa87: If the RunPodSandbox allocates a resource that is finite, and StopPodSandbox errors, and we continue this would result in a problem? Today, we stop the process and keep trying.

docs/pod-sandbox-lifecycle.md
MikeZappa87: We have discussed the required vs optional and the role of each plugin in the past

SergeyKanzhelev:
For people on this thread - I have critest for this started. PRs will be linked from https://github.com/kubernetes-sigs/cri-tools/issues/2046

This PR has some discrepancies documented in NRI integration between Containerd and CRI-O:  https://github.com/kubernetes-sigs/cri-tools/pull/2069

docs/pod-sandbox-lifecycle.md
```diff
@@ -0,0 +1,163 @@
...
+
+When the runtime fires the RunPodSandbox NRI event, it guarantees:
+
+- The Pod-level cgroup hierarchy has been established
+- The Sandbox namespaces (IPC, Network, UTS) are created and active
+- The Sandbox will not be reused
+- Network setup has been fully configured (network interfaces are up and assigned addressing)
+- The pod IP address (if applicable) is assigned and available
+- The "pause" container (if the runtime uses one) is running
+- All prerequisite operations for workload container startup are complete, the pod is in the "unknown state" and will become "Ready" once the NRI event is processed. This guarantees the NRI plugin has a window to allocate resources for the pod before any workload containers are started.
```
tallclair, line 42: What about volumes? Currently, Kubelet waits for volumes to be attached & mounted before calling RunPodSandbox. Since the sandbox doesn't actually reference volumes at all, I want to move the WaitForVolumeAttachAndMount to after RunPodSandbox, which lets the Kubelet do both operations in parallel, cutting about 1/3rd of the latency on a basic pod with a prepulled image. Would this break the NRI contract?

SergeyKanzhelev: we may need somebody from sig storahe to assist here to understand the expectations NRI plugins may have

PR #289, opened 2026-04-28
moving up to latest golang

mikebrow:


samuelkarp:
> Error: can't load config: the Go language version (go1.25) used to build golangci-lint is lower than the targeted Go version (1.26.2)
>  The command is terminated due to an error: can't load config: the Go language version (go1.25) used to build golangci-lint is lower than the targeted Go version (1.26.2)

samuelkarp:
Probably better to upgrade the installed toolchain rather than the go.mod: https://github.com/containerd/nri/blob/57442b2a20c63d48d69dc600c3fb4f0c063c7ef0/.github/actions/install-go/action.yml#L6. The module itself doesn't need to require 1.26.

PR #292, opened 2026-05-07
Implementation of WithPluginRegistrationTimeout option

egorikas:
On a busy Kubernetes node, NRI plugin startup races against the runtime and loses. The plugin-side `DefaultRegistrationTimeout` is hardcoded at 5s with no override, while the runtime-side `plugin_registration_timeout` is operator-configurable (we set it to 10s in Uber env). 

When the plugin's 5s timer fires first. It gives up and closes its end of the ttrpc connection. Then the runtime finally accepts the connection. It reads the buffered register frame, which succeeds against stale state. The next call, Configure, returns `ttrpc.ErrClosed` right away because the plugin's side is already gone. The user sees `failed to register with NRI/Runtime: ttrpc: closed`. That hides the real cause: registration timed out before the runtime got to it.

The error made a lot of confusion for me. And in this PR I mirror the internal patch we did.

### DIFF
--- a/pkg/stub/stub.go
+++ b/pkg/stub/stub.go
@@ -283,6 +283,15 @@ func WithLogger(logger nrilog.Logger) Option {
 	}
 }
 
+// WithPluginRegistrationTimeout overrides the plugin's registration timeout
+// (default DefaultRegistrationTimeout, 5s).
+func WithPluginRegistrationTimeout(d time.Duration) Option {
+	return func(s *stub) error {
+		s.registrationTimeout = d
+		return nil
+	}
+}
+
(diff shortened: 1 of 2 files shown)


egorikas:
We've been testing this change on our prod for a month now, and it showed dramatic improvement in the plugin registration failures (we also bumped the containerd-side registration timeout as well). 

A review, anyone? That's a Uber-scale tested fix :)

egorikas:
Hey @samuelkarp! Sorry for pinging you directly, but I've noticed that you've reviewed some PRs in the past. Would you mind taking a look at this small bug fix I'm proposing?

PR #295, opened 2026-05-12
Support returning ip addresses with RunPodSandbox

MikeZappa87:
This PR came out of a discussion with @mikebrow regarding if an alternative exists to adding PodSandboxStatus to NRI. He proposed adding ip addresses to the RunPodSandboxResponse. 

#267


pkg/api/api.proto
```diff
@@ -206,7 +206,9 @@ message RunPodSandboxRequest {
   PodSandbox pod = 1;
 }
 
-message RunPodSandboxResponse{}
+message RunPodSandboxResponse{
+  repeated string ips = 1;
```
squeed, line 225: Should we add anything else? Interfaces? Routes?

MikeZappa87: I created an envelope msg that sets the stage for additional fields. Right now containerd has fields for the ip, additional ips but the fields you mention are part of the CNI result string. We could add them.

mikebrow: nod we can add the info map record for verbose status reporting later on.. with a follow-up

MikeZappa87: In the current we have a PodSandboxNetwork message which could provide typed fields instead? That might be preferable as we evolve the networking capabilities in NRI.

mikebrow: nod i presumed either map[key]=any json string... or typed fields if there's agreement on the definition

mikebrow: worst case you could type what is well known and likely.. then also allow a flex info map

MikeZappa87: I think we should do both. The same convention in cri-api helps with extensibility and luckily for networking, a lot of fields have been well defined for some time. I think once this is merged, we can look to adding those fields.

mikebrow: err ^ best case

mikebrow: i like our pattern of supporting annotation meta for optional/experimental projects, and typed fields where we have solid ground

MikeZappa87: +1 for that as well. Perhaps we can just go ahead and add

MikeZappa87: https://github.com/containerd/nri/blob/520641efc61c07ccb5a39b13dcfe5598ab1f9ab9/pkg/api/api.proto#L402 We 'could' just return PodSandbox type as well and extend it with new fields?

aojea: who will consume those fields? containerd and crio only knows about `ips` ... CNI ran already ...  kubelet only careas about the list of `ips`

pkg/stub/stub.go
```diff
@@ -320,7 +331,7 @@ type handlers struct {
 	Configure                   func(context.Context, string, string, string) (api.EventMask, error)
 	Synchronize                 func(context.Context, []*api.PodSandbox, []*api.Container) ([]*api.ContainerUpdate, error)
 	Shutdown                    func(context.Context)
-	RunPodSandbox               func(context.Context, *api.PodSandbox) error
+	RunPodSandbox               func(context.Context, *api.PodSandbox) (*api.PodSandboxAdjustment, error)
```
aojea, line 334: will this break plugins that need to upgrade to the new version?

pkg/api/api.proto
MikeZappa87: @aojea that has been my thought on why keeping it to just ip address will suffice for now. Unless other orchestrators (non k8s) have use cases. It probably makes sense to just keep it simple for now. I went back and forth

pkg/adaptation/adaptation.go
```diff
@@ -242,18 +242,46 @@ func (r *Adaptation) Stop() {
...
 // RunPodSandbox relays the corresponding CRI request to plugins.
-func (r *Adaptation) RunPodSandbox(ctx context.Context, req *RunPodSandboxRequest) error {
+// Plugins can return a PodSandboxAdjustment for the pod sandbox in their
+// response; the network IPs from every plugin are concatenated, in plugin
+// invocation order, into the response's PodSandboxAdjustment so they can be
+// propagated through the runtime (e.g. surfaced to the kubelet via CRI).
+// Duplicate IP addresses (whether reported by a single plugin or by multiple
+// plugins) are collapsed; first occurrence wins. Plugins that have nothing to
+// contribute can return an empty response or a nil response.
+func (r *Adaptation) RunPodSandbox(ctx context.Context, req *RunPodSandboxRequest) (*RunPodSandboxResponse, error) {
```
aojea, line 252: This requires changes on containerd and crio, otherwise it will not work, since CNI happens before this hook ... and now it will need to check after this hook again

PR #302, opened 2026-07-22
api,adaptation: add container image info

eshaanm25:
Add image information to container properties exposed to NRI plugins.

The following fields have been added:
- `name`: The image reference (ex. `registry.k8s.io/pause:3.10`)
- `digest`: Digest the image reference resolves to. Based on the image reference, it is a digest of one of:
  - image index
    - [OCI](https://github.com/opencontainers/image-spec/blob/main/image-index.md): `application/vnd.oci.image.index.v1+json`
    - [Docker](https://github.com/distribution/distribution/blob/main/docs/content/spec/manifest-v2-2.md): `application/vnd.docker.distribution.manifest.list.v2+json`
  - manifest 
    - [OCI](https://github.com/opencontainers/image-spec/blob/main/manifest.md): `application/vnd.oci.image.manifest.v1+json`
    - [Docker](https://github.com/distribution/distribution/blob/main/docs/content/spec/manifest-v2-2.md): `application/vnd.docker.distribution.manifest.v2+json`
- `config_digest`: The image configuration digest
  - [OCI](https://github.com/opencontainers/image-spec/blob/main/config.md): `application/vnd.oci.image.config.v1+json`
  - [Docker](https://github.com/distribution/distribution/blob/main/docs/content/spec/manifest-v2-2.md): `application/vnd.docker.container.image.v1+json`


klihub:
@eshaanm25 Thanks for the PR. You mention in #303 that you have PR for containerd with necessary runtime-side boilerplates/integration to populate the new fields. Do you have a similar PR or a working tree for CRI-O ?

eshaanm25:
I actually do have a mock PR for CRI-O, I wanted to ensure feature parity! Happy to make that PR too. If you'd like, I can make a draft PR to both container runtimes to show how the field will be populated.

klihub:
> I actually do have a mock PR for CRI-O, I wanted to ensure feature parity! Happy to make that PR too. If you'd like, I can make a draft PR to both container runtimes to show how the field will be populated.

Actually, I was mostly interested in whether you've tested that all the new bits of info are available in the CRI-O integration, too. The PR(s for both CRI-O and containerd) can probably wait till this is merged and available behind a new version tag. Although once this is merged, we could (and probably should) file PRs against containerd and CRI-O main/HEAD with the NRI deps pointing to the sha1-exact untagged version to get the necessary bits in.

klihub (approved):
LGTM.

mikebrow (approved):
LGTM thanks!

mikebrow:
nit on the field description .. oci is the primary support, but we should also work just fine with the schema v2 docker versions https://deepwiki.com/openshift/containers-image/4.1-docker-schema-support#manifest-media-types from which oci begat

samuelkarp (approved):

eshaanm25:
> nit on the field description .. oci is the primary support, but we should also work just fine with the schema v2 docker versions https://deepwiki.com/openshift/containers-image/4.1-docker-schema-support#manifest-media-types from which oci begat

Totally agreed, updated the PR description to include schema v2 docker versions as well as OCI. Let me know if you'd like me to update the commit message or if there is place in the code where I can make this clear! I kept the NRI descriptions short, but happy to add detail if you think it would be helpful.

mikebrow:
> > nit on the field description .. oci is the primary support, but we should also work just fine with the schema v2 docker versions https://deepwiki.com/openshift/containers-image/4.1-docker-schema-support#manifest-media-types from which oci begat
> 
> Totally agreed, updated the PR description to include schema v2 docker versions as well as OCI. Let me know if you'd like me to update the commit message or if there is place in the code where I can make this clear! I kept the NRI descriptions short, but happy to add detail if you think it would be helpful.

PR description is good for this PR IMO.  Perhaps with a follow up add more description to the api / parameters in the image structure, only because there have been previous compatibility issues between containerd/crio for some of the CRI fields in the image service.  In that follow up could add image to one of the sample plugins.

At some point in the future we may end up adding image services to the plugin interface.

PR #310, opened 2026-08-20
chore(deps): bump google.golang.org/protobuf v1.36.12, google.golang.org/grpc v1.79.3, sigs.k8s.io/yaml v1.5.0

thaJeztah:
- [x] stacked on https://github.com/containerd/nri/pull/309

### chore(deps): bump google.golang.org/grpc v1.79.3

Update to a newer version to reduce transitive dependencies on deprecated
modules.

This also pleases vulnerability scanners scanners on GHSA-p77j-4mvh-x3m3
(CVE-2026-33186); this is a false positive, but most consumers of this
module already updated to a newer version og grpc, so we may as well
update.

Note that there's newer versions available, which also fix some CVEs
(high, not critical), but keeping the version slightly conservative.

### chore(deps): bump sigs.k8s.io/yaml v1.5.0

Update to a newer version that drops dependencies on the deprecated
gopkg.in/yaml.v2, gopkg.in/yaml.v3 modules. I did not pick the latest
(v1.6.0) version as that introduces an extra dependency that may not
be needed.

Manually updated the indirect go.yaml.in/yaml/v2 to its latest
version (v2.4.4) which no longer has any dependencies.

PR #311, opened 2026-08-20
chore(deps): bump sigs.k8s.io/yaml v1.5.0

thaJeztah:
- extracted from https://github.com/containerd/nri/pull/310

Update to a newer version that drops dependencies on the deprecated gopkg.in/yaml.v2, gopkg.in/yaml.v3 modules. I did not pick the latest (v1.6.0) version as that introduces an extra dependency that may not be needed.

Manually updated the indirect go.yaml.in/yaml/v2 to its latest version (v2.4.4) which no longer has any dependencies.


klihub (approved):

mikebrow (approved):
LGTM

PR #312, opened 2026-08-20
chore(deps): bump google.golang.org/grpc v1.65.1

thaJeztah:
Update to a newer version to remove the dependency on the deprecated github.com/golang/protobuf module. We cannot update to v1.66.0 or newer because those pull in github.com/planetscale/vtprotobuf v0.6.x, which is incompatible with github.com/knqyf263/go-plugin v0.9.0 that's used in this repository.

PR #313, opened 2026-08-20
chore(deps): plugins/differ bump github.com/r3labs/diff/v3 v3.0.2, github.com/goccy/go-yaml v1.13.7

thaJeztah:
- [x] stacked on https://github.com/containerd/nri/pull/312

### chore(deps): plugins/differ bump github.com/r3labs/diff/v3 v3.0.2

Updates the indirect github.com/vmihailenco/msgpack to v5, removing
dependency on google.golang.org/appengine and through that,
github.com/golang/protobuf.

### chore(deps): plugins/differ bump github.com/goccy/go-yaml v1.13.7

v1.13.7 and up no longer have any dependencies.

### DIFF
--- a/plugins/differ/go.mod
+++ b/plugins/differ/go.mod
@@ -4,7 +4,7 @@ go 1.24.0
 
 require (
 	github.com/containerd/nri v0.6.1
-	github.com/r3labs/diff/v3 v3.0.0
+	github.com/r3labs/diff/v3 v3.0.2
 	github.com/sirupsen/logrus v1.9.4
 	github.com/sters/yaml-diff v0.4.0
 	sigs.k8s.io/yaml v1.5.0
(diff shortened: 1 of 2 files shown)


samuelkarp (approved):

thaJeztah:
rebased, after https://github.com/containerd/nri/pull/312 was merged

mikebrow (approved):
oh nice!

plugins/differ/go.mod
```diff
@@ -4,7 +4,7 @@ go 1.24.0
 
 require (
 	github.com/containerd/nri v0.6.1
```
thaJeztah, line 6: @mikebrow @samuelkarp One thing we probably need to do is update these versions; currently all the modules use a `replace`, so they're effectively testing against "latest nri", and not the version in `go.mod`, so the most correct thing would be to raise the minimum version for NRI itself.

mikebrow: nod

